PR #24, opened 2024-10-03
1 enhance docstrings to use google format

albertohernandez1995:
Created docstrings using google type format.


javihern98 (approved):
All good on Google documentation

PR #26, opened 2024-10-04
Add mypy

mla2001:
Add mypy to the project and solve the incoming errors.


javihern98 (approved):
All good on types. Will also be checked on development over the upcoming months.

PR #32, opened 2024-10-17
Fixed non mandatory missing components not autofilled with NaN on dat…

mla2001:
Fixed  non mandatory missing components not autofilled with NaN on datapoints.


src/vtlengine/files/parser/__init__.py
```diff
@@ -143,8 +143,16 @@ def _validate_pandas(
...
     # Identifier checking
+
     id_names = [comp_name for comp_name, comp in components.items() if comp.role == Role.IDENTIFIER]
 
+    missing_columns = [name for name in components.keys() if name not in data.columns.tolist()]
+    if missing_columns:
+        for name in missing_columns:
+            if components[name].nullable is False:
+                raise SemanticError("0-1-1-10", name=dataset_name, comp_name=name)
+        data = data.reindex(columns=list(components.keys()))
```
javihern98, line 154: Do not reindex as we may rename already present columns with wrong names.

Suggestion:
```python
data[name] = None

javihern98 (changes requested):
A small change is needed to prevent reindex. Rest is OK.

javihern98 (approved):
All good now.

PR #33, opened 2024-10-17
Added Month Time Operator with its tests

albertohernandez1995:
Month operator has been created.

### DIFF
--- a/src/vtlengine/Operators/Time.py
+++ b/src/vtlengine/Operators/Time.py
@@ -1,12 +1,10 @@
 import re
-from datetime import date
+from datetime import date, datetime
 from typing import Optional, Union, List
 
-import vtlengine.Operators as Operators
 import pandas as pd
-from vtlengine.DataTypes import Date, TimePeriod, TimeInterval, Duration
-from vtlengine.DataTypes.TimeHandling import DURATION_MAPPING, date_to_period, TimePeriodHandler
 
+import vtlengine.Operators as Operators
 from vtlengine.AST.Grammar.tokens import (
     TIME_AGG,
     TIMESHIFT,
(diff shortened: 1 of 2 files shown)


src/vtlengine/Operators/Time.py
```diff
@@ -759,3 +760,22 @@ def evaluate(cls):
...
+
+
+class Month(Operators.Unary):
+    op = 'month'
+
+    @classmethod
+    def py_op(cls, value: str):
+        if '/' in value:
+            raise RuntimeError
+        if '-' in value:
```
javihern98, line 772: Need to check if there is a Date here.

Suggestion: use this python code to check if there are two hyphens on same str:

```python
if value.count('-') == 2:
   # Date processing
  .....
```

src/vtlengine/Operators/Time.py
```diff
@@ -759,3 +760,22 @@ def evaluate(cls):
         result = cls.validate()
         result.value = date.today().isoformat()
         return result
+
+
+class Month(Operators.Unary):
```
javihern98, line 765: Must inherit from a superclass that prevents the use of the Dataset evaluation

```python
class SimpleUnaryTime(Operators.Unary):

    def validate(operand: .....):
         if isinstance(operand: Dataset):
              raise SemanticError(............)
         # Validation for DataComponent and Scalar
        ..................

    # Same for evaluate

class Month(SimpleUnaryTime):
      ....................
```

javihern98 (changes requested):
Does not cover still all cases

javihern98 (approved):
All good on code, peer reviewed with @albertohernandez1995

PR #44, opened 2024-10-29
15 add implementation of dateadd operator

mla2001:


src/vtlengine/Operators/Time.py
```diff
@@ -810,14 +825,92 @@ def validate(cls, Date1: Any, Date2: Any) -> Any:
...
+                    measure.data_type = Date
+
+        if isinstance(result, (Scalar, DataComponent)):
+            result.data_type = Date
+        return result
 
     @classmethod
-    def validate(cls, operand: Any, param_list: List[Any]) -> Any:
-        pass
+    def op_date(cls,
```
javihern98, line 893: Maybe change the name to py_op for consistency with other operators?

src/vtlengine/Operators/Time.py
```diff
@@ -810,14 +825,92 @@ def validate(cls, Date1: Any, Date2: Any) -> Any:
...
+    @classmethod
+    def evaluate(cls,
+                 operand: Union[Scalar, DataComponent, Dataset],
+                 param_list: List[Scalar]
+                 ) -> Union[Scalar, DataComponent, Dataset]:
+        result = cls.validate(operand, param_list)
+        shift, period = param_list[0].value, param_list[1].value
+        is_tp = isinstance(operand, (Scalar, DataComponent)) and operand.data_type == TimePeriod
+
+        if isinstance(result, Scalar) and isinstance(operand, Scalar) and operand.value is not None:
```
javihern98, line 872: On evaluate, we must check only if the operand is instance of Dataset, DataComponent or Scalar, at no point the result and the operand should be different in this operator.

src/vtlengine/Operators/Time.py
```diff
@@ -810,14 +825,92 @@ def validate(cls, Date1: Any, Date2: Any) -> Any:
...
+            result.value = cls.op_date(operand.value, shift, period, is_tp)
+        elif (isinstance(result, DataComponent) and isinstance(operand, DataComponent) and
+              operand.data is not None):
```
javihern98, line 875: No need to check if operand.data is None, as this only happens on the semantic analysis and the validate method will handle it on its own.

javihern98 (commented):
Slight changes to be checked before merging

src/vtlengine/Operators/Time.py
mla2001: Changed!

src/vtlengine/Operators/Time.py
mla2001: This was done to resolve mypy errors. If one of them is missing, mypy will generate a union-attr error.
Example: On line 872, if 'isinstance(result, Scalar)' is missing, mypy will display the following errors:

- src\vtlengine\Operators\Time.py:873:13: error: Item "DataComponent" of "Scalar | DataComponent | Dataset" has no attribute "value"  [union-attr]
- src\vtlengine\Operators\Time.py:873:13: error: Item "Dataset" of "Scalar | DataComponent | Dataset" has no attribute "value"  [union-attr]

src/vtlengine/Operators/Time.py
mla2001: This was done to resolve mypy errors. If one of them is missing, mypy will generate a union-attr error.
Example: On line 876, if 'operand.data is not None' is missing, mypy will display the following error:

- src\vtlengine\Operators\Time.py:875:27: error: Item "None" of "Any | None" has no attribute "map"  [union-attr]

src/vtlengine/Operators/Time.py
javihern98: Noted, keeping it

javihern98 (approved):
All good

javihern98 (approved):

PR #46, opened 2024-11-08
45 check time period comparison

mla2001:


javihern98:
Awaiting for clarification on the criteria to compare Time Periods

src/vtlengine/DataTypes/TimeHandling.py
```diff
@@ -236,9 +236,16 @@ def _meta_comparison(self, other: Any, py_op: Any) -> Optional[bool]:
             if len(other) == 0:
                 return False
             other = TimePeriodHandler(other)
-        return py_op(
-            DURATION_MAPPING[self.period_indicator], DURATION_MAPPING[other.period_indicator]
-        )
+
+        if DURATION_MAPPING[self.period_indicator] != DURATION_MAPPING[other.period_indicator]:
```
javihern98, line 240: We need to review the criteria here, I believe we should check first the duration, then the year then the period number. A ticket for the VTL community is still pending to be produced to address the criteria

javihern98 (commented):
We still need to decide on the criteria

src/vtlengine/DataTypes/TimeHandling.py
javihern98: As specified in: https://github.com/VTL-Community/VTL-Community/issues/58#issuecomment-2511320746

The only way of comparing time periods is to specify left and and right time points:

2022Q1 = [2022-01-01, 2022-03-31) 2023M1 = [2023-01-01, 2023-01-31]

Now the periods could be disjoint (as in this case), so the answer to the initial question is very easy: 2023M1 is greater than 2022Q1 because in the time axis 2023M1 is completely on the right of 2022Q1.
For overlapping time periods, since usually one of the two ends coincides, the "largest" or the one with greater right end is the greater, as we have introduced a "default order" in VTL 2.1 (Y>S>Q>M) so 2022S1 >= 2022Q2 and 2022Q1>=2022M3.
I hope this can help.

javihern98: @mla2001 Please change this

mla2001: Refactored to follow this new criteria.

mla2001: **New comparison criteria:**
Translate the period to its extreme dates that make it up and do the following:
- Something is greater or lesser if:
    - Its temporal end is more to the right (more in the future) if it is > or more to the left (more in the past) if it is < 
    - In case the extreme to be compared is equal, the one with the longest time span covers: Y > S > Q > M...
- The same operation will only be true with the same date and period.

javihern98 (commented):
@mla2001 Please merge main here solving conflicts. This is the last PR pending. Double check after the merge the Files Changed in the PR to ensure we are only modifying the related files. Ensure we are using ruff 0.8.3 and mypy 1.14.1

javihern98 (approved):
All good 😊

PR #62, opened 2025-01-24
55 check duration type

albertohernandez1995:
Duration class was updated to support iso 8601 format. There are 2 tests that are failing with some weird errors that aren't related to the Duration datatype. Please @javihern98 check it.


src/vtlengine/DataTypes/__init__.py
```diff
@@ -425,6 +436,32 @@ def explicit_cast(cls, value: Any, from_type: Any) -> Any:
...
+                "2-1-19-15", "{op} can only be applied according to the iso 8601 format mask"
+            )
+
+        match = re.match(cls.iso8601_duration_pattern, value)
+
+        years = 0
+        months = 0
+        days = 0
+
+        if match:
```
javihern98, line 452: Need to handle better this if, what happens if there is no match? Currently it will return None, maybe check if match is None then make a explicit return as None or raise a Semantic Error

src/vtlengine/Operators/Time.py
```diff
@@ -1113,28 +1114,26 @@ class Year_to_Day(Operators.Unary):
     return_type = Integer
 
     @classmethod
-    def py_op(cls, value: str) -> int:
-        if "/" in value:
+    def py_op(cls, value: Duration) -> int:
+        if "/" in value:  # type: ignore[operator]
```
javihern98, line 1118: If we only have a Duration value, we cannot have a / (by definition). Please remove this if or add a specific test for it

src/vtlengine/Operators/Time.py
```diff
@@ -1113,28 +1114,26 @@ class Year_to_Day(Operators.Unary):
...
 
 class Month_to_Day(Operators.Unary):
     op = MONTHTODAY
     return_type = Integer
 
     @classmethod
-    def py_op(cls, value: str) -> int:
-        if "/" in value:
+    def py_op(cls, value: Duration) -> int:
+        if "/" in value:  # type: ignore[operator]
```
javihern98, line 1133: Same as above

src/vtlengine/Operators/Time.py
```diff
@@ -1113,28 +1114,26 @@ class Year_to_Day(Operators.Unary):
...
-        if "M" not in value:
-            raise SemanticError("2-1-19-16", op=cls.op)
-        index_m = value.index("M")
-        months = int(value[1:index_m])
-        days = int(value[(index_m + 1) : -1])
-        return months * 30 + days
+        try:
+            days = Duration.to_days(value)
+        except SemanticError:
+            raise SemanticError("0-1-1-12", op=cls.op)
```
javihern98, line 1138: Cannot use error 0 inside an operator, need to use another error message. In any case check if this try-except is necessary

src/vtlengine/files/parser/__init__.py
```diff
@@ -199,18 +199,38 @@ def _validate_pandas(
...
+                            raise ValueError(
+                                f"Duration values are not correct in column {comp_name}"
+                            )
+                    except ValueError:
+                        raise ValueError(f"Duration values are not correct in column {comp_name}")
+                    else:
+                        data[comp_name] = data[comp_name].map(
+                            lambda x: str(x).replace('"', ""), na_action="ignore"
+                        )
+                    data[comp_name] = data[comp_name].astype(np.object_, errors="raise")
```
javihern98, line 227: This line is redundant, it is already covered by line 232

src/vtlengine/files/parser/__init__.py
```diff
@@ -199,18 +199,38 @@ def _validate_pandas(
...
+                            )
+                            .all()
+                        )
+                        if not values_correct:
+                            raise ValueError(
+                                f"Duration values are not correct in column {comp_name}"
+                            )
+                    except ValueError:
+                        raise ValueError(f"Duration values are not correct in column {comp_name}")
+                    else:
```
javihern98, line 223: Check the usefulness of having else here, I believe this is redundant

javihern98 (changes requested):
Need to make some minor changes, overall the solution is good

albertohernandez1995:
All requested changes have been made.

src/vtlengine/DataTypes/__init__.py
```diff
@@ -461,6 +461,8 @@ def to_days(cls, value: Any) -> int:  # type: ignore[return]
                 days = int(days_str[:-1])
             total_days = years * 365 + months * 30 + days
             return int(total_days)
+        if not match:
```
javihern98, line 464: Not what we agreed, please remove both ifs and leave the above lines

src/vtlengine/Operators/Time.py
```diff
@@ -1113,28 +1114,16 @@ class Year_to_Day(Operators.Unary):
...
-    def py_op(cls, value: str) -> int:
-        if "/" in value:
-            raise SemanticError("2-1-19-11", op=cls.op)
-        if "Y" not in value:
-            raise SemanticError("2-1-19-15", op=cls.op)
-        index_y = value.index("Y")
-        years = int(value[1:index_y])
-        days = int(value[(index_y + 1) : -1])
-        return years * 365 + days
+    def py_op(cls, value: Duration) -> int:
```
javihern98, line 1117: Still need to change the signature, should be str

src/vtlengine/Operators/Time.py
```diff
@@ -1113,28 +1114,16 @@ class Year_to_Day(Operators.Unary):
...
-    def py_op(cls, value: str) -> int:
-        if "/" in value:
-            raise SemanticError("2-1-19-11", op=cls.op)
-        if "M" not in value:
-            raise SemanticError("2-1-19-16", op=cls.op)
-        index_m = value.index("M")
-        months = int(value[1:index_m])
-        days = int(value[(index_m + 1) : -1])
-        return months * 30 + days
+    def py_op(cls, value: Duration) -> int:
```
javihern98, line 1127: Same as above

javihern98 (commented):
Still need to make some changes @albertohernandez1995

src/vtlengine/DataTypes/__init__.py
albertohernandez1995: Removed if conditions

src/vtlengine/Operators/Time.py
albertohernandez1995: Signature was changed to str.

src/vtlengine/Operators/Time.py
albertohernandez1995: Signature was changed to str.

albertohernandez1995:
I reviewed and redid the requested changes because I misunderstood them.

javihern98 (approved):
All good now, great work! @albertohernandez1995

PR #81, opened 2025-02-21
Updated if condition merging process to work with new Parentheses AST…

mla2001:
… construction (not added in this branch). Optimized nested if condition resolution.


javihern98 (approved):
All good, thanks! 😊

PR #84, opened 2025-02-23
Library extras and refactor S3

javihern98:
# Summary of changes

- Implemented S3 reading using Pandas
- Defined S3 compatibility as extra
- Adapted Pyproject.toml to Poetry 2.0


javihern98:
Awaiting to fix https://github.com/bis-med-it/pysdmx/issues/236

src/vtlengine/files/parser/__init__.py
```diff
@@ -121,31 +121,14 @@ def _pandas_load_csv(components: Dict[str, Component], csv_path: Path) -> pd.Dat
             na_values=[""],
         )
     except UnicodeDecodeError:
-        raise InputValidationException(code="0-1-2-5", file=csv_path.name)
+        if isinstance(csv_path, Path):
```
albertohernandez1995, line 124: According to code coverage, those lines aren't covered

albertohernandez1995 (changes requested):
Code its fine, although I think that we can improve code coverage

src/vtlengine/files/parser/__init__.py
javihern98: Removed the exception on UnicodeDecode error and replaced with errors='replace' and utf-8 encoding on pandas read_csv and python IO handler

PR #97, opened 2025-04-16
Add prettify method

albertohernandez1995:
Prettify api method was implemented following the old engine structure. Now, when applying prettify, the vtl expression will be displayed formatted.
```python
def prettify(text: str) -> str:
    from vtlengine.AST.ASTComment import create_ast_with_comments

    ast = create_ast_with_comments(text)
    return ASTString(pretty=True).render(ast)
```
Example:

```python
# Vtl expression with no pretty:
DS_r := DS_1 + DS_2;
# Vtl expression with pretty:
DS_r :=
	DS_1 + DS_2;
```


src/vtlengine/AST/ASTString.py
```diff
@@ -178,15 +219,32 @@ def visit_Argument(self, node: AST.Argument) -> str:
...
         if self.is_first_assignment:
             self.is_first_assignment = False
-        expression = f"{self.visit(node.left)} {node.op} {self.visit(node.right)}"
+        if self.pretty:
+            if return_element == False:
+                expression = f"{self.visit(node.left)} {node.op}\n\t{self.visit(node.right)}"
+            else:
+                expression = f"{self.visit(node.left)} {node.op} {self.visit(node.right)}"
+        else:
+            expression = f"{self.visit(node.left)} {node.op} {self.visit(node.right)}"
```
javihern98, line 247: Re-check this to use the self.is_first_assignment condition to discriminate between a regular assignment and an assignment inside a calc

javihern98 (changes requested):
We need to use variables for the \t and \n, they will be set based on the value of the pretty flag:

```python
nl = child1 = ... = ''

if self.pretty:
  nl = '\n'
  child1 = '\t'
  child2 = '\t\t'
  ...

```

Overall code is fine but hard to read. Further discussions on windowing (analytic) with prettier is needed and an in-depth review on tests

src/vtlengine/AST/ASTString.py
albertohernandez1995: Changed condition by using is_first_assignment instead of return element

tests/AST/data/prettier/reference_filter_cast_time_period.txt
```diff
@@ -0,0 +1,6 @@
+test2_1 :=
+	BE2_DF_NICP
+		[filter
+				FREQ = "M"and
+					TIME_PERIOD = cast("2020-01",time_period)
```
javihern98, line 5: Review this reference to add a space between "M" and "and"

tests/AST/data/prettier/reference_group_all_time_aggr.txt
```diff
@@ -0,0 +1,4 @@
+DS_r :=
+	sum(
+			DS_1 group all time_agg("D",_,Id_1)
```
javihern98, line 3: Do not use here double indentation for the operand, remove one tab

javihern98 (changes requested):
Review small changes missing

tests/AST/data/prettier/reference_group_all_time_aggr.txt
albertohernandez1995: Removed 1 tab for the operand

tests/AST/data/prettier/reference_filter_cast_time_period.txt
albertohernandez1995: Removed .strip() on operand to solve this issue

javihern98 (approved):
All good now, thanks! 😊

PR #104, opened 2025-04-29
Update code coverage up to 90%

albertohernandez1995:
When the issue was created, general code coverage was at 89%, now it is at 90%. We have focused on ASTTemplate module, where there was only a 62% lines covered. At this point, it is at 95%. The reason why it is not at 100% is that there are 4 functions that are unused and I think we will have to take a look, which are:

- visit_Optional
- visit_Role
- visit_Type
- visit_NoOp

In fact, the last one has a pylint pointing that is disabled.


tests/AST/test_AST.py
```diff
@@ -111,3 +135,354 @@ def test_visit_TimeAggregation_error():
...
+
+
+def test_visit_Argument():
+    from unittest.mock import Mock
+
+    visitor = ASTTemplate()
+    visitor.visit = Mock()
+    type_node = Mock()
+    default_node = Mock()
+    node = Argument(
```
jfrancma, line 379: We should check if these mocks are correct in all tests and if they are valid for this kind of tests and we can't do it in any other way.

jfrancma (changes requested):
Add correct assertions for each test and check the tests with mocks.

tests/AST/test_AST.py
albertohernandez1995: Mocks were changed and checked. Now they should be fine.

jfrancma (approved):

PR #105, opened 2025-05-02
Added check_script function

albertohernandez1995:
Check_script function was created to support the use of TransformationScheme objects as vtl expressions by treating them with generating_vtl_expression tool from pysdmx library. As a result, now TS objects can be treated as inputs, increasing the compatibility between pysdmx and vtlengine.

## Major changes:
- AstString render was modified so it can be use in ast_to_sdmx function properly. Because of the implement of prettify, there were conflicts with the original function that are now solved. 

- _check_script function was created.

- Ast_to_sdmx function was updated by making optional the existence of RulesetSchemes and UDOSchemes objects, There was a problem of incompatibility between pysdmx model_validation and the previous version of this function. As a result, the TransformationScheme object now will be returned and validated.

> [!IMPORTANT]
@javihern98 These changes were made because of the last release of pysdmx. Please check them.


src/vtlengine/API/_InternalApi.py
mla2001, line 1: It seems mostly good, just a minor change:
Don't use local imports in the _check_script method (move it to the top of the file)

mla2001 (changes requested):
It seems mostly good, just a minor change:
Don't use local imports in the _check_script method (move it to the top of the file)

src/vtlengine/API/_InternalApi.py
albertohernandez1995: I did it like that to avoid circular imports. Not sure how to solved this issue without importing it locally

albertohernandez1995: Thanks for the feedback!!

mla2001: Ok then, no problem in that case!

albertohernandez1995:
Mypy issues solved!

mla2001 (approved):
Looks fine!

PR #112, opened 2025-05-06
Improved ast_to_sdmx function

albertohernandez1995:
ast_to_sdmx function was improved by replacing ruleset_schemes and udo_schemes with kwarg arguments. With this approach, now the TS object, if there aren't any UDOS or RS, it won't show those arguments.

A big test has been created that covers the whole process and compatibility with pysdmx. It starts with a vtl_expression that it is transformed to a TS, and then it is rollbacked again to a vtl expression.

Small bug in AST.render() was solved that didn't allow the showcase of error levels and codes when a hierarchical ruleset is used to be transformed on a TS.


src/vtlengine/API/_InternalApi.py
```diff
@@ -500,26 +500,27 @@ def ast_to_sdmx(ast: AST.Start, agency_id: str, version: str) -> TransformationS
...
-        else [],
+
+    transformation_scheme = TransformationScheme(
+        items=list_transformation,
+        agency=agency_id,
+        id="TS1",
+        vtl_version="2.1",
+        version=version,
+        **references,  # type: ignore[arg-type]
     )
```
javihern98, line 524: Double check the mypy errors. Why are there any ignores? Please describe the mypy errors

tests/API/test_api.py
```diff
@@ -1177,25 +1180,12 @@ def test_generate_sdmx(script, agency_id, version):
...
-        assert result.user_defined_operator_schemes == []
-
-
-def test_check_script_with_transformation_scheme():
-    mock_scheme = TransformationScheme(id="TS1", agency="MD", version="1.0", vtl_version="2.1")
-    with patch("pysdmx.toolkit.vtl.generate_vtl_script.generate_vtl_script") as mock_generate:
-        mock_generate.return_value = "DS_r := DS_1 + DS_2;"
-        result = _check_script(mock_scheme)
-        assert result == "DS_r := DS_1 + DS_2;"
-        mock_generate.assert_called_once_with(mock_scheme, model_validation=True)
```
javihern98, line 1198: If we remove this test, we still need to add a new test that starts from a static TransformationScheme object (with some transformations but no RulesetScheme or UserDefinedOperationScheme). This test will be used for easier troubleshooting of the check_script function only, it should have a dedicated vtl file as reference with some transformations.

albertohernandez1995: Added a test that covers it.

src/vtlengine/API/_InternalApi.py
albertohernandez1995: Solved mypy issue

javihern98 (approved):
All good now, thanks! 😊

PR #114, opened 2025-05-08
Updated fail threshold on coverage to 90

javihern98:
- Updated GH workflow to fail below 90%


albertohernandez1995 (approved):
Nice update. All good!

PR #116, opened 2025-05-08
Add compatibility with python 3.13

javihern98:
- Updated workflow to add python 3.13 in OS matrix
- Small changes on numpy handling to avoid access violations on windows
- Added numpy >2.1.0 on python_version >= 3.13 to prevent access violations on python 3.13 for windows
- Added pandas version > 2.2 on python_version >= 3.13 (available at Ubuntu 25.04)


jfrancma (approved):
Seems good, ready to merge.

PR #117, opened 2025-05-08
Documentation updates

albertohernandez1995:
Added documentation on Sphinx to the following API methods:

- Prettify
- run_sdmx
- generate_sdmx

Added docstrings to API and internal methods

### DIFF
--- a/docs/walkthrough.rst
+++ b/docs/walkthrough.rst
@@ -4,7 +4,7 @@
 
 Summarizes the main functions of the VTL Engine
 
-The VTL Engine API implements four basic methods:
+The VTL Engine API implements five basic methods:
 
 * **Semantic Analysis**: aimed at validating the correctness of a script and computing the data
   structures of the data sets created in the script.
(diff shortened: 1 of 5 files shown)


javihern98:
We need to add as well the documentation in sphinx, it is necessary that the users are aware of these new features

docs/walkthrough.rst
```diff
@@ -192,4 +197,50 @@ returns:
...
+
+    """
+    define hierarchical ruleset accountingEntry(variable rule ACCOUNTING_ENTRY) is
+        B = C - D
+	    errorcode "Balance (credit-debit)"
+	    errorlevel 4;
+
+	    N = A - L
+	    errorcode "Net (assets-liabilities)"
+	    errorlevel 4
```
javihern98, line 236: Review indent here, this rule needs to be at the same level as the previous one

javihern98 (changes requested):
Pending small change in the documentation only

docs/walkthrough.rst
albertohernandez1995: It seems like in github in github is showcasing the result wrongly. My pycharm is showing it at the same level.

javihern98 (approved):
All good now, thanks! 😊

PR #120, opened 2025-05-13
Added mapping argument to run_sdmx function

albertohernandez1995:
# Overview

As part of the compatibility implementation with pysdmx, run_sdmx function was updated with a new argument called mappings. This argument will allow  users to map a SDMX Dataflow to a VTL Dataset. 

The function was updated by adding the following conditions:
- If it is None, mapping will be None
- If it is None and we are trying to load more than 1 dataset, it will raise an error.
- If we are providing a dictionary with the mapping, it will be loaded.
- If we are loading by using pysdmx VTLDataflowMapping object, it will be loaded. Currently, to_vtl_mapping_method and from_vtl_mapping_method are not implemented yet, so a warning will raise if both of them have some information
- typing errors were also added, according to the typing codified in pysdmx library.
- Finally, dataset's name will be provided from the schema´s id if we are using the mapping argument.

### DIFF
--- a/src/vtlengine/API/__init__.py
+++ b/src/vtlengine/API/__init__.py
@@ -1,3 +1,4 @@
+import warnings
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Sequence, Union
 
@@ -7,6 +8,7 @@ from antlr4.error.ErrorListener import ErrorListener  # type: ignore[import-unty
 from pysdmx.io.pd import PandasDataset
 from pysdmx.model import TransformationScheme
 from pysdmx.model.dataflow import Schema
+from pysdmx.model.vtl import VtlDataflowMapping
 
 from vtlengine.API._InternalApi import (
     _check_output_folder,
@@ -321,6 +323,7 @@ def run(
 def run_sdmx(
     script: Union[str, TransformationScheme, Path],
     datasets: Sequence[PandasDataset],
+    mappings: Optional[Union[VtlDataflowMapping, Dict[str, str]]] = None,
     value_domains: Optional[Union[Dict[str, Any], Path]] = None,
     external_routines: Optional[Union[str, Path]] = None,
     time_period_output_format: str = "vtl",
@@ -359,6 +362,8 @@ def run_sdmx(
 
         datasets: A list of PandasDataset.
 
+        mappings: A dictionary or VtlDataflowMapping object that maps the dataset names.
+
         value_domains: Dict or Path of the value domains JSON files. (default:None)
 
         external_routines: String or Path of the external routines SQL files. (default: None)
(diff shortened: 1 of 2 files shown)


src/vtlengine/API/__init__.py
```diff
@@ -379,6 +384,26 @@ def run_sdmx(
         SemanticError: If any dataset does not contain a valid `Schema` instance as its structure.
 
     """
+    mapping_dict = {}
+
+    if mappings is None:
+        if len(datasets) != 1:
+            raise ValueError("If no mappings are provided, only one dataset is allowed.")
```
javihern98, line 391: We only handle here if datasets != 1, but if datasets = 1, we need to add a new internal method called "extract_single_input_dataset", that will analyse the script and check the global inputs is only one (using the DAG Analyzer) and extract the dataset name of that global input, to add it to the mapping.

src/vtlengine/API/__init__.py
```diff
@@ -387,7 +412,8 @@ def run_sdmx(
             raise SemanticError("0-3-1-2", schema=schema)
         vtl_structure = to_vtl_json(schema)
         data_structures.append(vtl_structure)
-        datapoints[schema.id] = dataset.data
+        dataset_name = mapping_dict.get(schema.id, schema.id)
+        datapoints[dataset_name] = dataset.data
```
javihern98, line 416: Before sending it to the run function, we need as well to check that all datasets present in the mapping have been found. Later on in the process it will be checked that we have all datasets present that we need.

src/vtlengine/API/__init__.py
```diff
@@ -379,6 +384,26 @@ def run_sdmx(
...
+    if mappings is None:
+        if len(datasets) != 1:
+            raise ValueError("If no mappings are provided, only one dataset is allowed.")
+    elif isinstance(mappings, Dict):
+        mapping_dict = mappings
+    elif isinstance(mappings, VtlDataflowMapping):
+        if mappings.to_vtl_mapping_method is not None:
+            warnings.warn("This method is not implemented yet.")
+        if mappings.from_vtl_mapping_method is not None:
+            warnings.warn("This method is not implemented yet.")
```
javihern98, line 398: Need to improve the warning message (to_vtl_mapping_method/from_vtl_mapping_method mapping method is not implemented yet, we will use the Basic method)

albertohernandez1995: Requested change was implemented.

src/vtlengine/API/__init__.py
```diff
@@ -387,7 +412,8 @@ def run_sdmx(
             raise SemanticError("0-3-1-2", schema=schema)
         vtl_structure = to_vtl_json(schema)
         data_structures.append(vtl_structure)
-        datapoints[schema.id] = dataset.data
+        dataset_name = mapping_dict.get(schema.id, schema.id)
```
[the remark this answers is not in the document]
albertohernandez1995: At the moment, I changed only the key item of the dictionary to schema.short_urn. Further investigation and discussion is needed to implement it correctly.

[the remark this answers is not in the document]
javihern98: Need to await for pysdmx 1.3.0 to address this correctly

PR #128, opened 2025-05-21
Add qualifier for numpy and updated pysdmx minimum version

javihern98:
- Added qualifier for numpy on python < 3.13
- Updated pysdmx minimum to 1.2.0


mla0303 (approved):
Looks good. Also where the problem was for not specifying Numpy for lower versions of Python.

mla2001 (approved):

PR #133, opened 2025-05-22
Prevent use of VTL scripts with duplicated output datasets

albertohernandez1995:
There was a bug that allowed the load of the following script:
```python

script = TransformationScheme(
        id="TS1",
        version="1.0",
        agency="MD",
        vtl_version="2.1",
        items=[
            Transformation(
                id="T1",
                uri=None,
                urn=None,
                name=None,
                description=None,
                expression="DSD_1 [calc Me_4 := OBS_VALUE];",
                is_persistent=False,
                result="DS_r",
                annotations=(),
            ),
            Transformation(
                id="T2",
                uri=None,
                urn=None,
                name=None,
                description=None,
                expression="DSD_1 [rename OBS_VALUE to Me_5];",
                is_persistent=False,
                result="DS_r",
                annotations=(),
            ),
        ],
    )
```
As you can notice, there are two DS_r with the same name. This is wrong and should not be allowed.
To solve this problem, I added an exception in DAG Analyzer, specifically in _ds_usage_analysis method. Now, when an output dataset with same name is intended to be loaded as one that is already in the all_output list, a ValueError will rise.
Testing was also added, using the same script the bug was detected and using another script to ensure it wasn't a specific case


albertohernandez1995:
Still need to study the behaviour, moved to in progress

src/vtlengine/AST/DAG/__init__.py
```diff
@@ -100,6 +100,10 @@ def _ds_usage_analysis(self):
             else:
                 statements[DELETE][deletion_key] = reference
 
+        duplicates = [x for x in all_output if all_output.count(x) > 1]
+        if duplicates:
+            raise ValueError("There are two or more output datasets with the same name.")
+
```
jfrancma, line 106: Seems fine to me, the only thing I would add would be to show the duplicates in the error message so the user knows exactly which one is wrong.

jfrancma (commented):
Seems all good, just one little suggestion

src/vtlengine/AST/DAG/__init__.py
albertohernandez1995: Added f string to show what output datasets are duplicated

jfrancma (approved):
All good

PR #135, opened 2025-05-23
Prevent use of VTL scripts with duplicated output datasets #2. 

albertohernandez1995:
The first approach to solve this issue was wrong because the error must be risen before the ds_usage_analysis. Code was refactored and now the error will be raised when AST is created.

Testing was also added


jfrancma (approved):
Seems good

tests/AST/test_AST.py
```diff
@@ -42,6 +42,48 @@
...
+        """
+        define operator suma (ds1 dataset, ds2 dataset)
+            returns dataset is
+            ds1 + ds2
+        end operator;
+    """,
+        "MD",
+        "1.0",
+    ),
+]
```
jfrancma, line 79: Why is this here? it dont seems to be used in any test

jfrancma (changes requested):
Just a small change

tests/AST/test_AST.py
albertohernandez1995: It was old code. I have deleted it

jfrancma (approved):
Seems all good

PR #143, opened 2025-05-28
Fixed ASTString not keeping simple commas

javihern98:
- Changes to ensure ASTString is keeping simple commas in all tokens
- Fixed prettify generation
- Added tests for all operators and ensure ASTString and prettify are generating syntactically valid VTL Scripts


albertohernandez1995 (approved):
Seems nice! Well done!

PR #155, opened 2025-06-08
Created Virtual Counter Asset

albertohernandez1995:
As part of the investigation in Datasets Larger than the memory, an approach has been done by creating a VirtualCounter to identify the creation of VirtualDatasets and Virtual DataComponents within the memory. This Virtual objects may disappear when the Transformation is created. This Counter is as follows:
```python
 class VirtualCounter:
    _instance = None
    dataset_count: int = 0
    component_count: int = 0

    def __init__(self) -> None:
        self.dataset_count = 0
        self.component_count = 0

    def __new__(cls):  # type: ignore[no-untyped-def]
        if cls._instance is None:
            cls._instance = super(VirtualCounter, cls).__new__(cls)
            cls._instance.reset()
        return cls._instance

    @classmethod
    def reset(cls) -> None:
        cls.dataset_count = 0
        cls.component_count = 0

    @classmethod
    def _new_ds_name(cls) -> str:
        cls.dataset_count += 1
        name = f"@VDS_{copy(cls.dataset_count)}"
        return name

    @classmethod
    def _new_dc_name(cls) -> str:
        cls.component_count += 1
        name = f"@VDC_{copy(cls.component_count)}"
        return name
```
As you can see, is a singleton. This is important because it may be restarted when the process is finished. To ensure it is reseted, there is a reset method that helps us to manage when we want the counter to be at 0. This is the case on regular aggregations, which structure affects the use of the VDS counter by adding one unit. This is solved by using the mentioned method.
Finally, the counter has been applied in all operators except Hierarchical, which is not necessary.
Testing has been also done.


javihern98 (approved):
Looks good, thanks! 😊

PR #169, opened 2025-06-26
Scalar Values were imported to run function.

albertohernandez1995:
This update introduces full support for handling scalars in the run() function. The following tasks were completed and verified:

## Functional Implementation
- Added a scalar_values parameter to run(), accepting int, float, str, bool, or None.

- Scalars are checked against global_inputs and validated to not conflict with dataset names.

- Scalars are handled independently from datasets without passing in ds_analysis.

- Scalars are returned as Scalar objects in run_result, just like datasets.

## Scalars are saved as .csv files when output_folder is provided:

1. One value per file, no header.
2.  Empty file if the scalar value is None.
3. Output scalars are included in final results and saved along with datasets.

Testing was also added.


javihern98 (approved):
Looks good now, thanks! 😊

PR #199, opened 2025-07-18
Cr 182

mla2001:


mla2001:
Summary of changes:

- Implemented Join operators.
- Fixed the operator Calc rearranging NULL values to the end of the column.
- Exception assertion failing tests are now shown as ignored.
- Updated custom method loading.  Now its done only after connection creation ensuring its done whener connection its called.
- Added clean_execution_graph methods to avoid memory overload.
- Fixed fillna drops columns in some ops.
- Upadted aggregation.
- Fixed general errors.

javihern98 (approved):

PR #203, opened 2025-07-28
Cr 188. Duck_db time operator implementation.

albertohernandez1995:
DuckDB time operators such as year, month, dayofmonth, dayofyear, yeartoday, monthtoday, daytoyear, daytomonth, time_agg, period_ind, date_add, and date_diff were implemented.

The main goal was to create custom DuckDB functions to replicate these operators as SQL expressions whenever they are called. Additionally, the evaluation methods of these operators were updated to support DuckDB relations and SQL queries. 
Evidence of these updates is provided by the following testing results:

- Unary time operators:
image

- Date_add:
image

- Date_diff:
image

- Time_agg:
image

-Period_indicator:
image

**Note**: The are some time_agg test that failed because they were involved in more complex operations such as group all or sum that at the moment are not implemented. Same with period_indicator, were the data type result was changed so it can pass to check if the operand was working. This is because this data type is Duration, and at the moment it is not implemented as intended.  That is why there is a TODO mark in that test and RM099 test is failing.


tests/Additional/data/DataStructure/output/7-1-DS_r.json
```diff
@@ -24,7 +24,7 @@
         {
           "name": "duration_var",
           "role": "Measure",
-          "type": "Duration",
+          "type": "String",
```
javihern98, line 27: Do not change this DataType (I guess is from PeriodIndicator)

tests/Additional/test_additional.py
```diff
@@ -3439,6 +3439,7 @@ def test_1(self):
         """
         Basic behaviour for datasets.
         """
+        # TODO: Change data structure's data type result to Duration when it is implemented. At the moment, it is set to String.
```
javihern98, line 3442: Remove this TODO, we need to analyze Duration datatype separatedly, at the moment we need to adapt to duckdb, not change any data type

javihern98 (changes requested):
Review the period indicator test

src/vtlengine/Operators/__init__.py
```diff
@@ -129,6 +129,9 @@ def apply_bin_op_scalar(cls: Type["Binary"], left: Any, right: Any) -> Any:
     left = handle_sql_scalar(left)
     right = handle_sql_scalar(right)
 
+    # if cls.op == DATEDIFF:
+    #     query = f"{op_token}({left}, {right})"
+    # else:
```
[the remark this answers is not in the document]
albertohernandez1995: Yes, it was a fragment of my first approach to the operator. My bad. It should be clear now.

jfrancma (approved):
Looks good to me.

PR #205, opened 2025-07-29
Fix if then else with explicit nulls

mla2001:


javihern98 (approved):
Looks good, thanks! 😊

PR #212, opened 2025-08-04
Implement Validation operators

mla2001:
Reference manual related tests:
image

Valiadtion related tests (the others are related to HR Ops)
image

Additional Check validation tests:
image
image
image

### DIFF
--- a/src/vtlengine/duckdb/custom_functions/Numeric.py
+++ b/src/vtlengine/duckdb/custom_functions/Numeric.py
@@ -39,7 +39,7 @@ def round_duck(value: Optional[Union[int, float]], decimals: Optional[int]) -> O
 def count_decimals(x: Optional[float]) -> int:
     if x is None:
         return 0
-    d = Decimal(x).normalize()
+    d = Decimal(str(x)).normalize()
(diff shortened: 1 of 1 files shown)


src/vtlengine/duckdb/custom_functions/Numeric.py
```diff
@@ -35,6 +35,33 @@ def round_duck(value: Optional[Union[int, float]], decimals: Optional[int]) -> O
     return int(rounded_value)
 
 
+def count_decimals(val: Optional[Union[int, float]]) -> int:
+    s = f"{val:.16f}".rstrip("0")
+    return len(s.split(".")[1]) if "." in s else 0
```
javihern98, line 40: We cannot use here string representation, we need to look for other way. The input data type should always be float

javihern98 (changes requested):
Minor change needed

src/vtlengine/duckdb/custom_functions/Numeric.py
mla2001: Remade using Decimals.

javihern98 (approved):
Looks good now, thanks! 😊

PR #219, opened 2025-08-07
Fix nullability on conditional operators

javihern98:
## Summary of changes
- Changed nullability to False on if-then-else and case when we have two constants not-null
- (Out of scope) Added main.py to excluded from ruff


mla2001 (approved):
Looks fine!

PR #229, opened 2025-08-26
Implement duplicate identifiers check

mla2001:
The  duplicate identifiers check was implemented, raising an exception whenever a group of duplicate identifiers is found in the same dataset.
Fixed tests with duplicate identifiers.


javihern98 (approved):
Looks good, thanks! 😊 Note the big amount of changes is because of adding a new id in these big csv files

PR #231, opened 2025-08-26
Test suite for duck db Performance

### LINKED ISSUE #221: (Duckdb) Testing single transformations with large data
Testing over the following conditions:

- Data size must be 5GB, then 10GB (per csv file)
- Use WSL to test memory limit, need to change variable BASE_MEMORY_LIMIT in connection/connection.py. Values must be 512MB, 1GB, 2GB, 4GB
- Test the following operators and show evidences in this issue:
  - Basic operators (Binary and Unary from Numeric, Boolean, Comparison, String)
  - Aggregation operators (sum, count)
  - Join operators
  - Analytic operators
    - With window (1 preceding and unbounded following, unbounded preceding and 1 following, unbounded preceding and current row)
    - Without window

Evidences to add
- Screenshots from htop command
- Screenshot of `ls` command in DUCKDB_TEMP_FOLDER (suggested to set it to Path(__file__).parent.parent.parent.parent/".tmp" on connection.py)

albertohernandez1995:
Close #221 

As part of the implementation of duck_db it was necessary to create a test suite so we can get an overall view of the state of the update.

# Major changes:
## File generator
Firstly, a file generator has been include so it can be possible to generate big csv files with random data. The script works as it follows:
- Firstly, you must set the number of components you want the csv to have an its type of data that will hold (e.g Id_1, Integer, Me_1, string)
- Secondly, you must set the length of the rows you wish the csv to have. (e.g. 2_000_000)
- Finally, you can set the number of rows that will be written at the same time by setting as you desire de chunk size param

## Test Memory Usage
This file is a performance testing harness for VTL scripts. It executes transformations on large datasets while tracking execution time and memory usage. The goal is to evaluate scalability and resource consumption of the VTL engine.

### How it works
- Setup: defines input/output directories, configures DuckDB logs, and prepares a results CSV file.
- Memory monitoring: the `MemAnalyzer` class runs in a background thread, sampling the process memory (RSS), capturing peak usage, and storing a timeline of memory consumption.
- Test execution: For each VTL script with its input CSV/JSON datasets, a test is executed under a configurable memory limit. A live timer reports memory usage while the script runs, which is going to be collecting the execution time and the memory metrics. Finally, a CSV file called `test_results` is created with the information collected and the id of the memory series, timeline and png created. A JSON file called log will be also created with more detail information such as the latency or the cpu time.


pyproject.toml
```diff
@@ -35,6 +35,8 @@ dependencies = [
     "networkx>=2.8,<3.0",
     "numpy>=1.23.2,<2 ; python_version < '3.13'",
     "numpy>=2.1.0; python_version >= '3.13'",
+    "psutil (>=7.0.0,<8.0.0)",
+    "seaborn (>=0.13.2,<0.14.0)",
```
javihern98, line 39: These two dependencies must be changed to dev dependencies

albertohernandez1995: Resolved

PR #240, opened 2025-09-10
Cr 196 - Handle of floating point values and rounding

javierdc200:
# Summary of changes
- Implemented rounding based on an environment variable to solve floating point precision issues  
- Refactored some tests that had incorrect expected results  

# Changes out of scope
- Refactor of `random_duck` function to handle null values


javihern98 (approved):
Looks good 😊

PR #248, opened 2025-09-17
(Duckdb) Removed s3fs dependency and secure handling with httpfs extension

albertohernandez1995:
# Changes included
Updated ConnectionManager to enforce safe defaults:

1. SET autoinstall_known_extensions = false;
2. SET autoload_known_extensions = false;
3. SET lock_configuration = true;

Extended the configuration to allow controlling these flags via configure().

# check_extra_S3:
This extra util was refactored to support httpfs extension and remove the S3 dependency. As a result, it has now a new flag that allows the installation of the extension if it is not installed in local and secures the loading of S3 via httpfs extension.


src/vtlengine/API/_InternalApi.py
```diff
@@ -120,17 +120,17 @@ def _load_single_datapoint(datapoint: Union[str, Path]) -> Dict[str, Any]:
     Returns a dict with the data given from one dataset.
     """
     if not isinstance(datapoint, (Path, str)):
-        raise Exception("Invalid datapoint. Input must be a Path or an S3 URI")
+        raise Exception("Invalid datapoint. Input must be a Path or an S3/HTTP(s) URL.")
     if isinstance(datapoint, str):
         if datapoint.startswith(("http:/", "https:/", "s3:/")):
-            # __check_s3_extra()
+            __check_s3_extra(allow_installation=True)
```
javihern98, line 126: The allowance of installation must be handled using an environment variable, called "PROD_MODE", with default value as False (which allows the installation)

src/vtlengine/__extras_check.py
```diff
@@ -9,9 +12,13 @@
 )
 
 
-def __check_s3_extra() -> None:
-    package_loc = importlib.util.find_spec("s3fs")
-    if package_loc is None:
-        raise ImportError(
-            ERROR_MESSAGE.format(extra_name="s3", extra_desc="over csv files using S3 URIs")
-        ) from None
+def __check_s3_extra(allow_installation: bool = True) -> None:
```
javihern98, line 15: Need to handle it here as well with the environment variable

javihern98 (changes requested):
Looks good, but an environment variable is missing

src/vtlengine/API/_InternalApi.py
albertohernandez1995: Adapted code as requested

PR #259, opened 2025-09-26
Solved bug with errorlevel

albertohernandez1995:
Previously, errorlevel was always considered an integer. Now, it can accepts string values.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2215,31 +2215,31 @@ files = [
 
 [[package]]
 name = "ruff"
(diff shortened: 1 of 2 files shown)


albertohernandez1995:
The VTL Reference Manual doesn't specify what type the errorlevel parameter should be. Based on this, I solved your problem by changing the typo to Optional[Union[int,str]] for all errorlevel parameters so that it can handle both types and, when omitted, accept null as a value.

PR #262, opened 2025-10-01
Time operator only usable with EQ and NEQ operators

mla2001:
## Changelog

- Time comparison operations <, <=, <, >= are now prohibited (added error related messages)
- Max and Min operations with Time operands are now forbidden too (added related error message)

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2557,14 +2557,14 @@ files = [
 
 [[package]]
 name = "types-jsonschema"
-version = "4.25.1.20250822"
+version = "4.25.1.20251009"
 description = "Typing stubs for jsonschema"
 optional = false
 python-versions = ">=3.9"
 groups = ["dev"]
 files = [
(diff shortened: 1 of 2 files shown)


src/vtlengine/Exceptions/messages.py
```diff
@@ -53,6 +53,7 @@
     "1-1-1-2": "Invalid implicit cast from {type_1} and {type_2} to {type_check}.",
     "1-1-1-3": "At op {op}: {entity} {name} cannot be promoted to {target_type}.",
     "1-1-1-4": "At op {op}: Operation not allowed for multimeasure datasets.",
+    "1-1-1-5": "At op {op}: Invalid type {type}.",
```
javihern98, line 56: We need a special error for this, to clarify that Time comparison is not supported except for EQ and NEQ. Please check as well the Aggregation operators max/min

javihern98 (changes requested):

src/vtlengine/Exceptions/messages.py
mla2001: Added related error message.

mla2001:
> Missing here 2 tests, one for Aggregation and another with binary operators

Added related tests

javihern98 (approved):
Looks good now, thanks! 😊

PR #304, opened 2025-11-03
Refactored VD and Ext_Routines semantic_analysis signature

albertohernandez1995:
Now, it has the same signature as run function.

```def semantic_analysis(
    script: Union[str, TransformationScheme, Path],
    data_structures: Union[Dict[str, Any], Path, List[Dict[str, Any]], List[Path]],
    value_domains: Optional[Union[Dict[str, Any], Path, List[Union[Dict[str, Any], Path]]]] = None,
    external_routines: Optional[
        Union[Dict[str, Any], Path, List[Union[Dict[str, Any], Path]]]
    ] = None,
) -> Dict[str, Dataset]:
```

```
def run(
    script: Union[str, TransformationScheme, Path],
    data_structures: Union[Dict[str, Any], Path, List[Dict[str, Any]], List[Path]],
    datapoints: Union[Dict[str, pd.DataFrame], str, Path, List[Dict[str, Any]], List[Path]],
    value_domains: Optional[Union[Dict[str, Any], Path, List[Union[Dict[str, Any], Path]]]] = None,
    external_routines: Optional[
        Union[Dict[str, Any], Path, List[Union[Dict[str, Any], Path]]]
    ] = None,
    time_period_output_format: str = "vtl",
    return_only_persistent: bool = True,
    output_folder: Optional[Union[str, Path]] = None,
    scalar_values: Optional[Dict[str, Optional[Union[int, str, bool, float]]]] = None,
) -> Dict[str, Union[Dataset, Scalar]]:
```

### DIFF
--- a/src/vtlengine/API/__init__.py
+++ b/src/vtlengine/API/__init__.py
@@ -14,6 +14,7 @@ from pysdmx.util import parse_urn
 from vtlengine.API._InternalApi import (
     _check_output_folder,
     _check_script,
+    _get_persistence,
     _return_only_persistent_datasets,
     ast_to_sdmx,
     load_datasets,
@@ -360,6 +361,7 @@ def run(
                 format_time_period_external_representation(obj, time_period_representation)
 
     # Returning only persistent datasets
+    result = _get_persistence(result, ast)
     if return_only_persistent:
         return _return_only_persistent_datasets(result, ast)
     return result
--- a/src/vtlengine/API/_InternalApi.py
+++ b/src/vtlengine/API/_InternalApi.py
@@ -437,17 +437,31 @@ def load_external_routines(
     return external_routines
 
 
+def _get_persistence(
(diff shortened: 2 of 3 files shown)


src/vtlengine/API/__init__.py
```diff
@@ -194,10 +196,20 @@ def semantic_analysis(
...
+            vd = load_value_domains(value_domains)
     ext_routines = None
     if external_routines is not None:
-        ext_routines = load_external_routines(external_routines)
+        if isinstance(external_routines, list):
+            ext_routines = {}
+            for item in external_routines:
+                ext_routines.update(load_external_routines(item))
+        else:
+            ext_routines = load_external_routines(external_routines)
```
javihern98: Refactor this code to leave it as before but with this logic inside the loading function, both in semantic and run

javihern98 (commented):
A test is missing here with a list of value domains and external routines when calling the semantic analysis, please add it @albertohernandez1995

PR #310, opened 2025-11-04
Add persistent attribute to Dataset and Scalar

### LINKED ISSUE #307: Add persistent flag to Scalar and Dataset object.
Actually, we filter between persistent and temporary assignments just by checking whether the result names AST nodes are PersistentAssignments or not in the [_return_only_persistent_datasets](https://github.com/Meaningful-Data/vtlengine/blob/ded7634d5de445db7671f206443d5be33f29b98c/src/vtlengine/API/_InternalApi.py#L423) method.
In order to differentiate them after the execution, we should add a '**persistent**' flag into the [Dataset](https://github.com/Meaningful-Data/vtlengine/blob/ded7634d5de445db7671f206443d5be33f29b98c/src/vtlengine/Model/__init__.py#L177) and [Scalar](https://github.com/Meaningful-Data/vtlengine/blob/ded7634d5de445db7671f206443d5be33f29b98c/src/vtlengine/Model/__init__.py#L23) objects.

mla2001:
Closes #307 

## Changelog

- A **persistent** flag was added to **Dataset** and **Scalar** objects. 
  - Now the resulting Datasets and Scalars can be distinguished without using the AST.
- Added copy into VarID return statements to avoid reference overwriting.

### DIFF
--- a/src/vtlengine/API/_InternalApi.py
+++ b/src/vtlengine/API/_InternalApi.py
@@ -420,6 +420,24 @@ def load_external_routines(input: Union[Dict[str, Any], Path, str]) -> Any:
     return external_routines
 
 
+def _get_persistence(
+    datasets: Dict[str, Union[Dataset, Scalar]], ast: Start
+) -> Dict[str, Union[Dataset, Scalar]]:
+    """
+    Set the persistence attribute to the datasets according to the AST.
+    """
+    persistent = []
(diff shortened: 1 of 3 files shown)


albertohernandez1995 (approved):
Seems good! Congrats!

javihern98:
Will be added in #314

PR #331, opened 2025-11-07
Added handling of VD paths when they are input from a list.

### LINKED ISSUE #330: Accept value domains and external routines loading by list of paths
Currently, if a list of paths is passed as an argument to the load_value_domains method, it raises the following [exception](https://github.com/Meaningful-Data/vtlengine/blob/22fdcc261cee56a4bd301427ab8a24da6ca5c2a7/src/vtlengine/API/_InternalApi.py#L400). 
`Invalid vd file. Input is not a Path object`

It would be good to load them via a List, as we do with data_structures and datapoints.
For reference, see the [load_datasets](https://github.com/Meaningful-Data/vtlengine/blob/22fdcc261cee56a4bd301427ab8a24da6ca5c2a7/src/vtlengine/API/_InternalApi.py#L235) method.

### Code suggestion
```
if isinstance(input, list):
    value_domains: Dict[str, Any] = {}
    for x in input:
        vd = load_value_domains(x)
        value_domains = {**value_domains, **vd}
    return value_domains
```

albertohernandez1995:
# Major changes:
- Added a patch to load_value_domains method to ensure it can load multiple paths holded in an input list.
- Added test to ensure it works as intended.
---

Closes #330

### DIFF
--- a/src/vtlengine/API/_InternalApi.py
+++ b/src/vtlengine/API/_InternalApi.py
@@ -401,7 +401,7 @@ def load_value_domains(
     if not input.exists():
         raise Exception("Invalid vd file. Input does not exist")
     if input.is_dir():
-        value_domains: Dict[str, Any] = {}
+        value_domains = {}
         for f in input.iterdir():
             vd = _load_single_value_domain(f)
             value_domains = {**value_domains, **vd}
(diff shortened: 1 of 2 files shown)


mla2001 (approved):
Looks fine!

PR #334, opened 2025-11-10
Fix ASTString prettyprint over Hierarchical Ruleset with more than one rule and condition

### LINKED ISSUE #333: Fix ASTString prettyprint HRuleset

mla2001:
Closes #333 

## Changelog
- Fixed ASTTempalte HRuleset couldn't handle node.element as a list (HRuleset with conditions)
- Fixed that ASTString HRuleset and Check_Hierarchy could not handle the conditions in HRuleset. 
- Fixed missing ';' separator between HRules when pretty print was active. 
- Fixed some indentation issues with pretty print active.
- Added related test.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "vtlengine"
-version = "1.3.0rc4"
+version = "1.3.0rc5"
 description = "Run and Validate VTL Scripts"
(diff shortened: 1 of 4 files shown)


javihern98:
Missing specific tests for this in the AST String (and the DP Rulesets one) @mla2001

PR #339, opened 2025-11-11
Fixed DAG input collector

### LINKED ISSUE #338: Fix DAG input collector
Currently, **DAG ds_structure** method sometimes store some Components as inputs ([`global_inptus`](https://github.com/Meaningful-Data/vtlengine/blob/760e41e0c4b5ffb9c55a84f480fdaa7e1e567626/src/vtlengine/AST/DAG/__init__.py#L90C9-L90C22)). 
Check in which cases this happens and fix them.
- As it could be seen, it may be related to the input components of an HRuleset.

mla2001:
Closes #338 

## Changelog
- Added internal DAG method visit_UDOCall to avoid collecting Components as global inputs when an UDO is the Aggregation-Dataset of a RegularAggregation operand.
- Added related tests.


albertohernandez1995 (approved):
Looks good!! Congrats!

PR #364, opened 2025-11-19
Fixed wrong parameter in UDO causes NotImplementedError instead of syntax error

### LINKED ISSUE #353: Wrong parameter in UDO causes NotImplementedError instead of syntax error
This code cause a `NotImplementedError` in create_ast function.

```
define operator filRoundEr (ds dattt, comp component, margin number)
  returns dataset is
    ds[filter comp > margin or comp < -margin]
end operator;
```

This should be catched as a Syntax error as the parameter type of ds is wrong (should be `dataset`)

mla2001:
Closes #353 

## Changelog

- Updated wrong parameter handler. 

Previously, whenever the parameter type did not match any of the expected ('dataset', 'component'...), the constructor assumed it was a Value Domain, raising a NotImplementedError. It has now been updated with a way more descriptive SyntaxError.

### DIFF
--- a/src/vtlengine/AST/ASTConstructorModules/Terminals.py
+++ b/src/vtlengine/AST/ASTConstructorModules/Terminals.py
@@ -355,8 +355,8 @@ class Terminals(VtlVisitor):
 
         else:
             raise SyntaxError(
-                f"Invalid type definition {scalartype.children[0]} at line "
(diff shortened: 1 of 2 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #404, opened 2025-12-17
Refactored Duration Data type

### LINKED ISSUE #394: Remove use of ISO 8601 in Duration to align with the current version of the standard
Revert duration values changes in #55 to ensure we can only use the values defined in [behaviour of time operators](https://sdmx-twg.github.io/vtl/2.1/html/reference_manual/typical_behaviour.html#behaviour-of-time-operators)

Image

As there is an inconsistency over the use of the Duration data type, we will use the Duration values specified and use String data types as arguments and result types of the Time Operators that use it

Check the following operators and related tests to ensure the Duration is applied correctly:

- The result of `period_indicator` has a Duration data type
- The input of `time_agg` supports both String and Duration
- The check_duration function checks that the values of the Duration data type are in the `PERIOD_INDICATOR_MAPPING` dict, and that those are the ones specified above
- The Duration data type comparison is perform adequately using the `PERIOD_INDICATOR_MAPPING` dict
- The return data type of the new operators `daytoyear` and `daytomonth` is String
- The input data type of the new operators `yeartoday` and `monthtoday` is String, a regex will be used to ensure it matches an ISO 8601 Duration

Relevant operators links on VTL docs:
- [Days to duration](https://sdmx-twg.github.io/vtl/2.1/html/reference_manual/operators/Time%20operators/Number%20days%20to%20duration/index.html#)
- [Duration to number](https://sdmx-twg.github.io/vtl/2.1/html/reference_manual/operators/Time%20operators/Duration%20to%20number%20days/index.html)
- [Period indicator](https://sdmx-twg.github.io/vtl/2.1/html/reference_manual/operators/Time%20operators/Period%20indicator/index.html)

albertohernandez1995:
## Major Changes
- Refactored Duration Data Type by rollbacking it to the previous state, without the ISO 8601 check.
- Refactored related test.
- Moved to_days method to Unary main Datatype to keep supporting the ISO 8601 check at new time unary operators.
- Refactored new time operators as requested in the issue.

> [!NOTE]
>  When Duration data type is clearly defined, we should reuse the comment exception tests in typechecking / string module. They don't raise up right now because they are not longer exceptions and its values are in fact Duration values.

---
closes #394

### DIFF
--- a/tests/Additional/test_additional_scalars.py
+++ b/tests/Additional/test_additional_scalars.py
@@ -5,11 +5,12 @@ import pandas as pd
 import pytest
 
 from tests.Helper import TestHelper
+from vtlengine import DataTypes
 from vtlengine.API import create_ast, run
 from vtlengine.DataTypes import Boolean, Integer, Null, Number, String
 from vtlengine.Exceptions import SemanticError
 from vtlengine.Interpreter import InterpreterAnalyzer
-from vtlengine.Model import Scalar
+from vtlengine.Model import Component, Dataset, Role, Scalar
 
 
 class AdditionalScalarsTests(TestHelper):
@@ -230,7 +231,6 @@ ds_param = [
     ("13-9", "DS_1[aggr attribute Me_2 := sum(Me_1) group by Id_1]"),
 ]
 
-
 params_scalar_operations = [
     ("Sc_r <- sc_1 + sc_2 + 3 + sc_3;", {"Sc_r": Scalar(name="Sc_r", data_type=Integer, value=21)}),
     (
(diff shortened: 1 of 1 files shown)


src/vtlengine/Operators/__init__.py
```diff
@@ -898,3 +899,24 @@ def component_evaluation(cls, operand: DataComponent) -> DataComponent:
             operand.data.copy() if operand.data is not None else pd.Series()
         )
         return result_component
+
+    @classmethod
+    def to_days(cls, value: Any) -> int:
```
javihern98: Typing of value has to be str

javihern98 (commented):
Small changes are needed, but overall looks good. I suggest to add as well tests using filter and a Scalar of type Duration with any of the Duration values and None (in test_additional_scalars)

PR #408, opened 2025-12-18
Fix create_ast_with_comments method crash when only comments script is passed

mla2001:
## Changelog
- Added check to avoid calling create_ast method inside create_ast_with_comments method when only comments and white spaces are passed.


javihern98 (dismissed):
Looks good, thanks!😊

javihern98:
@mla2001 Please update this branch

PR #439, opened 2026-01-13
Added conf parameter in ASTString's visit Time_Aggregation method

### LINKED ISSUE #430: Bug on AST String rendering with time_agg and first/last parameter
# Overview

Using this script:

```python
    from vtlengine import create_ast
    from vtlengine.AST.ASTString import ASTString

    script = """
        DS_A <- time_agg("A", DS_1, last);
    """

    ast = create_ast(script)
    text = ASTString().render(ast)
    print(text)
```

The result is:
```text
DS_A <- time_agg("A", _, DS_1);
```

This should much the `script` value, therefore the AST string rendering is wrong.

albertohernandez1995:
## Major Changes:
- Previously, conf parameter was not handled by this method, so I included it so render method can detect it.
- Added a TS in time.vtl to handle this bug in test_ASTString.

---
Closes #430

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -42,133 +42,133 @@ files = [
 
 [[package]]
 name = "aiohttp"
-version = "3.13.2"
+version = "3.13.3"
 description = "Async http client/server framework (asyncio)"
 optional = true
 python-versions = ">=3.9"
 groups = ["main"]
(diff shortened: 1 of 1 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #484, opened 2026-02-05
Fix #483: Use configurable precision for numeric operations

### LINKED ISSUE #483: Numeric operators round results to 10 significant digits instead of using configured precision
## Description
Numeric operators in VTL are being rounded to 10 significant digits, losing precision compared to standard IEEE 754 double precision (which supports ~15-17 significant digits).

## Minimal Reproducible Example

```python
from vtlengine import run

script = """
    DS_r <- 33998.0 / 6.329830373;
"""

run_result = run(
    script=script,
    data_structures=[],
    datapoints=[],
)

result = run_result["DS_r"].value
expected = 33998.0 / 6.329830373

print(f"VTL result:      {result}")
print(f"Python expected: {expected}")
print(f"Match: {result == expected}")
```

## Current Behavior

```
VTL result:      5371.076
Python expected: 5371.075999922376
Match: False
```

The VTL result is rounded to 10 significant digits (`5371.076000`), losing the decimal precision.

## Expected Behavior

The result should maintain precision consistent with the `COMPARISON_ABSOLUTE_THRESHOLD` configuration (default 15 significant digits):

```
VTL result:      5371.07599992238
Python expected: 5371.075999922376
Match: True (within tolerance)
```

## Root Cause

The `Binary.op_func()` method in the numeric operators was setting the Decimal context precision to a hardcoded value of 10 instead of using the configurable precision from `OUTPUT_NUMBER_SIGNIFICANT_DIGITS` / `COMPARISON_ABSOLUTE_THRESHOLD`.

## Impact

Any division operation (and potentially other binary numeric operations) loses precision beyond 10 significant digits, which can cause:
- Accumulated errors in chained calculations
- Failed equality comparisons when results should match
- Inconsistency between VTL and native Python calculations

javihern98:
## Summary
- Changed Decimal context precision from hardcoded 10 to use `OUTPUT_NUMBER_SIGNIFICANT_DIGITS` (default 15)
- Added `get_effective_numeric_digits()` function for arithmetic operations precision
- Updated documentation to explain the environment variable affects both numeric operations and CSV output
- Added missing error message `0-1-2-6` for scalar not found in datastructures

## Problem

Division operations (and other binary numeric operations) were rounding results to 10 significant digits instead of using the configurable precision.

**Before:**
```
33998.0 / 6.329830373 = 5371.076  (less than 10 significant digits)
```

**After:**
```
33998.0 / 6.329830373 = 5371.07599992238  (15 significant digits)
```

## Out of Scope

- The error message `0-1-2-6` is added but not yet used in the codebase. Implementation of the actual error raising will be addressed in a separate issue.

Closes #483

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "vtlengine"
-version = "1.5.0rc8"
+version = "1.5.0rc9"
 description = "Run and Validate VTL Scripts"
 license = "AGPL-3.0"
 readme = "README.md"
--- a/src/vtlengine/__init__.py
+++ b/src/vtlengine/__init__.py
@@ -24,4 +24,4 @@ __all__ = [
     "validate_external_routine",
 ]
 
-__version__ = "1.5.0rc8"
+__version__ = "1.5.0rc9"


mla2001 (approved):
Looks fine!

PR #517, opened 2026-02-17
Fix #504: Adapt implicit casting to VTL 2.2

### LINKED ISSUE #504: Adapt implicit casting to VTL 2.2
## Summary
Adapt implicit casting to this:

https://sdmx-twg.github.io/vtl/2.2/reference_manual/operators/General%20purpose%20operators/Type%20conversion/index.html

The VTL allows for these implicit casts:

-  From integer to number: a number value is returned having the integer part equal to the integer value, and the decimal part equal to zero.
-  From number to integer: an integer value is returned, representing the sign and the integer part of the input number value; implementations may raise an error if the magnitude of the integer part is too big to be represented internally, provided that they document such a limit.
- From boolean to string: Either one of the string values “true” and “false” is returned, depending on the input. 
- From date to time: a time value is returned having its start equal to its end, and both equal to the input date value.
- From time_period to time: a time value is returned having the stame start and end of the input time_period value.

## Scope
- Files/areas impacted: Casting and operators
- Out of scope: None

## Testing
How will you verify (targeted tests, existing suite, manual checks)? Current tests must work, change relevant tests

javihern98:
## Summary

Restricts implicit casting to the 5 conversions specified by VTL 2.2:

1. **Integer → Number**: returns float with decimal part zero
2. **Number → Integer**: truncates to integer part (sign preserved)
3. **Boolean → String**: returns `"True"` or `"False"`
4. **Date → Time (TimeInterval)**: returns interval `"{date}/{date}"`
5. **TimePeriod → Time (TimeInterval)**: returns interval `"{start}/{end}"`

Previously allowed implicit casts that are now **removed**:
- String ↔ Number/Integer (must use explicit `cast()`)
- String → Boolean (must use explicit `cast()`)

Also fixes a pre-existing bug where **Date vs TimePeriod comparisons** failed because `binary_implicit_promotion` didn't compute the intersection of promotion sets when neither type was directly in the other's set.

### Core type system (`DataTypes/__init__.py`)
- Updated `IMPLICIT_TYPE_PROMOTION_MAPPING` to remove String from Number/Integer sets and vice versa
- Restricted `implicit_cast()` methods (String, TimeInterval, Date, TimePeriod, Duration) to only accept VTL 2.2 allowed source types
- Added `explicit_cast()` handling for types moved from implicit to explicit
- Added intersection-based fallback in `binary_implicit_promotion` for types sharing a common promoted type (Date + TimePeriod → TimeInterval)
- Added same fallback in `check_binary_implicit_promotion`
- Fixed `TimeInterval.implicit_cast` bug where `.isoformat()` was called on an already-string result from `str_period_to_date`

### Runtime casting (`Operators/__init__.py`)
- Updated `dataset_evaluation` to implicitly cast both operands to the common promoted type when their types differ and neither is already the target type (e.g. Date + TimePeriod → TimeInterval)

### Comparison operator (`Operators/Comparison.py`)
- Removed String ↔ Number runtime conversions in `_cast_values()` since those comparisons are now blocked at semantic analysis

### Tests
- Tests that relied on removed implicit casts changed to expect `SemanticError("1-1-1-2")`
- Fixed VTL scripts in AnaVal Monthly, API, and SDMX tests that used invalid implicit casts (e.g. integer literals vs string identifiers)
- Removed `# BUG` comments from Boolean → String tests (now spec-correct)
- Added parametrized tests for Date vs TimePeriod implicit promotion and end-to-end comparison
- Deleted ~250 unused output reference files for tests changed to expect errors
- Updated CC_48 semantic test from error-expected to pass (Date vs TimePeriod is now valid)

## Test plan
- [x] All 3575 tests pass
- [x] ruff format/check clean
- [x] mypy clean
- [x] Verified all 5 VTL 2.2 implicit casts work correctly
- [x] Verified removed casts raise SemanticError
- [x] Verified Date vs TimePeriod comparison works end-to-end (=, <>, both operand orders)
- [x] Verified Integer + Number arithmetic not affected (no truncation)

### DIFF
--- a/src/vtlengine/DataTypes/__init__.py
+++ b/src/vtlengine/DataTypes/__init__.py
@@ -344,8 +344,8 @@ class TimeInterval(ScalarType):
             return f"{value}/{value}"
 
         if from_type in {TimePeriod}:
-            init_value = str_period_to_date(value, start=True).isoformat()
-            end_value = str_period_to_date(value, start=False).isoformat()
+            init_value = str_period_to_date(value, start=True)
+            end_value = str_period_to_date(value, start=False)
             return f"{init_value}/{end_value}"
 
         raise RunTimeError(
@@ -764,6 +764,17 @@ def binary_implicit_promotion(
     if right_type.is_included(left_implicities):
         return right_type
 
+    # Fallback: check if both types can be promoted to a common type
+    # e.g. Date → TimeInterval and TimePeriod → TimeInterval
+    common = left_implicities.intersection(right_implicities)
+    if common:
+        if return_type:
+            return return_type
+        # Return the common promoted type (exclude Null if present)
+        common.discard(Null)
+        if len(common) == 1:
+            return common.pop()
+
     raise SemanticError(
         code="1-1-1-1",
         type_1=SCALAR_TYPES_CLASS_REVERSE[left_type],
@@ -791,7 +802,11 @@ def check_binary_implicit_promotion(
     if type_to_check:
         return type_to_check.is_included(set_=left_implicities.intersection(right_implicities))
 
-    return left.is_included(right_implicities) or right.is_included(left_implicities)
+    return (
+        left.is_included(right_implicities)
+        or right.is_included(left_implicities)
+        or bool(left_implicities.intersection(right_implicities))
+    )
 
 
 def unary_implicit_promotion(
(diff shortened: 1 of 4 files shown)


albertohernandez1995 (approved):
Overall, looks good.

PR #521, opened 2026-02-19
Updated null distinction between empty string and null.

### LINKED ISSUE #512: VTL 2.2 make distinction between null and empty string
## Summary
With the VTL 2.2 changes, we can now make a distinction between a null value and an empty string, to remove the possibility of using exactly the same for these two cases. This issue aims to review the current implementation of this and ensures we apply correctly this distinction in the Operators

## Scope
- Files/areas impacted: Interpreter, Operators
- Out of scope:

## Testing
All existing tests must pass. Possibility of changing some tests from empty string to null

albertohernandez1995:
## Major Changes:
All string operators now propagate null instead of substituting "" when inputs are null, aligning with VTL 2.2 semantics (empty string is a distinct, valid value,  not a substitute for null):

## Operators involved:
lower, upper, trim, ltrim, rtrim, lenght, concat, substr, instr and replace where updated aligning them with VTL2.2 semantics. 

The replace operator required a sentinel value (_REPLACE_PARAM2_OMITTED) to distinguish between an omitted new_string parameter and an explicitly passed null new_string. 

----

Test where updated with these changes. 

Closes #512

### DIFF
--- a/src/vtlengine/DataTypes/TimeHandling.py
+++ b/src/vtlengine/DataTypes/TimeHandling.py
@@ -551,13 +551,16 @@ def check_max_date(str_: Optional[str]) -> Optional[str]:
     if len(str_) == 9 and str_[7] == "-":
         str_ = str_[:-1] + "0" + str_[-1]
 
-    # Format 2010-01-01. Prevent passthrough of other ISO 8601 formats.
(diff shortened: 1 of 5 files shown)


javihern98 (approved):
Looks good now, thanks! 😊

PR #642, opened 2026-04-06
Fix #641: Validate ISO 8601 duration format for yeartoday and monthtoday

### LINKED ISSUE #641: Inconsistent Validation and Undefined Format for yeartoday and monthtoday Operators
### Initial Checks

### Problem Description

# Description
There is an inconsistency in how the yeartoday and monthtoday operators handle period strings. Currently, both operators accept the same input and produce results even when the period units do not match the operator's specific purpose (e.g., passing months to a year-based operator).

# Current Behavior
Inconsistent Runtime Behavior: The operators accept period strings that do not correspond to their specific scope.

```
yt := yeartoday("P2M230D");
dt := monthtoday("P2M230D");
```
At execution time, both operators function identically, failing to enforce the semantic difference between "year" and "month" contexts.

Runtime Crash with Weeks: When the code is executed using weeks in the period string, the following exception is thrown:
'NoneType' object has no attribute 'group'

# Identified Issues

- Lack of Runtime Validation: The operators do not validate at execution time whether the input period is compatible with the operator's intended scope (Years vs. Months).

- Undefined Period Support: It is not clearly defined if combinations like Years/Weeks/Days are valid, leading to unexpected failures.

- Parsing Failure: The NoneType error indicates a runtime failure where the parsing logic (likely a regex) fails to match the string when a week component (W) is present, causing the process to crash when attempting to access a match group.

### Proposed Solution

1. **Implement Runtime Validation:**
Add checks during execution to ensure that the input period string aligns with the operator's scope as specified in the documentation. 
   - `yeartoday()` should only accept periods containing Year (**Y**) and Day (**D**) components.
   - `monthtoday()` should only accept periods containing Month (**M**) and Day (**D**) components.
   - Throw a clear runtime exception if an invalid unit is provided (e.g., passing months to `yeartoday`), ensuring the implementation adheres to the documented behavior.

### Alternatives Considered

Instead of restricting the input, an alternative approach is to allow all valid ISO-8601 duration components while ensuring a stable execution without unexpected crashes.

**Note:**
It is important to note that implementing this unified support would result in both `yeartoday` and `monthtoday` accepting the same inputs and producing the same outputs. Consequently, the two operators would become functionally identical and **redundant**, as the semantic distinction between "year" and "month" contexts would be removed.

1. **Broaden Duration Support:**
Modify the parser to correctly handle all standard ISO-8601 units, including Years (**Y**), Months (**M**), Weeks (**W**), and Days (**D**), regardless of whether the operator is `yeartoday` or `monthtoday`.

2. **Standardize Conversion Logic:**
Define a clear internal logic for how mixed durations are handled to prevent the `'NoneType'` runtime error. This includes:
   - Ensuring the regex/parser recognizes the **W** (Week) attribute.
   - Normalizing the duration internally so that `yeartoday("P1W")` or `monthtoday("P1W")` calculates the correct offset without crashing.

3. **Graceful Error Handling:**
If a duration string is truly malformed (not following ISO-8601), the system should throw a descriptive error message explaining the expected format, rather than a low-level Python attribute error.

4. **Updated Documentation:**
Clearly document that both operators accept full ISO-8601 durations, while clarifying if there are any specific internal behaviors when mixing large units (e.g., how a "Month" is calculated in a `yeartoday` context).

### Additional Context

_No response_

javihern98:
## Summary

Fixes #641

Add per-operator ISO 8601 duration validation for `yeartoday` and `monthtoday` operators:

- `yeartoday` now only accepts `PnYnD` format (years + days)
- `monthtoday` now only accepts `PnMnD` format (months + days)
- Invalid formats raise `RunTimeError("2-1-19-22")` with a descriptive message
- The shared `to_days()` method now has a null-check safety net to prevent `AttributeError` on malformed input (e.g., `"P1W"` previously crashed)
- Fixed test data that used invalid compound durations (`P1Y6M15D` → `P1Y195D`, `P2Y9M21D` → `P9M21D`)

## Checklist

- [x] Tests pass (`pytest`) — 4067 tests pass

## Impact / Risk

- Breaking changes: Yes — `yeartoday` and `monthtoday` now reject durations with components outside their scope (e.g., `yeartoday("P1Y6M15D")` with months). Previously these were silently accepted.
- Data/SDMX compatibility concerns: None
- Notes for release/changelog: New error code `2-1-19-22` for invalid duration format

## Notes

- The `(?=\d)` lookahead in the regex ensures bare `"P"` (no components) is rejected
- Dataset test files already used valid formats and required no changes


mla2001 (approved):
Looks fine! 😊

PR #675, opened 2026-04-28
Fix #674: Add Python 3.14 to CI; remove 3.9 from CI

### LINKED ISSUE #674: Add Python 3.14 to CI matrix; remove Python 3.9 from CI
### Summary

Update the CI test matrix to add Python 3.14 and remove Python 3.9. The `requires-python = ">=3.9"` declaration in `pyproject.toml` is intentionally preserved so users on Python 3.9 can still install the package — they will simply no longer be covered by automated CI tests. While editing `pyproject.toml`, also add the missing `Programming Language :: Python :: 3.X` classifiers so PyPI accurately reflects the supported version range, and set `target-version = "py39"` under `[tool.ruff]` so contributors cannot accidentally introduce 3.10+ syntax that would break the 3.9 install target.

### Scope

**In scope:**

- `.github/workflows/testing.yml` line 21 — replace `"3.9"` with `"3.14"` in the matrix; final list `["3.10", "3.11", "3.12", "3.13", "3.14"]`.
- `pyproject.toml` lines 16–22 — add Python version classifiers (3.9, 3.10, 3.11, 3.12, 3.13, 3.14).
- `pyproject.toml` line 74 — add `target-version = "py39"` under `[tool.ruff]` so lint blocks 3.10+ syntax.
- Fix any 3.10+ syntax violations the new ruff `target-version` surfaces (typically `X | Y` type unions, `match` statements).
- Verify all binary runtime dependencies (pandas, duckdb, pyarrow, numpy) publish cp314 wheels for Linux/macOS/Windows before enabling 3.14 in CI.

**Out of scope:**

- Bumping `requires-python` past `>=3.9` (`pyproject.toml` lines 7 and 52).
- The conditional numpy pin at `pyproject.toml` lines 36–37 — must stay; 3.9 users still need numpy `<2.1`.
- `README.md`, `docs/index.rst`, `CONTRIBUTING.md` wording (the "Python 3.9 or higher" claim is still accurate as install policy).
- Other workflows: `docs.yml`, `version.yml`, `release.yml`, `ubuntu_test_24_04.yml`.
- `bug_report.yml` placeholder.
- Any code changes under `src/` other than ruff-flagged 3.10+ syntax rewrites.

### Testing

- All ruff/mypy/pytest quality checks pass locally on a CI-supported Python.
- CI matrix shows green jobs for Python 3.10, 3.11, 3.12, 3.13, **3.14** across Ubuntu, macOS, Windows after the workflow change.
- Coverage gate (`--fail-under=90`) still passes.
- If a Python 3.14 interpreter is locally available: `poetry env use python3.14 && poetry install --all-extras && poetry run pytest` smoke test.
- `poetry build` + inspect resulting `.tar.gz` `PKG-INFO` to confirm new classifiers are present.

### Additional Context

> [!IMPORTANT]
> Dropping 3.9 from CI while keeping `requires-python = ">=3.9"` means `pip install vtlengine` still succeeds on Python 3.9, but breakage on 3.9 will not be caught by CI. This trade-off is intentional. Bug reports from 3.9 users will be handled reactively.

The conditional numpy pin (`numpy>=2.0.2,<2.1; python_version < '3.10'`) must be preserved because numpy 2.2+ requires Python 3.10+. Removing the conditional would silently break 3.9 installs we still claim to support.

Setting `target-version = "py39"` under `[tool.ruff]` is the principal mitigation for the "CI no longer covers 3.9" gap: ruff will reject 3.10+ syntax (e.g., `X | Y` unions, `match` statements) at lint time, so contributors cannot accidentally land code that breaks 3.9 import. Existing 3.10+ syntax in the codebase, if any, must be rewritten to 3.9-compatible forms as part of this work.

**Pre-flight wheel-availability check (run before enabling 3.14 in CI):**

```bash
for PKG in pandas duckdb pyarrow numpy pysdmx; do
  echo "=== $PKG ==="
  curl -sSL "https://pypi.org/pypi/$PKG/json" \
    | python -c "import sys,json; d=json.load(sys.stdin); v=d['info']['version']; print(f'latest={v}'); [print(f['filename']) for f in d['urls'] if 'cp314' in f['filename']]"
done
```

If any binary dep is missing cp314 wheels for any OS, pause and either wait for upstream or ship the 3.9 drop alone (skipping the 3.14 add).

javihern98:
## Summary

- Swap Python 3.9 for 3.14 in the CI test matrix (`.github/workflows/testing.yml`). `requires-python = ">=3.9"` is preserved on purpose — 3.9 stays installable, just no longer CI-tested.
- Add the missing `Programming Language :: Python :: 3.X` classifiers (3.9–3.14) to `pyproject.toml` so PyPI metadata reflects the supported install range.
- Bump `pyarrow` upper bound from `<20.0` to `<25.0` and `numpy` (3.10+ branch) from `<2.3` to `<2.5` so 3.14 picks versions with cp314 wheels and 3.10–3.13 also pick up the latest. Pip's resolver falls back to older versions where wheels are missing — Python 3.10 still gets `numpy 2.2.6` because numpy 2.3+ dropped cp310.
- Add `target-version = "py39"` under `[tool.ruff]` so 3.10+ syntax (e.g. `X | Y` unions, `match` statements) cannot slip in now that CI no longer covers 3.9.
- **Three Python 3.14 compatibility fixes** surfaced by the first CI run on 3.14 (originally out of scope; folded in here because they all stem from CPython error/behavior changes between 3.13 and 3.14):
  - `AST.__all_annotations` was reading `c.__dict__["__annotations__"]` to walk the MRO. Python 3.14 (PEP 649) defers annotation evaluation, so `__annotations__` may not appear in `__dict__` until first accessed. Result: AST nodes' `__str__` rendered as `<DefIdentifier()>` (empty) and `toJSON` returned only `{"class_name": ...}`. Switched to `inspect.get_annotations` (3.10+) with a 3.9 fallback.
  - `check_date` pattern-matched the upstream `ValueError` text from `date.fromisoformat` (`"is out of range"`). Python 3.14 reworded that text, so `"2014-02-31"` surfaced as a format error instead of "out of range for the month". Replaced the message-string match with our own structural YYYY-MM-DD check, so error classification no longer depends on Python's internal wording.
  - `Logarithm.py_op` was leaking `math.log`'s `ValueError("math domain error")` for `log(0, base)`. Python 3.14 reworded that to `"expected a positive input, got 0.0"`. Following the existing `param <= 0` → `2-1-15-3` pattern, added explicit `x <= 0` → `2-1-15-8` (new code). The numeric-exception test now asserts the code instead of CPython's text.
- **Improved logging for `SemanticError 1-1-10-10`** — the `{rule}` placeholder was being filled with the whole `DefIdentifier` AST node, rendering as `<DefIdentifier(column_stop=10, ..., value=EU)>`. Now passes `left_part.value` so the message reads `"Check the EQ rules: EU."` (orthogonal cleanup, but the same site that the PEP 649 issue made empty on 3.14).

## Checklist

- [x] Tests pass (`pytest`) — 4072 passed, coverage 91% (≥90% gate), with `pyarrow 24.0.0` + `numpy 2.4.4` on Python 3.12
- [ ] Documentation updated (not applicable — README/docs wording "Python 3.9 or higher" remains accurate as install policy; new error code `2-1-15-8` will be picked up automatically by `docs/error_messages.rst` regeneration)

## Impact / Risk

- **Breaking changes:** None for end users. `pip install vtlengine` still resolves on Python 3.9; the package classifier list also still advertises 3.9.
- **Resolver behavior with broadened upper bounds (verified via PyPI wheel matrix):**
  - pyarrow: 3.10–3.14 all resolve to `24.0.0` (cp310–cp314 wheels available).
  - numpy: 3.10 → `2.2.6` (only version with cp310 wheel in range), 3.11–3.14 → `2.4.4` (latest).
  - pandas: pin (`<3.0`) already covers cp314-ready `2.3.3`.
  - duckdb: pin (`<1.5`) already covers cp314-ready `1.4.4`.
  - pysdmx: pure Python (`py3-none-any.whl`), no wheel concern.
- **CI/install mismatch:** Dropping 3.9 from CI while keeping `requires-python = ">=3.9"` means breakage on 3.9 will not be caught by CI. The `ruff target-version = "py39"` guard catches the most common breakage path (3.10+ syntax slipping in); residual risk (a dep update silently dropping 3.9, or a runtime-only behavior change ruff cannot catch) is accepted by design.
- **Error-message stability:** The 3.14 fixes (date, log) move classification away from regex-matching CPython's internal `ValueError` strings into structure-based checks we own. Less future churn when CPython reworks error wording again.
- **Conditional numpy pin** for `python_version < '3.10'` retained intentionally — numpy 2.2+ requires Python 3.10+, so 3.9 users still need numpy `<2.1`.

## Notes

cp314 wheel pre-flight + cross-Python wheel-matrix check confirmed:

| package | 3.10 | 3.11 | 3.12 | 3.13 | 3.14 | resolver picks (after bump) |
| --- | :-: | :-: | :-: | :-: | :-: | --- |
| pyarrow 24.0.0 | ✓ | ✓ | ✓ | ✓ | ✓ | 24.0.0 across the board |
| numpy 2.4.4   | — | ✓ | ✓ | ✓ | ✓ | 2.4.4 on 3.11+, 2.2.6 on 3.10 |
| pandas 2.3.3  | ✓ | ✓ | ✓ | ✓ | ✓ | 2.3.3 across the board |
| duckdb 1.4.4  | ✓ | ✓ | ✓ | ✓ | ✓ | 1.4.4 across the board |

`ruff target-version = "py39"` surfaced **zero** violations in the existing codebase, confirming it has been organically 3.9-compatible.

Other workflows (`docs.yml`, `version.yml`, `release.yml`, `ubuntu_test_24_04.yml`) intentionally untouched.

Fixes #674


mla2001 (approved):
Looks fine! 😊

PR #685, opened 2026-05-05
Fix #684: Cache ANTLR4 runtime separately from the C++ parser wheel

### LINKED ISSUE #684: Cache ANTLR4 runtime separately from the C++ parser wheel
### Summary

The `Build C++ parser wheel` step in CI takes ~53s on Linux/macOS and ~2:25 on Windows on every cache miss because it compiles the entire vendored ANTLR4 C++ runtime (138 of 145 objects) plus the 7 VTL-specific files. The current cache key invalidates on any change to `pyproject.toml`, `CMakeLists.txt`, or the parser sources, so most of the build time is recompiling unchanged third-party code.

This task splits the cache so the ANTLR4 runtime is built once per OS + Python combination and reused across wheel builds. Expected savings on warm cache: ~50s/job on Linux/macOS, ~2:00/job on Windows.

### Scope

**Modified files**:

- `CMakeLists.txt` — add `VTLENGINE_USE_PREBUILT_ANTLR4` / `VTLENGINE_BUILD_ANTLR4_PREBUILT` options
- `pyproject.toml` — exclude `lib/*` and `include/*` from the wheel
- `.github/workflows/testing.yml` — split into ANTLR4 cache + wheel cache
- `.github/workflows/ubuntu_test_24_04.yml` — same

**Out of scope**:

- ccache / sccache integration
- System `libantlr4-runtime-dev` via apt (Ubuntu-only path)
- Trimming the vendored ANTLR4 runtime to only required files

### Testing

- Local: build runtime separately via standalone CMake invocation, build wheel against the produced prefix, verify import + smoke tests pass
- Local backward-compat: build with default flags to confirm the from-source path still works
- CI: open PR; first run = cache misses (parity with current); subsequent runs = ANTLR4 runtime cache hits, build step drops from ~50s to <10s on Linux/macOS

### Additional Context

The C++ parser was introduced in #588. `CMakeLists.txt` and the build scripts are identical between `main` and `duckdb/main`, so a single PR onto `main` covers both branches once they reconcile.

javihern98:
## Summary

The CI `Build C++ parser wheel` step was recompiling the full vendored ANTLR4 C++ runtime (138 of 145 translation units) on every cache miss, because the wheel cache key included `pyproject.toml`, `CMakeLists.txt`, and the parser sources — any of which routinely change while the runtime itself does not.

This PR splits the cache:

- **`antlr4-runtime` cache**: keyed on `scripts/setup_antlr4_runtime.sh` + `CMakeLists.txt`. Holds a prebuilt static library + headers under `build/antlr4-prefix/`.
- **`cpp-parser` wheel cache**: keyed on parser sources + `pyproject.toml` + `CMakeLists.txt`. Only invalidates on changes that genuinely affect the wheel.

`CMakeLists.txt` gains three opt-in options used only by CI; the default build path is unchanged.

| Local benchmark (parallel build) | Time |
|---|---|
| Runtime build | 8.5s |
| Wheel build against prebuilt runtime | 8.5s |

Expected CI savings on warm cache: ~50s/job on Linux/macOS, ~2:00/job on Windows. Across the 15-job `Testing` matrix + the `Ubuntu 24.04 Tests` job, that's roughly **~20 minutes per CI run** when the runtime cache hits but the wheel cache misses.

## Checklist

- [x] Code quality checks pass (`ruff format`, `ruff check`, `mypy`) — no Python sources touched, ran the suite anyway as a sanity check
- [x] Tests pass (`pytest tests/Aggregate -x -q`: 71 passed; full suite untouched)
- [x] Documentation updated (if applicable) — N/A

## Impact / Risk

- **Breaking changes**: none. The new `VTLENGINE_USE_PREBUILT_ANTLR4` / `VTLENGINE_BUILD_ANTLR4_PREBUILT` options default to `OFF`. Downstream consumers (incl. `cibuildwheel` for release artifacts) continue to build from source via the existing path.
- **Data/SDMX compatibility**: none.
- **Release notes**: CI optimization only, no user-visible behavior change.

## Notes

- The first CI run on this PR will populate both caches, so the build steps will look like the current behavior (~50s on Linux/macOS, ~2:25 on Windows). Subsequent runs (or rebases that don't touch the runtime cache key inputs) should drop the build step from ~50s to <10s on Linux/macOS.
- `find_package(pybind11)` was moved into the `if(NOT VTLENGINE_BUILD_ANTLR4_PREBUILT)` block so the standalone runtime build doesn't require pybind11 to be available.
- The `target_include_directories(vtl_cpp_parser PRIVATE "${ANTLR4_SRC_DIR}")` line was dropped because the include path now propagates via `antlr4_runtime`'s `INTERFACE_INCLUDE_DIRECTORIES` (PUBLIC for the from-source target, set explicitly on the imported target for the prebuilt path). This makes the prebuilt path work without re-downloading the source at wheel-build time.

Closes #684


mla2001 (approved):
Looks fine! 😊

PR #713, opened 2026-05-12
Wire sccache into release wheel builds

### LINKED ISSUE #694: Wire sccache compiler caching into release.yml
## Summary

Cut release wheel build wall-clock by adding `sccache` as a CMake compiler launcher in `.github/workflows/release.yml`. Today the 145-file ANTLR4 C++ runtime is recompiled from scratch for every Python version in every job, even though that source is Python-version-independent. Compilation dominates wall-clock — Windows AMD64 alone is ~21 min/job.

With sccache wired in:

- Within a job, builds 2–6 should cache-hit on the runtime objects; only the small pybind11 binding files genuinely rebuild per Python.
- Across runs, the GitHub Actions cache backend persists object files so cp39 hits in subsequent releases too.

Expected per-platform wall-clock improvement:

| Platform        | Current | Target     |
|-----------------|---------|------------|
| Windows AMD64   | ~21 min | ~10–12 min |
| Linux x86_64    | ~7 min  | ~3–4 min   |
| Linux aarch64   | ~7 min  | ~3–4 min   |
| macOS arm64     | ~8 min  | ~4–5 min   |

## Scope

Single file: [.github/workflows/release.yml](.github/workflows/release.yml).

1. Add a `Setup sccache` step before `Build wheels`:

   ```yaml
   - name: Setup sccache
     uses: mozilla-actions/sccache-action@<commit-sha>  # pin to v0.0.10
   ```

2. Extend the `Build wheels` step `env:` block:

   ```yaml
   env:
     CIBW_ARCHS: ${{ matrix.arch }}
     CMAKE_C_COMPILER_LAUNCHER: sccache
     CMAKE_CXX_COMPILER_LAUNCHER: sccache
     SCCACHE_GHA_ENABLED: "true"
     # Linux runs in manylinux containers — install sccache there + pass GHA env vars in
     CIBW_BEFORE_ALL_LINUX: |
       ARCH=$(uname -m)
       curl -fsSL "https://github.com/mozilla/sccache/releases/download/v0.15.0/sccache-v0.15.0-${ARCH}-unknown-linux-musl.tar.gz" \
         | tar -xz --strip-components=1 -C /usr/local/bin --wildcards '*/sccache'
       chmod +x /usr/local/bin/sccache
     CIBW_ENVIRONMENT_PASS_LINUX: >-
       SCCACHE_GHA_ENABLED
       ACTIONS_CACHE_URL
       ACTIONS_RUNTIME_TOKEN
       ACTIONS_RESULTS_URL
       ACTIONS_CACHE_SERVICE_V2
   ```

No `pyproject.toml` change needed — CMake reads `CMAKE_*_COMPILER_LAUNCHER` from env automatically and scikit-build-core forwards env vars to CMake. The existing `[tool.cibuildwheel.linux]` `before-build` hook (`scripts/setup_antlr4_runtime.sh`) is orthogonal to `CIBW_BEFORE_ALL_LINUX` and they don't collide.

## Open question to resolve before implementation

> [!IMPORTANT]
> Whether scikit-build-core's CMake invocation passes Python-version-specific flags (e.g., `-I.../cp310/include`, `-DPYBIND11_PYTHON_VERSION=3.10`) **globally** to the ANTLR4 runtime targets, or only to the pybind11 bindings target. If global, the runtime objects get distinct cache keys per Python and we lose the within-job hits — sccache would still help across-release reuse, but the within-job speedup would evaporate.

Verify before committing the workflow change. One quick way: build twice locally with sccache enabled and `sccache --show-stats` after each build, comparing against an isolated build that targets only the runtime library.

## Risks

- **Stale-cache wrong-output bugs.** Low probability; sccache hashes source + flags + compiler version. Surface here is small (well-known toolchains, stable ANTLR runtime).
- **Third-party in the release critical path.** sccache or `mozilla-actions/sccache-action` could break (it has during past GHA cache API rollovers). Pin by SHA.
- **Cache eviction** under GHA's 10 GB per-repo LRU. Hit rate may drift; rare for a single-repo workflow.
- **First release after merge**: cp39 cold-misses; only cp310+ within the same job benefit on the first run.
- **Compile-error UX**: hits skip the compiler entirely, so flaky build failures might appear/disappear depending on cache state.

## Verification (after implementation)

1. The `Setup sccache` step prints `sccache --show-stats` after each job — confirm non-zero hits.
2. Build times for cp310–cp314 wheels in a job are visibly shorter than cp39 in the same job.
3. Per-platform wall-clock falls into the target band above (otherwise revert).
4. Repo `Caches` page shows an `sccache` entry.

## References

- Background analysis from #692 / #693: per-Python recompilation of ANTLR4 C++ runtime is the main contributor to job wall-clock.
- mozilla-actions/sccache-action: <https://github.com/mozilla-actions/sccache-action>
- sccache GHA backend: <https://github.com/mozilla/sccache/blob/main/docs/GHA.md>

jii001:
## Summary
- add pinned mozilla-actions/sccache-action setup to the release wheel job
- configure CMake compiler launcher variables so scikit-build-core uses sccache
- install sccache inside Linux cibuildwheel containers and pass GitHub cache env vars through

## Testing
- `python -c "import yaml, pathlib; yaml.safe_load(pathlib.Path('.github/workflows/release.yml').read_text()); print('yaml ok')"`
- `git diff --check`

Closes #694

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,10 +29,27 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: Setup sccache
+        uses: mozilla-actions/sccache-action@1583d6b38d7be47f593cb472781bbb21cab4321e # v0.0.10
+
       - name: Build wheels
         uses: pypa/cibuildwheel@d16cc544083d96996f923fc4b318fee3cb14b34e # v3.4.1
         env:
           CIBW_ARCHS: ${{ matrix.arch }}
+          CMAKE_C_COMPILER_LAUNCHER: sccache
+          CMAKE_CXX_COMPILER_LAUNCHER: sccache
+          SCCACHE_GHA_ENABLED: "true"
+          CIBW_BEFORE_ALL_LINUX: |
+            ARCH=$(uname -m)
+            curl -fsSL "https://github.com/mozilla/sccache/releases/download/v0.15.0/sccache-v0.15.0-${ARCH}-unknown-linux-musl.tar.gz" \
+              | tar -xz --strip-components=1 -C /usr/local/bin --wildcards '*/sccache'
+            chmod +x /usr/local/bin/sccache
+          CIBW_ENVIRONMENT_PASS_LINUX: >-
+            SCCACHE_GHA_ENABLED
+            ACTIONS_CACHE_URL
+            ACTIONS_RUNTIME_TOKEN
+            ACTIONS_RESULTS_URL
+            ACTIONS_CACHE_SERVICE_V2
 
       - name: Upload wheels
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2


javihern98:
Closed as external contrbutions in workflows are not allowed

PR #717, opened 2026-05-12
Fix #716: Read DuckDB env vars lazily at run time

### LINKED ISSUE #716: DuckDB env vars (VTL_THREADS, VTL_MEMORY_LIMIT, etc.) are cached at module import time
## Summary

When a user script mutates DuckDB-related environment variables via `os.environ` **after** importing `vtlengine`, the values are silently ignored. The DuckDB engine continues to use whatever the env vars were at the moment `vtlengine.duckdb_transpiler.Config.config` was first imported.

> [!IMPORTANT]
> This affects scripts that try to configure DuckDB at runtime — a common pattern with notebooks, test harnesses, and any code path that wants to override defaults programmatically rather than via the shell environment.

Affected variables (all read into module-level constants):

| Env var | Module |
| --- | --- |
| `VTL_MEMORY_LIMIT` | `duckdb_transpiler/Config/config.py` |
| `VTL_THREADS` | `duckdb_transpiler/Config/config.py` |
| `VTL_TEMP_DIRECTORY` | `duckdb_transpiler/Config/config.py` |
| `VTL_MAX_TEMP_DIRECTORY_SIZE` | `duckdb_transpiler/Config/config.py` |
| `VTL_USE_IN_MEMORY_DB` | `duckdb_transpiler/Config/config.py` |
| `VTL_SKIP_LOAD_VALIDATION` | `duckdb_transpiler/io/_io.py` |

`VTL_DUCKDB_DECIMAL_WIDTH` and `OUTPUT_NUMBER_SIGNIFICANT_DIGITS` are **not** affected — they are already re-read at call time inside `set_decimal_config()` and `_parse_env_value()`.

## Reproducible Example

```python
import os

import pandas as pd
from vtlengine import run
from vtlengine.duckdb_transpiler.Config import config as cfg


# Mutate env AFTER `from vtlengine import run` — this is silently ignored
os.environ["VTL_THREADS"] = "7"

script = "DS_A := DS_1 * 10;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

datapoints = {"DS_1": pd.DataFrame({"Id_1": [1, 2, 3], "Me_1": [10, 20, 30]})}

# Spy on what DuckDB actually receives
orig = cfg.configure_duckdb_connection


def spy(conn):
    orig(conn)
    name, value = conn.execute(
        "SELECT name, value FROM duckdb_settings() WHERE name = 'threads'"
    ).fetchone()
    print(f"DuckDB {name} = {value!r}  (env said VTL_THREADS={os.environ['VTL_THREADS']!r})")


cfg.configure_duckdb_connection = spy

run(script=script, data_structures=data_structures, datapoints=datapoints, use_duckdb=True)
```

## Expected vs Actual

Expected:

```
DuckDB threads = '7'  (env said VTL_THREADS='7')
```

Actual:

```
DuckDB threads = '1'  (env said VTL_THREADS='7')
```

The same divergence is observable for `VTL_MEMORY_LIMIT`, `VTL_TEMP_DIRECTORY`, `VTL_USE_IN_MEMORY_DB`, `VTL_MAX_TEMP_DIRECTORY_SIZE`, and `VTL_SKIP_LOAD_VALIDATION`.

## vtlengine version

1.7.0

## Python version

3.12.3

## OS

Ubuntu 24.04

## Additional Context

Root cause is module-level capture in [`config.py`](https://github.com/Meaningful-Data/vtlengine/blob/main/src/vtlengine/duckdb_transpiler/Config/config.py) lines 122, 125, 128, 131, 134 and [`_io.py`](https://github.com/Meaningful-Data/vtlengine/blob/main/src/vtlengine/duckdb_transpiler/io/_io.py) line 37:

```python
MEMORY_LIMIT: str = os.getenv("VTL_MEMORY_LIMIT", "80%")
THREADS: int = int(os.getenv("VTL_THREADS", "1"))
TEMP_DIRECTORY: str = os.getenv("VTL_TEMP_DIRECTORY", tempfile.gettempdir())
MAX_TEMP_DIRECTORY_SIZE: str = os.getenv("VTL_MAX_TEMP_DIRECTORY_SIZE", "")
USE_IN_MEMORY_DB: bool = os.getenv("VTL_USE_IN_MEMORY_DB", "1").lower() in ("1", "true")
```

These are evaluated once at module import time. `configure_duckdb_connection()` then uses the cached values rather than re-reading `os.environ`.

> [!NOTE]
> Setting the env vars **before** `import vtlengine` (or via the shell before launching Python) works correctly — the issue is only with mutations made after the import has happened.

javihern98:
## Summary

Fixes #716. Six DuckDB-related env vars were read into module-level constants when `duckdb_transpiler.Config.config` (or `duckdb_transpiler.io._io`) was first imported, so mutations to `os.environ` from user code after `import vtlengine` were silently ignored. Replace each module-level capture with a small accessor function that reads `os.environ` on every call.

Affected variables:

- `VTL_MEMORY_LIMIT`, `VTL_THREADS`, `VTL_TEMP_DIRECTORY`, `VTL_MAX_TEMP_DIRECTORY_SIZE`, `VTL_USE_IN_MEMORY_DB` (in `config.py`)
- `VTL_SKIP_LOAD_VALIDATION` (in `_io.py`)

`VTL_DUCKDB_DECIMAL_WIDTH` and `OUTPUT_NUMBER_SIGNIFICANT_DIGITS` were already lazy and remain unchanged.

## Checklist

- [x] Tests pass (`pytest` — 4882 passed, 27 skipped)

## Impact / Risk

- **Breaking changes?** No. The removed module-level names (`MEMORY_LIMIT`, `THREADS`, `TEMP_DIRECTORY`, `MAX_TEMP_DIRECTORY_SIZE`, `USE_IN_MEMORY_DB`, `SKIP_LOAD_VALIDATION`) are not re-exported anywhere and not referenced outside the modules where they were defined (verified by repo-wide grep).
- **Data/SDMX compatibility:** None.
- **Behavior changes:** Previously dead branch `if THREADS is not None` was dropped (the accessor always returns `int`). Existing tests cover the behavior end-to-end.
- **Notes for release/changelog:** Worth flagging in the changelog under "Bug fixes" — users who set env vars in Python code instead of the shell will get correct behavior for the first time.

## Notes

- Added `tests/duckdb_transpiler/test_env_var_lazy.py` (8 tests) covering each accessor individually plus one integration check that mutates env after import and reads back from `duckdb_settings()`.
- Follow-up worth doing separately: document the env-var precedence in `docs/environment_variables.rst` (when shell env vs Python `os.environ` mutations take effect). Intentionally not in this PR.

### DIFF
--- a/src/vtlengine/duckdb_transpiler/Config/config.py
+++ b/src/vtlengine/duckdb_transpiler/Config/config.py
@@ -118,20 +118,31 @@ def set_decimal_config() -> None:
 # Memory & Performance Configuration
 # =============================================================================
 
-# Default memory limit (80% of system RAM)
-MEMORY_LIMIT: str = os.getenv("VTL_MEMORY_LIMIT", "80%")
+# Accessor functions read os.environ on every call so user scripts can mutate
+# os.environ after `import vtlengine` and have those values take effect on the
+# next run(). Capturing them in module-level constants would freeze the values
+# at import time.
 
-# Default thread count (default = 1)
-THREADS: int = int(os.getenv("VTL_THREADS", "1"))
 
-# Temp directory for spill-to-disk
-TEMP_DIRECTORY: str = os.getenv("VTL_TEMP_DIRECTORY", tempfile.gettempdir())
+def _memory_limit() -> str:
+    return os.getenv("VTL_MEMORY_LIMIT", "80%")
 
-# Max temp directory size for spill-to-disk (empty = use available disk space)
-MAX_TEMP_DIRECTORY_SIZE: str = os.getenv("VTL_MAX_TEMP_DIRECTORY_SIZE", "")
 
-# Use file-backed database instead of in-memory (better for large datasets)
-USE_IN_MEMORY_DB: bool = os.getenv("VTL_USE_IN_MEMORY_DB", "1").lower() in ("1", "true")
+def _threads() -> int:
+    return int(os.getenv("VTL_THREADS", "1"))
+
+
+def _temp_directory() -> str:
+    return os.getenv("VTL_TEMP_DIRECTORY", tempfile.gettempdir())
+
+
+def _max_temp_directory_size() -> str:
+    return os.getenv("VTL_MAX_TEMP_DIRECTORY_SIZE", "")
+
+
+def _use_in_memory_db() -> bool:
+    return os.getenv("VTL_USE_IN_MEMORY_DB", "1").lower() in ("1", "true")
+
 
 # Minimum storage version required by the transpiler (typed macro parameters need >= v1.4.0).
 # DuckDB defaults to an older on-disk format for portability, so it must be set explicitly.
@@ -150,7 +161,7 @@ def get_memory_limit_bytes() -> int:
     Returns:
         Memory limit in bytes
     """
-    limit = MEMORY_LIMIT.strip().upper()
+    limit = _memory_limit().strip().upper()
 
     total_ram = psutil.virtual_memory().total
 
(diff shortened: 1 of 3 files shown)


mla2001 (approved):
Looks fine! 😊

PR #725, opened 2026-05-14
Fix #724: Raise VTLSyntaxError instead of crashing on malformed VTL

### LINKED ISSUE #724: Invalid VTL syntax raises AttributeError instead of a clean error
### Initial Checks

### Summary

When a VTL script contains a syntax error (for example, using `=` instead of `:=` inside a `calc` clause), the engine fails with an internal Python `AttributeError: 'TerminalNode' object has no attribute 'children'` originating from the AST visitor. The user sees a stack trace pointing inside `ExprComponents.py`, with no indication of where the problem is in their script.

The root cause is in the C++ binding (`src/vtlengine/AST/Grammar/_cpp_parser/bindings.cpp`): `g_state.parser->removeErrorListeners()` is called before parsing, so ANTLR's diagnostics are silenced. The parser then performs error recovery and produces a partially-malformed parse tree, and the Python visitor crashes downstream because it expects a non-terminal context but receives a `TerminalNode`.

The engine should raise a clean, user-facing error with the offending line and column instead.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = """
    DS_A <- DS_1[calc Me_2 = 42];
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

data_df = pd.DataFrame({"Id_1": [1, 2, 3], "Me_1": [10, 20, 30]})

print(run_result)
```

Note: the script uses `=` instead of `:=` inside `calc`. The VTL grammar requires `:=` (the `ASSIGN` token). The reproduction is the syntax error itself — any malformed script triggers the same class of failure.

### Expected vs Actual

Expected:

```
VTLSyntaxError: VTL syntax error at line 2, column 28: mismatched input '=' expecting ':='
```

with a referenceable error code and `.pos` carrying the line/column.

Actual:

```
Traceback (most recent call last):
  ...
  File ".../AST/ASTConstructorModules/ExprComponents.py", line 54, in visitExprComponent
    ctx_list = ctx.children
AttributeError: 'vtlengine.AST.Grammar._cpp_parser.vtl_cpp_parser.TerminalNode' object has no attribute 'children'
```

### vtlengine version

1.7.0

### Python version

3.12

### OS

Linux

### Additional Context

Root cause is the unconditional `removeErrorListeners()` call in the C++ binding:

https://github.com/Meaningful-Data/vtlengine/blob/main/src/vtlengine/AST/Grammar/_cpp_parser/bindings.cpp#L493

ANTLR finds and reports the syntax error correctly, but the binding discards every listener (default and otherwise), so the error never reaches Python. Replacing the silent removal with a collecting listener that captures `{line, column, message, offending_text}` and exposing it to Python is enough to surface a clean error before the AST visitor runs.

javihern98:
## Summary

- Replace silent `removeErrorListeners()` in the C++ binding with a `CollectingErrorListener` that captures ANTLR syntax errors.
- Surface them in Python: `create_ast()` checks `get_syntax_errors()` and raises the new `VTLSyntaxError` (error code `0-1-4-1`) with the offending line and column before the AST visitor runs.
- Eliminates the `AttributeError: 'TerminalNode' object has no attribute 'children'` crash that previously masked the real problem.

Before:
```
AttributeError: 'TerminalNode' object has no attribute 'children'
```
After:
```
VTLSyntaxError: VTL syntax error at line 1, column 26: mismatched input '=' expecting ':='
```

## Checklist

- [x] Tests pass (`pytest`)
- [ ] Documentation updated (if applicable) — N/A; new error codes are auto-included in `docs/error_messages.rst`.

## Impact / Risk

- **Public API**: adds `VTLSyntaxError` under `vtlengine.Exceptions`. Existing callers catching `Exception` keep working; no caller is expected to rely on the prior `AttributeError`.
- **C++ binding**: rebuild required (wheel builds pick this up automatically; no surface change on the C++ side).
- **Behaviour**: parser errors are now raised eagerly at parse time. No change for valid VTL.

## Notes

The original investigation also surfaced a separate bug: `max(<identifier> over(partition by ...))` silently returns the original column instead of computing the max — caused by a duplicated-name column in DuckDB (`VALPAR` / `VALPAR_1`). Not addressed in this PR; will be filed separately.

Fixes #724

### DIFF
--- a/src/vtlengine/API/__init__.py
+++ b/src/vtlengine/API/__init__.py
@@ -30,7 +30,7 @@ from vtlengine.AST.Grammar._cpp_parser import vtl_cpp_parser
 from vtlengine.duckdb_transpiler.Config.config import configured_connection
 from vtlengine.duckdb_transpiler.io import execute_queries, extract_datapoint_paths
 from vtlengine.duckdb_transpiler.Transpiler import SQLTranspiler
-from vtlengine.Exceptions import InputValidationException
+from vtlengine.Exceptions import InputValidationException, VTLSyntaxError
 from vtlengine.files.output import format_date_iso8601
 from vtlengine.files.output._time_period_representation import (
     TimePeriodRepresentation,
@@ -90,6 +90,15 @@ def create_ast(text: str) -> Start:
     """
     text = text + "\n"
     cst = vtl_cpp_parser.parse(text)
+    errors = vtl_cpp_parser.get_syntax_errors()
+    if errors:
+        first = errors[0]
+        raise VTLSyntaxError(
+            "0-1-4-1",
+            line=first["line"],
+            column=first["column"] + 1,
+            detail=first["message"],
+        )
     visitor = ASTVisitor()
     ast = visitor.visitStart(cst)
     DAGAnalyzer.create_dag(ast)
--- a/src/vtlengine/AST/Grammar/_cpp_parser/bindings.cpp
+++ b/src/vtlengine/AST/Grammar/_cpp_parser/bindings.cpp
@@ -348,9 +348,45 @@ struct ParserState {
         int column;
     };
     std::vector<CommentInfo> comments;
+
+    struct SyntaxErrorInfo {
+        int line;
+        int column;
+        std::string message;
+        std::string offending_text;
+    };
+    std::vector<SyntaxErrorInfo> syntax_errors;
 };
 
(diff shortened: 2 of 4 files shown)


albertohernandez1995 (approved):
Well done

PR #727, opened 2026-05-14
Fix #726: Source-line preview and token-underline caret in VTLSyntaxError

### LINKED ISSUE #726: Add Python-style source preview to VTLSyntaxError
### Initial Checks

### Problem Description

`VTLSyntaxError` (introduced in #724/#725) currently reports the location and the underlying ANTLR message:

```
VTLSyntaxError: VTL syntax error at line 1, column 26: mismatched input '=' expecting ':='
```

This is correct but minimal. When the script has more than a handful of lines, the reader still has to count columns by hand to find the offending token. Python's own `SyntaxError` solves this by quoting the offending source line and putting a caret directly under the problem; the same convention would make VTL errors a lot quicker to read.

### Proposed Solution

Extend `VTLSyntaxError` to include a compact source-line preview with a caret underneath the offending token, in the same single-message format already used for `str(exc)`:

```
VTL syntax error at line 1, column 26: mismatched input '=' expecting ':='
    DS_A <- DS_1[calc test_2 = 42];
                             ^
```

For multi-character offending tokens, the caret should underline the full token (e.g. `^^^^^^^` for an identifier):

```
VTL syntax error at line 2, column 21: missing ':=' at 'foo_bar'
    DS_B := DS_2[calc x foo_bar 5];
                        ^^^^^^^
```

Implementation notes:

- C++ binding (`bindings.cpp`): extend `CollectingErrorListener` to also capture `source_line` and `underline_length` (from the offending token's `getStartIndex()`/`getStopIndex()`); expand tabs to 4 spaces and remap the column so the caret stays aligned.
- Python (`vtlengine.Exceptions.VTLSyntaxError`): rebuild the constructor as keyword-only, build the message inline (no error code), append the preview block when `source_line` is non-empty.
- Drop the `0-1-4-1` entry from `centralised_messages` — parser errors are user-facing, location-stamped messages and don't need a stable code; without one, `str(exc)` is a plain string instead of the tuple `("...", "0-1-4-1")`.

### Alternatives Considered

- **Rust/cargo-style multi-line layout** with a `|` gutter and `-->` marker. More visually polished but also more characters; rejected for compactness.
- **GCC/clang `<file>:line:col` prefix**. Familiar to compiler users but VTL doesn't always have a file (the script can be a bare string).
- **`getExpectedTokens()` enrichment** (e.g. "expected one of: ':='"). Useful but would inflate the header line; deferred.

### Additional Context

Implementation already written and tested locally on top of #725. Will be opened as a draft PR alongside this issue. Adds five new tests in `tests/API/test_syntax_errors.py` covering: single-char caret, multi-char underline, multi-line scripts, tab-indented sources, and a regression guard for valid scripts.

javihern98:
## Summary

- Extend the `CollectingErrorListener` in the C++ binding to also capture `source_line` and `underline_length` from the offending token's start/stop indices.
- Expand tabs to 4 spaces and remap the column so the caret stays aligned with the offending text.
- Drop the `0-1-4-1` error-code entry — parser errors are free-form and location-stamped; without a code, `str(exc)` is a plain string instead of the tuple `("message", "0-1-4-1")`.
- Rewrite `VTLSyntaxError`: keyword-only constructor, builds the message inline (with the preview block when `source_line` is present).

Before:
```
VTLSyntaxError: ('VTL syntax error at line 1, column 26: ... ', '0-1-4-1')
```
After:
```
VTL syntax error at line 1, column 26: mismatched input '=' expecting ':='
    DS_A <- DS_1[calc test_2 = 42];
                             ^
```

Multi-character offending tokens (e.g. an identifier) get a `^^^...` underline matching the token's length.

## Checklist

- [x] Tests pass (`pytest tests/API` — 208, +5 new in `test_syntax_errors.py`)
- [ ] Documentation updated (if applicable) — N/A.

## Impact / Risk

- Public API: `VTLSyntaxError` constructor signature changed (keyword-only `line`/`column`/`detail`/`source_line`/`underline_length`; no longer takes an error code as first positional arg). Callers within vtlengine are updated; external callers constructing `VTLSyntaxError` directly would need to update, but external construction of this class is not expected (it's raised by the engine).
- `str(VTLSyntaxError(...))` returns the formatted message string instead of a `("message", "0-1-4-1")` tuple. Callers comparing against the tuple would break; comparing against the message string still works.
- C++ binding: rebuild required (wheel build picks this up automatically).
- The `0-1-4-1` entry is removed from `centralised_messages`; it was introduced in the same PR series (#725) and never had external consumers.

## Notes

Follows up on #725. Tests cover: single-char caret, multi-char underline, errors on line ≥ 2, tab-indented sources, and a regression guard for valid scripts.

Fixes #726

### DIFF
--- a/src/vtlengine/API/__init__.py
+++ b/src/vtlengine/API/__init__.py
@@ -94,10 +94,11 @@ def create_ast(text: str) -> Start:
     if errors:
         first = errors[0]
         raise VTLSyntaxError(
-            "0-1-4-1",
             line=first["line"],
             column=first["column"] + 1,
             detail=first["message"],
+            source_line=first["source_line"],
+            underline_length=first["underline_length"],
         )
     visitor = ASTVisitor()
     ast = visitor.visitStart(cst)
--- a/src/vtlengine/API/__init__.py
+++ b/src/vtlengine/API/__init__.py
@@ -90,15 +90,14 @@ def create_ast(text: str) -> Start:
     """
     text = text + "\n"
     cst = vtl_cpp_parser.parse(text)
-    errors = vtl_cpp_parser.get_syntax_errors()
-    if errors:
-        first = errors[0]
+    error = vtl_cpp_parser.get_syntax_error()
+    if error is not None:
         raise VTLSyntaxError(
-            line=first["line"],
-            column=first["column"] + 1,
-            detail=first["message"],
-            source_line=first["source_line"],
-            underline_length=first["underline_length"],
+            line=error["line"],
+            column=error["column"] + 1,
+            detail=error["message"],
+            source_line=error["source_line"],
+            underline_length=error["underline_length"],
         )
     visitor = ASTVisitor()
     ast = visitor.visitStart(cst)
--- a/src/vtlengine/AST/Grammar/_cpp_parser/bindings.cpp
+++ b/src/vtlengine/AST/Grammar/_cpp_parser/bindings.cpp
@@ -18,6 +18,7 @@
 #include "VtlParser.h"
 
 #include <memory>
+#include <optional>
 #include <string>
 #include <typeindex>
 #include <unordered_map>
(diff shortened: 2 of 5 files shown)


mla2001 (approved):
Looks fine!

PR #729, opened 2026-05-14
Fix #728: Filter with max() over(partition by ...) fails in both engines

### LINKED ISSUE #728: Filter with max() over(partition by ...) fails in both engines
### Initial Checks

### Summary

A `filter` clause whose condition uses a window aggregate over an identifier (e.g. `max(Id_2 over(partition by Id_1))`) fails in **both** engines, for **different root causes**:

- The **pandas engine** silently returns all input rows. The window `max(Id_2 over(partition by Id_1))` evaluates to `Id_2` itself row-by-row instead of the partition max, so the filter condition `Id_2 = Id_2` is trivially satisfied and the filter is a no-op.
- The **DuckDB engine** raises `BinderException: WHERE clause cannot contain window functions!`. The transpiler emits SQL of the form `SELECT * FROM "DS_1" WHERE ("Id_2" = MAX("Id_2") OVER (PARTITION BY "Id_1"))`, which is invalid SQL — window functions are not allowed inside `WHERE` and must be wrapped in a subquery/CTE.

Both behaviors are wrong: the script should keep only rows where the identifier equals the per-partition maximum (i.e. one row in the minimal repro).

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


def main():
    script = """
        DS_A <- DS_1[filter Id_2 = max(Id_2 over(partition by Id_1))];
    """

    data_structures = {
        "datasets": [
            {
                "name": "DS_1",
                "DataStructure": [
                    {"name": "Id_2", "type": "Integer", "role": "Identifier", "nullable": False},
                    {"name": "Me_1", "type": "Integer", "role": "Measure", "nullable": False},
                ],
            }
        ]
    }

    data_df = pd.DataFrame({"Id_1": [1, 1], "Id_2": [10, 20], "Me_1": [100, 200]})

    print("=== Input ===")
    print(data_df.to_string(index=False))
    print("\nExpected: only row with Id_2=20 (partition max)")

    print("\n=== Pandas engine ===")
    try:
        r = run(
            script=script,
            datapoints={"DS_1": data_df.copy()},
            use_duckdb=False,
        )
        print(r["DS_A"].data.to_string(index=False))
    except Exception as e:
        print(f"ERROR: {type(e).__name__}: {e}")

    print("\n=== DuckDB engine ===")
    try:
        r = run(
            script=script,
            datapoints={"DS_1": data_df.copy()},
            use_duckdb=True,
        )
        print(r["DS_A"].data.to_string(index=False))
    except Exception as e:
        print(f"ERROR: {type(e).__name__}: {e}")


if __name__ == "__main__":
    main()
```

### Expected vs Actual

**Expected output** — only the row with `Id_2 = 20` (the partition max for `Id_1 = 1`):

| Id_1 | Id_2 | Me_1 |
|------|------|------|
| 1    | 20   | 200  |

**Actual** (verbatim console output of the script above):

```
=== Input ===
 Id_1  Id_2  Me_1
    1    10   100
    1    20   200

Expected: only row with Id_2=20 (partition max)

=== Pandas engine ===
 Id_1  Id_2  Me_1
    1    10   100
    1    20   200

=== DuckDB engine ===
ERROR: BinderException: Binder Error: WHERE clause cannot contain window functions!

LINE 1: ...REATE TABLE "DS_A" AS SELECT * FROM "DS_1" WHERE ("Id_2" = MAX("Id_2") OVER (PARTITION BY "Id_1"))
                                                                      ^
```

- Pandas engine returns **both** input rows (filter is a no-op) — wrong.
- DuckDB engine raises `BinderException` — wrong.

### vtlengine version

1.7.0

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

> [!NOTE]
> Two distinct underlying defects surface together on this minimal input, so they are filed as one issue:
> 1. **pandas semantics**: `max(<identifier> over(partition by ...))` returns the identifier value of the row instead of the partition max. This also reproduces with `calc` (e.g. `DS_A <- DS_1[calc Id_2_max := max(Id_2 over(partition by Id_1))]` gives `Id_2_max = Id_2` row-by-row). The `filter` form is just the cleanest way to make the wrong behavior observable as wrong filtering rather than a wrong added column.
> 2. **DuckDB transpiler**: window expressions inside a `filter` condition are emitted directly into the SQL `WHERE` clause. They must be wrapped in a subquery / CTE so the comparison happens outside `WHERE`, e.g.:
>    ```sql
>    SELECT * FROM (
>      SELECT *, MAX("Id_2") OVER (PARTITION BY "Id_1") AS __w FROM "DS_1"
>    ) WHERE "Id_2" = __w
>    ```

Fixing either engine in isolation will not make the script work end-to-end; both code paths need to be addressed.

mla2001:
## Summary

- Pandas: When an analytic op is performed over an existing dataset Id, it renames the output to Id_name_1, therefore, a component with a temporary name is created for the result of the analytic operation.
- Duckdb: Window queries (analytic queries) cannot be set on a WHERE clause, so its set as a subquery instead. 

## Checklist

- [x] Tests pass (`pytest`)


javihern98 (approved):
Looks good, thanks!😊

PR #739, opened 2026-05-18
Bump version 1.6.10 with branch-aware release workflow

javihern98:
## Summary

Release 1.6.10 from the `1.6.X` LTS branch, with a refactored release pipeline that lets `1.6.X` carry its own publish/docs recipe rather than depending on `main`'s C++/cibuildwheel flow.

- **`create-release.yml`** — anchors the tag to the selected branch via `target_commitish`, scopes previous-release lookup and pre-release cleanup to commits reachable from the branch's HEAD, uses `GITHUB_SHA` in the compare API (literal `HEAD` was resolving to the default branch), and replaces the `repository_dispatch` fan-out with two `createWorkflowDispatch` calls keyed to the just-created tag. Removes the empty-notes early-return so 0-PR releases still trigger cleanup and dispatch.
- **`release.yml`** — switches publish to OIDC trusted publishing via `pypa/gh-action-pypi-publish` with `environment: pypi`, sets the checkout `ref` to the released tag, and adds a `workflow_dispatch` trigger so the tag-scoped dispatch from `create-release.yml` can invoke it. Keeps the existing poetry build (pure-Python wheel) — the 1.6.X branch does not adopt the C++ parser.
- **`docs.yml`** — sets the checkout `ref` to the released tag so docs build from the tagged commit instead of the default branch's tip.

Mirrors the main-side changes in #737 (checkout the released tag) and #738 (dispatch by tag).

## Checklist

- [x] Code quality checks pass (workflow YAML only — no Python touched)
- [ ] Tests pass — N/A
- [x] Documentation updated — N/A (no user-facing API change)

## Impact / Risk

- Breaking changes? None for users. CI behavior changes:
  - Releases from `1.6.X` now tag against `1.6.X`'s HEAD and not the default branch.
  - Pre-release cleanup is scoped to the branch — pre-releases on other branches are no longer affected when a stable on this branch is published.
  - Release notes enumerate commits actually on this branch.
- Notes for release/changelog? This is the **1.6.10** release. Test runs (rc1, rc2, rc3 on PyPI) validated the full create-release → dispatch → publish flow end-to-end on this branch; the published wheel was confirmed byte-identical to `src/vtlengine/` at the tag.

## Notes

- Closes the loop with main-side PRs #737 + #738. With those merged on `main`, this branch's `release.yml`/`docs.yml` are the ones that run at the released tag (poetry + OIDC), instead of falling back to `main`'s cibuildwheel pipeline.
- History contains rc2/rc3 testing commits; squash-on-merge will collapse them. The final tree state is what matters.
- After merge, the test-only branch policy entry added to the `pypi` GitHub Environment for `workflow-1.6.X-changes` can be removed.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -14,6 +14,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       - name: Create release with notes
         uses: actions/github-script@v8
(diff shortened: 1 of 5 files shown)


mla2001 (approved):
Looks fine! 😊

PR #745, opened 2026-05-19
Fix #744: Resolve docs versions list refs through origin/ fallback

### LINKED ISSUE #744: Bug: docs build fails with 'invalid reference: main' when dispatched from a tag
### Summary

The `Publish documentation` workflow fails with `fatal: invalid reference: main` when triggered via `workflow_dispatch` on a release tag (e.g. `v1.8.0`).

[Failed run #26095062718](https://github.com/Meaningful-Data/vtlengine/actions/runs/26095062718) on tag `v1.8.0`:

```
=== Building main -> _site/main ===
fatal: invalid reference: main
subprocess.CalledProcessError: Command '['git', 'worktree', 'add', '--detach',
  '/tmp/vtlengine-docs-main-.../wt', 'main']' returned non-zero exit status 128.
```

`docs/scripts/build_multiversion.py` calls `git worktree add --detach <wt> <ref>` with bare branch names taken from `_versions.json`. When `actions/checkout@v6` checks out a tag, HEAD is detached and no local `main` branch exists — only the remote-tracking ref `refs/remotes/origin/main`. `git worktree add` does not DWIM a bare branch name to its remote-tracking counterpart, so it fails. Tag refs resolve fine; only branch names break.

This is a regression introduced by #737 + #738, which together made `docs.yml` check out the released tag and dispatch the workflow from that tag. `v1.8.0` is the first release where the new `build_multiversion.py` (which replaced `sphinx-multiversion`) ran from a detached-tag checkout.

### Reproducible Example

CI/workflow bug — not reproducible via the `vtlengine` Python API. The following bash reproduces the failure against a local clone:

```bash
mkdir /tmp/docs-repro && cd /tmp/docs-repro
git init -q
git remote add origin <path-or-url-to-vtlengine>
git fetch --prune origin '+refs/heads/*:refs/remotes/origin/*' '+refs/tags/*:refs/tags/*'
git checkout --detach v1.8.0

git worktree add --detach /tmp/docs-repro-wt main
# -> fatal: invalid reference: main  (exit 128)

git worktree add --detach /tmp/docs-repro-wt origin/main
# -> succeeds; HEAD at the main commit
```

### Expected vs Actual

- **Expected:** `workflow_dispatch` on a release tag publishes documentation for `main` plus the latest stable tags.
- **Actual:** The build step fails on the very first version (`main`) with `fatal: invalid reference: main`; nothing is deployed.

### vtlengine version

1.8.0

### Additional Context

> [!NOTE]
> The 1.6.X docs workflow is unaffected — it still uses `sphinx-multiversion`, which handles refs differently.

Suggested fix: in `docs/scripts/build_multiversion.py`, resolve each ref before invoking `git worktree add` — try the bare ref first, fall back to `origin/<ref>` if it doesn't resolve. Keeps the fix contained to docs tooling and doesn't require changes to `docs.yml` checkout/dispatch logic.

javihern98:
## Summary

Fix the regression that caused [`Publish documentation` run #26095062718](https://github.com/Meaningful-Data/vtlengine/actions/runs/26095062718) to fail on `workflow_dispatch` against tag `v1.8.0` with:

```
=== Building main -> _site/main ===
fatal: invalid reference: main
git worktree add --detach /tmp/vtlengine-docs-main-.../wt main → exit 128
```

`docs/scripts/build_multiversion.py` passes bare branch names (e.g. `main`) from `_versions.json` straight to `git worktree add --detach`. When the workflow checks out a tag (post #737/#738), HEAD is detached and `main` only exists as `refs/remotes/origin/main` — `git worktree add` does not DWIM that resolution, so the build aborts on the first iteration. Tag refs were unaffected.

The fix adds `resolve_worktree_ref()` which tries the bare ref first, then falls back to `origin/<ref>`. Tags continue to resolve as-is; branch names get the `origin/` prefix only when no local branch is present.

## Checklist

- [x] Tests pass (`pytest`) — unchanged; helper exercised by the docs build itself

## Impact / Risk

- No API or behavior change for end users.
- Locally-resolvable refs (current behavior for `pull_request closed` runs) take the same path as before — the fallback only kicks in when the bare ref doesn't resolve.
- No changes to `docs.yml`; the intent of #737/#738 (tag-pinned dispatch) is preserved.

## Notes

Reproduced and validated locally against a detached-HEAD clone at `v1.8.0`:

```
git worktree add --detach <wt> main         # fatal: invalid reference: main
git worktree add --detach <wt> origin/main  # success
```

`resolve_worktree_ref()` returns `origin/main` for `main`, `v1.8.0` for `v1.8.0`, and raises a clear `SystemExit` for fully-unresolvable refs.

After this merges, re-running `Publish documentation` via `workflow_dispatch` (from `main` or any branch carrying the fix) should publish `main` plus the latest stable tags as configured by `configure_doc_versions.py`.

Fixes #744

### DIFF
--- a/docs/scripts/build_multiversion.py
+++ b/docs/scripts/build_multiversion.py
@@ -60,6 +60,25 @@ def overlay_current_docs(worktree: Path) -> None:
             shutil.copy2(src, dst)
 
 
+def resolve_worktree_ref(ref: str) -> str:
+    """Return a ref name that ``git worktree add`` can resolve.
+
+    When the CI checkout is a detached HEAD at a tag, branch names like ``main``
+    only exist as remote-tracking refs (``origin/main``); ``git worktree add``
+    does not DWIM a bare branch name to its remote counterpart. Tries the ref
+    as-is first, then falls back to ``origin/<ref>``.
+    """
+    for candidate in (ref, f"origin/{ref}"):
+        result = subprocess.run(  # noqa: S603
+            ["git", "rev-parse", "--verify", "--quiet", f"{candidate}^{{commit}}"],  # noqa: S607
+            cwd=REPO_ROOT,
+            capture_output=True,
+        )
+        if result.returncode == 0:
+            return candidate
+    raise SystemExit(f"Cannot resolve git ref {ref!r} (tried {ref!r} and {f'origin/{ref}'!r})")
+
+
 def build_one_version(
     ref: str,
     out_dir: Path,
@@ -67,10 +86,11 @@ def build_one_version(
     latest_stable: Optional[str],
 ) -> None:
     """Check out `ref` into a temp worktree, overlay current docs, and run sphinx-build."""
+    resolved_ref = resolve_worktree_ref(ref)
     with tempfile.TemporaryDirectory(prefix=f"vtlengine-docs-{ref.replace('/', '_')}-") as tmp:
         worktree = Path(tmp) / "wt"
         subprocess.run(  # noqa: S603
-            ["git", "worktree", "add", "--detach", str(worktree), ref],  # noqa: S607
+            ["git", "worktree", "add", "--detach", str(worktree), resolved_ref],  # noqa: S607
             cwd=REPO_ROOT,
             check=True,
         )


albertohernandez1995 (approved):
Nice job

PR #762, opened 2026-05-27
Implemented #288: Nullability over scalars

### LINKED ISSUE #288: Implement nullability over scalars
### Summary

Currently, the **datastructures** of some components may differ between semantic execution and full execution. 

This is due to how **nullability** is calculated in operations such as `dataset[calc measure := scalar | constant]`. This is because of how **nullability** is calculated in some RoleSetter operators. For example, in the **Mesure** operator, among the various checks, it is verified that the value of `scalar | constant` is not `None`. 

```python
if cls.role == Role.IDENTIFIER or operand.value is not None:
     nullable = False
```

However, in the `validate` checks, it is always set to `None`, and in the `evaluate` checks, a different value can be assigned to the **Scalar**, causing the **nullability** to differ between executions.

A **nullability** must be implemented internally in the **Scalars** model so that the structure and **nullability** are carried over correctly.

The **nullability** of a **Scalar** input will be taken as `True` by default, and that of **Constants** as `False` (except for explicitly **Null Constants**).

mla2001:
## Summary

Implemented nullability over scalars

## Checklist

- [ ] Tests pass (`pytest`)


albertohernandez1995 (approved):
Excelent job mate☺️

javihern98 (approved):
Looks good, thanks! :blush:

PR #783, opened 2026-06-01
Fix #771: Apply viral attribute propagation during DuckDB execution

### LINKED ISSUE #771: Viral attribute propagation rules are not applied during execution (DuckDB path)
### Initial Checks

### Summary

`define viral propagation` rules (engine support added in #631 / Fix #510; VTL 2.2 spec merged in [sdmx-twg/vtl#701](https://github.com/sdmx-twg/vtl/pull/701)) are parsed, registered, and semantically validated, but their value-combination logic is **never applied during execution**.

The DuckDB transpiler emits no SQL for viral attributes, so an operator's result *declares* the viral-attribute component in its structure while the executed data **omits the column**. The resolvers `ViralPropagationRegistry.resolve_pair` / `resolve_group` (`src/vtlengine/ViralPropagation/__init__.py`) currently have **zero callers**, and the end-to-end tests are skipped with *"not yet implemented in the DuckDB execution path"* (`tests/ViralAttributes/test_viral_propagation.py`, `tests/ViralAttributes/test_viral_operators.py`).

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = """
    define viral propagation CONF (variable VAt_1) is
        when "C" then "C";
        when "N" then "N";
        else "F"
    end viral propagation;
    DS_r <- DS_1 + DS_2;
"""

ds = {
    "name": "DS_1",
    "DataStructure": [
    ],
}

data_structures = {"datasets": [ds, {**ds, "name": "DS_2"}]}

datapoints = {
    "DS_1": pd.DataFrame({"Id_1": [1, 2, 3], "Me_1": [10.0, 20.0, 30.0], "VAt_1": ["C", "N", "F"]}),
    "DS_2": pd.DataFrame({"Id_1": [1, 2, 3], "Me_1": [5.0, 15.0, 25.0], "VAt_1": ["N", "F", "F"]}),
}

print("components:", list(run_result["DS_r"].components.keys()))
```

### Expected vs Actual

**Expected** — per the propagation rule (`C+N → C`, `N+F → N`, `F+F → F`):

| Id_1 | Me_1 | VAt_1 |
| ---- | ---- | ----- |
| 1 | 15.0 | C |
| 2 | 35.0 | N |
| 3 | 55.0 | F |

**Actual** — the result declares a `VAt_1` component, but the executed data omits the column:

```
components: ['Id_1', 'Me_1', 'VAt_1']
   Id_1  Me_1
0     1  15.0
1     2  35.0
2     3  55.0
```

### vtlengine version

1.9.0rc1

### Python version

3.12.3

### OS

Ubuntu 24.04.4 LTS

### Additional Context

> [!NOTE]
> Scope of this issue is the **DuckDB execution support** only: generate per-rule SQL in the transpiler so viral-attribute values are combined per the registered rule, at every operator that propagates them:
> - DS–DS binary operators (`_build_ds_ds_binary`)
> - Aggregation (`visit_Aggregation`)
> - Join operators (`visit_JoinOp`)
> - Analytic invocation (`_visit_analytic_dataset`)

Approach: a `CASE` expression for enumerated binary combining, native `MIN`/`MAX`/`SUM`/`AVG` for aggregate rules, and `list_reduce(list(col), …)` for N-ary enumerated reduction — verified to match the existing `resolve_pair` / `resolve_group` Python oracle.

> [!IMPORTANT]
> Out of scope (to be tracked separately): validation-only fixes (the dead-code `1-3-3-2` duplicate-value-domain check and a multiple-aggregate-clause guard) and value-domain-level rules (which require a `Component.value_domain` association — a public input-schema change).

Acceptance: the currently-skipped end-to-end tests in `tests/ViralAttributes/test_viral_propagation.py` and `tests/ViralAttributes/test_viral_operators.py` are un-skipped and pass.

javihern98:
## Summary

Makes `define viral propagation` rules actually combine viral-attribute **values** during execution. Previously the rules were parsed, registered and semantically validated, but their resolution logic (`resolve_pair`/`resolve_group`) had zero callers — so an operator's result *declared* a viral-attribute component while the executed data **dropped the column**.

Following the engine's SQL-native convention, this generates per-rule DuckDB SQL at transpile time (verified equivalent to the `resolve_pair`/`resolve_group` Python oracle):

- New `ViralPropagation/sql.py` emitters: a `CASE` expression for enumerated binary combining, native `MIN`/`MAX`/`SUM`/`AVG` for aggregate rules, and `list_reduce(list(col), …)` for N-ary enumerated reduction.
- `ViralPropagationRegistry.rule_for(component)` — variable-level overrides value-domain-level (value-domain-ready).
- Wired into every in-scope propagation site:
  - **DS–DS binary** operators (`_build_ds_ds_binary`)
  - **Aggregation** (`visit_Aggregation`)
  - **Analytic invocation** (`_visit_analytic_dataset`, partition-only window so the rule aggregates the whole partition)
  - **Joins** — same-named viral attributes now **merge** into one column via the rule, driven by a single shared predicate (`merged_viral_attribute_names`) used by all three structure producers (`merge_components`, `_build_join_structure`, `visit_JoinOp`).

Fixes #771

## Checklist

- [x] Tests pass (`pytest` — 4893 passed, 11 skipped, 0 failed; viral end-to-end tests un-skipped)
- [ ] Documentation updated (not applicable — no public API/schema change in this scope)

## Impact / Risk

- **Behavior change (joins):** when a viral attribute appears in 2+ join operands (viral in all of them), it now merges into a single unqualified column combined via the rule, instead of being `#`-qualified (`DS_1#VAt_1`/`DS_2#VAt_1`). Mixed-role shared components and non-viral components are unchanged. Verified `components == data.columns` across inner/left/full/cross/3-way joins and join bodies with `calc`/`rename`.
- **Behavior change (all in-scope operators):** viral attributes that were silently dropped from executed data are now produced and combined per the registered rule (no rule → NULL when combining; per-row passthrough for analytic).
- Data/SDMX compatibility: input/output structure JSON unchanged.
- Release/changelog: worth a note that viral propagation now executes.

## Notes

- Implementation approach (native SQL generation) and the verified DuckDB constructs are validated against the Python oracle in `tests/ViralAttributes/test_vp_sql.py`.
- **Deliberately out of scope** (suggested follow-ups):
  - Validation-only fixes: dead-code duplicate-value-domain check (`1-3-3-2`) and a multiple-aggregate-clause guard.
  - Value-domain-level rules (needs a `Component.value_domain` association — a public input-schema change). When added, also align `resolve_pair`/`resolve_group` (currently variable-level only) with `rule_for`'s value-domain fallback so the oracle stays faithful.
  - Unary / dataset-scalar viral passthrough (pre-existing on `main`: `abs(DS_1)`, `DS_1 * 10` declare the viral attr but drop it).


mla2001 (approved):
Looks fine! 😊

PR #794, opened 2026-06-02
Fix #792: Fall back to Python regex for match_characters patterns RE2 can't compile

### LINKED ISSUE #792: RE2 compilation failure when using lookahead assertions in `match_characters` with DuckDB engine
### Summary

When switching to the DuckDB engine (`use_duckdb=True`), the execution of the `LEI_VALIDATIONS` VTL script fails for the Canadian postal-code rule (`pc22`, `COUNTRY_INCORPORATION = "CA"`). This happens because its `match_characters` regex pattern uses lookahead assertions (`^((?=[^DdFfIiOoQqUu\d\s])[A-Z]\d(?=[^DdFfIiOoQqUu\d\s])[A-Z]\s{1}\d(?=[^DdFfIiOoQqUu\d\s])[A-Z]\d)$`), which are supported by the legacy Pandas engine (Python `re` module) but rejected by DuckDB's underlying Google RE2 library. As RE2 does not support lookarounds to ensure linear time complexity, it throws the following error when hitting the query `CREATE TABLE "pc22" AS <sql...>`: `invalid perl operator: (?=`.

The following script was used:
```
pc22 <- DS_1[filter COUNTRY_INCORPORATION = "CA"]
            [calc PC_VALID := match_characters(POSTAL_CODE, "^((?=[^DdFfIiOoQqUu\d\s])[A-Z]\d(?=[^DdFfIiOoQqUu\d\s])[A-Z]\s{1}\d(?=[^DdFfIiOoQqUu\d\s])[A-Z]\d)$")];

```

### Scope

_No response_

### Testing

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`match_characters` regexes that use lookaround assertions (and other PCRE-only constructs) crashed under the DuckDB engine with `invalid perl operator: (?=`. The cause is RE2 (DuckDB's regex engine), which deliberately doesn't support lookaround, backreferences, atomic groups or conditionals — features the legacy Python `re` engine handled fine.

The fix keeps the fast native `regexp_full_match` for patterns RE2 can compile, and only falls back to a small Python `re` UDF (`vtl_match_characters`) for the literal patterns it can't. So nothing changes for existing patterns — we just stop crashing on the ones RE2 rejects.

Fixes #792.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- No breaking changes — RE2-compatible patterns produce identical SQL/results; the fallback only kicks in where execution previously errored.
- No SDMX/data compatibility concerns.
- The UDF is registered on every DuckDB connection but has zero cost unless actually called.

## Notes

- The regex helpers (UDF + RE2-incompatibility detector) live in `Transpiler/operators.py`, next to where `match_characters` maps to SQL.
- Possessive quantifiers aren't auto-detected (too ambiguous to spot reliably); those still surface RE2's own error rather than silently misbehaving.
- Regression test `GL_792` uses the exact `pc22` script from the issue; unit tests cover the detector edge cases and the UDF/NULL handling.


mla2001 (approved):
Looks fine! 😊

PR #588, opened 2026-03-11
Replace Python ANTLR parser with C++ parser via pybind11

javihern98:
## Summary

- Replace `antlr4-python3-runtime` with a C++ ANTLR parser exposed via pybind11, achieving **95.7% performance improvement** (9.6s → 0.41s for 8000 statements)
- Switch build backend from `poetry-core` to `scikit-build-core` for C++ extension compilation
- Update CI workflows to build platform-specific wheels with `cibuildwheel` for Python 3.9-3.13 on Linux, macOS, and Windows

### C++ Parser (`src/vtlengine/AST/Grammar/_cpp_parser/`)
- `bindings.cpp`: pybind11 module with lazy-wrapping `ParseNode`/`TerminalNode` classes
- `_rule_constants.py`: 223 `(rule_index, alt_index)` constants for isinstance replacement
- Generated C++ parser/lexer from `Vtl.g4` (ANTLR 4.13.1)

### ASTConstructor Refactoring
- All `isinstance(ctx, Parser.XxxContext)` → `ctx.ctx_id == RC.XXX`
- All `isinstance(x, TerminalNodeImpl)` → `x.is_terminal`
- All `x.getSymbol().text` → `x.text`
- All `list(ctx.getChildren())` → `ctx.children`

### Build System
- `pyproject.toml`: scikit-build-core backend, cibuildwheel config
- `CMakeLists.txt`: builds pybind11 extension with vendored ANTLR4 C++ runtime
- `scripts/setup_antlr4_runtime.sh`: downloads ANTLR4 C++ runtime for development
- MSVC compatibility: `ANTLR4CPP_STATIC` define, `/FI"chrono"` for ProfilingATNSimulator

### CI Workflows
- **testing.yml**: C++ parser wheel cached via `actions/cache` (keyed on OS, Python version, and source hash). On cache hit, ANTLR download and C++ compilation are skipped entirely. Poetry dependency cache via `setup-python`
- **ubuntu_test_24_04.yml**: Same wheel caching strategy. Uses `--no-deps` to avoid numpy/pandas conflicts with system packages
- **version.yml**: Simplified to pure bash — extracts versions from source files with `grep`/`tomllib`, no build required
- **release.yml**: Wheels-only publishing via `cibuildwheel` (no sdist)

### CI Performance (Testing workflow)
| Run type | Duration |
|----------|----------|
| No cache (first build) | ~7m37s |
| C++ wheel cached | ~5m10s |

### Deleted Files
[...]

PR #795, opened 2026-06-02
Fix #785 (1.9.X): Stop swallowing parse errors in create_ast_with_comments

### LINKED ISSUE #785: (1.9.X) HRuleset with code items starting with digits silently produces empty AST
### Initial Checks

### Summary

When parsing a define hierarchical ruleset expression where the code items (rule identifiers on the left/right side of ruleset equations) start with a digit, create_ast_with_comments silently returns an empty Start node with no children instead of raising a parse error.

This is a regression — the same expression parsed correctly in v1.6.11.

### Reproducible Example

```python
from vtlengine.AST.ASTComment import create_ast_with_comments

# Fails silently (code item starts with digit)
vtl = """define hierarchical ruleset MY_HR (valuedomain rule SECTOR) is
    10A = 10B + 10C
end hierarchical ruleset;"""

ast = create_ast_with_comments(vtl)
print(len(ast.children))   # 0  ← expected: 1

# Works fine (code item starts with letter)
vtl_ok = """define hierarchical ruleset MY_HR (valuedomain rule SECTOR) is
    A10 = B10 + C10
end hierarchical ruleset;"""

ast_ok = create_ast_with_comments(vtl_ok)
print(len(ast_ok.children))  # 1  ✓
```

### Expected vs Actual

Expected:
- The AST is produced correctly (if digit-starting code items are valid in VTL 2.2), or
- A ParseError / SyntaxError is raised with a descriptive message
Current:
create_ast_with_comments returns Start(children=[]) — no exception, no warning, silent failure.

### vtlengine version

1.9.0rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

1.9.X backport of the #785 fix (main: #791).

`prettify()` / `create_ast_with_comments()` silently returned an empty AST when a script failed to parse — e.g. a hierarchical ruleset with code items starting with a digit (`10A = 10B + 10C`). The culprit was a broad `try/except Exception` in `create_ast_with_comments` that swallowed every error from `create_ast`.

`create_ast` already raises a descriptive `VTLSyntaxError` (with line/column + caret); it was just being eaten. The original intent of that block (back in #408) was only to handle comment-only scripts gracefully — that got over-broadened into "swallow everything" during the C++ parser migration (#588).

This drops the bare `except` and lets real errors propagate. Comment-only/empty scripts still work via the existing `if not ast.children` check.

For the record: unquoted digit-starting identifiers aren't valid VTL 2.2 (verified against the official grammar — `ID_PART` must start with a letter/underscore; in VTL 2.1 the grammar still allowed digit-starting names, which is why this worked before the 2.2 alignment). The correct form is the single-quoted `'10A'`, which parses fine.

Closes #785

## Checklist

- [x] Tests pass (`pytest`) — 5015 passed, added a regression test
- [ ] Documentation updated (if applicable) — n/a

## Impact / Risk

- **Behavior change:** `prettify()` / `create_ast_with_comments()` now raise instead of returning an empty AST on unparseable scripts. They also now surface `SemanticError` (dependency cycles, duplicate assignments) that were previously swallowed — these never actually formatted before (they returned empty), so no capability is lost.
- `create_ast_with_comments` is part of the public API, so this is a (correctness-improving) behavior change worth a changelog note.
- `run` / `semantic_analysis` / `run_sdmx` are unaffected — they use `create_ast` directly, which already raised.

## Notes

Out of scope: whether `prettify` (a formatter) should tolerate parseable-but-semantically-invalid scripts. That's a pre-existing coupling (`prettify` → `create_ast` → DAG analysis), not introduced here.

### DIFF
--- a/src/vtlengine/AST/ASTComment.py
+++ b/src/vtlengine/AST/ASTComment.py
@@ -55,12 +55,11 @@ def create_ast_with_comments(text: str) -> Start:
 
     comments = [generate_ast_comment(c) for c in comment_tokens]
 
-    # Try to parse: if no statements, it's only comments
-    try:
-        ast = create_ast(text)
-        if not ast.children:
-            ast = Start(line_start=1, line_stop=1, column_start=0, column_stop=0, children=[])
-    except Exception:
+    # Parse the statements. A script with no statements (only comments or empty) yields a
+    # Start node with no children; a malformed script raises VTLSyntaxError, which must
+    # propagate instead of being silently swallowed into an empty AST.
+    ast = create_ast(text)
+    if not ast.children:
         ast = Start(line_start=1, line_stop=1, column_start=0, column_stop=0, children=[])
 
     ast.children.extend(comments)
--- a/tests/AST/test_AST.py
+++ b/tests/AST/test_AST.py
@@ -45,7 +45,7 @@ from vtlengine.AST import (
 from vtlengine.AST.ASTEncoders import ComplexDecoder, ComplexEncoder
 from vtlengine.AST.ASTTemplate import ASTTemplate
 from vtlengine.DataTypes import ScalarType
-from vtlengine.Exceptions import SemanticError
+from vtlengine.Exceptions import SemanticError, VTLSyntaxError
 from vtlengine.Interpreter import InterpreterAnalyzer
 
 base_path = Path(__file__).parent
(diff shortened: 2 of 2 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #797, opened 2026-06-02
Fix #796: Membership operator no longer mutates its operand's structure

### LINKED ISSUE #796: DuckDB transpiler emits invalid SQL (str_var column) when union combines a single-measure count(ds # comp ...) with a multi-measure count(ds ...)
### Summary

When running `vtlengine` (v1.8.0) with the DuckDB engine (`use_duckdb=True`), a VTL script that performs a `union` of two different `count` paths produces invalid SQL and crashes with a `BinderException: Referenced column "str_var" not found in FROM clause!`. In contrast, the legacy Pandas engine (`use_duckdb=False`) executes successfully and returns the correct results.

### Root Cause & Trigger
The failure occurs when a union combines:
1. A count over a single-measure dataset created via the membership operator: `count(ds # LEI group by X)`
2. A count over a multi-measure dataset without membership: `count(ds group by X)`

While both aggregations work fine in isolation on DuckDB, their union fails. The multi-measure count is transpiled into a `COUNT(CASE WHEN <all measures> IS NOT NULL THEN 1 END)` using a generic internal measure name (`str_var`). When unioning both branches, this placeholder placeholder is not properly projected by the corresponding subquery's `FROM` clause, leading to the unbound column reference in the generated SQL.

### Environment
- **vtlengine:** 1.8.0
- **duckdb:** 1.4.4
- **pysdmx:** 1.16.0
- **pandas:** 2.3.3
- **Python / OS:** Python 3.12.3 / Linux

### Minimal Reproducible Example
```python
from vtlengine import run

script = '''
cd := DS[calc identifier CI := nvl(CI,""), identifier CH := nvl(CH,"")];
a  := count(cd # ID group by CI)[calc identifier T := "A"][rename int_var to V, CI to C];
c  := count(cd[filter CI <> CH] group by CI)[calc identifier T := "B"][rename int_var to V, CI to C];
res <- union(a, c);
'''
# DS has identifier ID, measures M1..M5 (some nullable), plus CI/CH promoted to identifiers in `cd`.
run(script, data_structures=..., datapoints={"DS": df}, use_duckdb=False)  # OK
run(script, data_structures=..., datapoints={"DS": df}, use_duckdb=True)   # BinderException: "str_var" not found

### Scope

_No response_

### Testing

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`count(ds ...)` over a multi-measure dataset crashed under DuckDB with
`BinderException: Referenced column "str_var" not found` when the same
dataset had earlier been used in a `ds # comp` membership.

The cause wasn't the transpiler — it was `Membership.validate`, which
promoted an identifier/attribute target to a generic-named measure
(`str_var`, `int_var`, …) by writing it straight into the **shared**
operand's components. That phantom measure leaked into the intermediate
dataset's structure, so a later `count(...)` picked it up and emitted SQL
referencing a column that doesn't exist.

Fix: build the promoted measure as a standalone component in the result
instead of mutating the operand. Behaviour is otherwise unchanged.

Fixes #796

## Checklist

- [x] Tests pass (`pytest`) — 4828 passed

## Impact / Risk

- No breaking changes — membership output structure is identical, just no
  longer leaks into the source dataset.
- No SDMX/data compatibility concerns.

## Notes

Added regression test `SetBugs::test_GH_796` (fails before the fix with the
exact `str_var` BinderException, passes after). It also asserts the
intermediate dataset keeps only its real measures.


javihern98 (approved):
Looks good, thanks! :blush:

PR #799, opened 2026-06-02
Fix #798: correct SDMX input mapping for inline filter + membership

### LINKED ISSUE #798: DAG analyzer misclassifies [filter ...] components as input datasets when followed inline by membership '#'
## Summary
The internal DAG analyzer (`_extract_input_datasets` -> `DAGAnalyzer.ds_structure(ast).global_inputs`) incorrectly extracts component names used within a `[filter ...]` clause as standalone input datasets when the filtered dataset is immediately followed by the membership operator `#` inside a `count`. 

This artificially inflates `global_inputs`, causing `run_sdmx`'s `_build_mapping_dict` (which auto-maps only when there is exactly one input dataset) to raise a validation exception.

### Environment & Context
- **vtlengine:** 1.8.0 | **duckdb:** 1.4.4 | **pysdmx:** 1.16.0 | **pandas:** 2.3.3 | **Python/OS:** 3.12.3 / Linux
- Context: Running an SDMX TransformationScheme via `run_sdmx(..., use_duckdb=True)`.

### Reproduction & Evidence
```vtl
cd := DS[calc identifier CI := nvl(CI,""), identifier CH := nvl(CH,"")];
ndh := count(cd[filter CI <> CH] # ID group by CI);   /* inline [filter ...] # ID */
res <- union(/* ... */ ndh);

### Scope

_No response_

### Testing

_No response_

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by albertohernandez1995**
## Summary

Two related bugs that broke `run_sdmx` auto-mapping when a script used the inline
`count(ds[filter a <> b] # comp ...)` pattern.

1. **DAG counted filter components as input datasets.** The membership `#` sets
   an `is_dataset` flag while visiting its left operand; when that operand is a
   filter clause, the flag leaked into the clause children, so the filter's
   component names (`a`, `b`) were reported as inputs. `run_sdmx` only auto-maps
   when there's exactly one input, so this raised `0-1-3-1`. The flag is now
   cleared before walking clause children — inputs come back as just the real
   dataset. The two equivalent forms (split the filter, or put `#` before it)
   were already correct and stay that way.

2. **`0-1-3-1` message typo.** The raise site passed `number_datasets` but the
   template expected `{number_Datasets}`, so building the message raised
   `KeyError` and hid the real error.
   Names aligned and the doc table regenerated.

Fixes #798

## Checklist

- [x] Tests pass (`pytest`)
- [x] Documentation updated (regenerated `error_messages.rst`)

## Impact / Risk

- No breaking changes; only corrects which names are treated as inputs and fixes
  an error message that previously failed to render.
- No data/SDMX compatibility concerns.

## Notes

Regression tests: `inline_filter_membership` DAG case and
`test_build_mapping_dict_multiple_inputs_message` — both fail on the unpatched
code and pass after.

### CODE REVIEW

File: src/vtlengine/AST/DAG/__init__.py [hunk 1]
```diff
@@ -290,10 +290,13 @@ def visit_RegularAggregation(self, node: RegularAggregation) -> None:
         self.visit(node.dataset)
         if node.op in [KEEP, DROP, RENAME]:
             return
+        saved_is_dataset = self.is_dataset
+        self.is_dataset = False
         for child in node.children:
             self.is_from_regular_aggregation = True
             self.visit(child)
             self.is_from_regular_aggregation = False
+        self.is_dataset = saved_is_dataset
```
javihern98, line 299: Why is this needed? This will keep the same memory reference, should use deepcopy

javihern98 (changes requested):
need to review the handling of is_dataset

src/vtlengine/AST/DAG/__init__.py
albertohernandez1995: `is_dataset` is a plain `bool` (`is_dataset: bool = False`), so `saved = self.is_dataset`
copies the value — there's no shared mutable reference, and `deepcopy` wouldn't apply to a bool.

That said, the save/restore isn't needed: the only place `is_dataset` is set to `True` is the
`MEMBERSHIP` branch of `visit_BinOp`, which resets it to `False` right after visiting its left
operand — so the caller already restores it. I've dropped the save/restore and just reset
`self.is_dataset = False` before walking the clause children. That reset is the load-bearing
part: it's what stops the `filter`/`calc` component names being misclassified as input datasets
(removing it makes the regression test fail again).

I agree with this approach.

javihern98: Ok, then lets keep it

javihern98 (approved):
Looks good after review, thanks! :blush:

PR #841, opened 2026-06-15
Fix #832: Different behaviour in time_agg with scalar or literal in semantic analysis

### LINKED ISSUE #832: Different behaviour in time_agg with scalar or literal
### Initial Checks

### Summary

If we use time_agg with literal like this example, works fine.
`DS_l <- avg(Day group all time_agg("A"));` 
 
But if we use a scalar
`DS_l <- avg(Day group all time_agg(sc_1));` 
fail with this error:
`('The Time aggregation operand has to be defined if not used inside an aggregation. Please check transformation with output Dataset avr_m_1', '1-3-2-4')`

### Reproducible Example

```python
DS_l <- avg(Day group all time_agg(sc_1));
```

### Expected vs Actual

Expected: Have same behavior in both cases

### vtlengine version

1.9.0rc6

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

mla2001:
## Summary

Semantic anaysis of time_agg with scalars could fail because the scalar value was taken as None.

## Checklist

- [x] Tests pass (`pytest`)


javihern98 (commented):
I do not understand this fix, lacks of clarity why we are only checking on execution this SemanticError instead of with the only semantic, as well as this arg-type ignore.

mla2001:
> I do not understand this fix, lacks of clarity why we are only checking on execution this SemanticError instead of with the only semantic, as well as this arg-type ignore.

The problem here was that semantic anaysis of time_agg with scalars could fail because the input scalars values were taken as None.

javihern98 (approved):
Approved after clarification, thanks! :blush:

PR #861, opened 2026-06-19
VTL regular names are now case-insensitive

### LINKED ISSUE #633: VTL regular names must be case insensitive
### Initial Checks

### Summary

According to the VTL User Manual, regular names must be case insensitive, so a dataset/scalar/component... name such as `NAME`, `Name` or `name` will be taken as the same VarID.

Image

### vtlengine version

1.6.2

### Python version

Any

### OS

Any

albertohernandez1995:
## Summary

VTL regular names are case-insensitive per the spec, but the engine was treating
them case-*sensitively* — so `DS_r <- ds_1;` or `DS_1[filter ME_1 > 15]` would blow
up during semantic analysis even though `DS_1`/`Me_1` existed. This makes regular
names (datasets, components, scalars, UDOs, hierarchical/datapoint rulesets, value
domains, external routines) resolve case-insensitively across the whole pipeline,
while keeping the casing the user actually wrote in the output.

The core of it is a small `CaseInsensitiveDict` (matches keys case-insensitively but
preserves the original casing for display) backing the symbol tables and
`Dataset.components`, plus shared `normalize_name` / `names_equal` helpers for the
remaining name comparisons. It's wired through both the interpreter (structure
inference) and the DuckDB transpiler (execution), the DAG analyzer, and input
loading. At the API boundary we convert back to plain dicts so the internal type
never leaks to callers or SDMX output.

"Written casing wins" for clauses: `DS_1[calc ME_1 := Me_1 * 2]` produces a single
`ME_1` column, and `rename`/`aggr` behave the same way.

Fixes #633

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- Scripts that were valid before keep working. The one behavioral change is that
  names differing only in case are now the *same* name — e.g. using `A` and `a` as
  two separate datasets is no longer allowed (that's the intended spec behavior).
  One DAG test fixture was updated to reflect this.
- No SDMX/output compatibility concerns: output structures keep the user's casing
  and the API returns plain dicts as before.
- Note for the changelog: regular-name resolution is now case-insensitive.

## Notes

- Quoted/non-regular names are treated case-insensitively too in this version.
  Making them spec-exact (case-*sensitive*) would mean threading a "quoted" flag
  through the parser and all lookups — deferred as a follow-up since it's out of
  scope for #633.


mla2001 (commented):
Looks fine!
It would only be necessary to add a few more VTL tests and verify that the case insensitive behavior of the operands actually works in a normal VTL run, not just check that the dictionary works.

albertohernandez1995:
> Looks fine! It would only be necessary to add a few more VTL tests and verify that the case insensitive behavior of the operands actually works in a normal VTL run, not just check that the dictionary works.

It should be now fine.

javihern98:
Need to be discussed internally before merging

PR #875, opened 2026-07-02
Improve SDMX-VTL mapping and add data_structures/datapoints to run_sdmx

### LINKED ISSUE #874: Map one SDMX Dataflow to multiple VTL datasets and accept data_structures/datapoints in run_sdmx
### Initial Checks

### Problem Description

`run_sdmx()` currently has two limitations when wiring SDMX inputs into a VTL script:

1. **One SDMX Dataflow cannot feed two different VTL datasets.** Internally the mapping is a `Dict[short_urn, vtl_name]` (`_build_mapping_dict` and the conversion loop in `run_sdmx`). Because the dict is keyed by the SDMX short-URN, a single `Dataflow` can only ever resolve to **one** VTL dataset name. The `mappings`/`sdmx_mappings` parameter also only accepts a *single* `VtlDataflowMapping` or a raw `dict` — not a list of `VtlDataflowMapping` nor a pysdmx `VtlMappingScheme` (`_convert_sdmx_mappings`) — so there is no typed way to express "this one dataflow is the source for both `DS_1` and `DS_2`".

   Verified against the test fixtures (`tests/API/data/SDMX/input/`):

   ```text
   short_urn: DataStructure=BIS:BIS_DER(1.0)
   # script references DS_1 and DS_2, dict mapping can only name one:
   InputValidationException: ("Input name ['DS_2'] not found in the input Datasets.", '0-1-3-6')

   # a list of VtlDataflowMapping is rejected outright:
   InputValidationException: Expected dict or VtlDataflowMapping type for mappings.
   ```

2. **`run_sdmx()` cannot mix SDMX inputs with plain VTL-JSON structures / datapoints.** Unlike `run()`, `run_sdmx()` only accepts `datasets: Sequence[PandasDataset]` and derives *both* structure and data from them. There is no way to supply an extra VTL-JSON `data_structures` entry or a CSV/DataFrame `datapoints` entry alongside the SDMX datasets in a single call.

### Proposed Solution

```python
# ── Part 1: allow one Dataflow to map to several VTL datasets ───────────────
#
# Accept a list of VtlDataflowMapping and/or a pysdmx VtlMappingScheme in the
# `mappings` (run_sdmx) / `sdmx_mappings` (run, semantic_analysis) parameter:

mappings: Optional[
    Union[
        VtlDataflowMapping,
        Sequence[VtlDataflowMapping],       # NEW
        VtlMappingScheme,                   # NEW (pysdmx-native container)
        Dict[str, Union[str, List[str]]],   # dict value may be a list of VTL names  # NEW
    ]
] = None

# Two VtlDataflowMapping entries can share the same `dataflow` but use
# different `dataflow_alias`, so the same source feeds DS_1 and DS_2:
mappings = [
    VtlDataflowMapping(dataflow="Dataflow=BIS:BIS_DER(1.0)", dataflow_alias="DS_1", id="m1"),
    VtlDataflowMapping(dataflow="Dataflow=BIS:BIS_DER(1.0)", dataflow_alias="DS_2", id="m2"),
]
# dict form:
mappings = {"DataStructure=BIS:BIS_DER(1.0)": ["DS_1", "DS_2"]}

# Recommended internal change: invert the mapping to Dict[vtl_alias, short_urn]
# (VTL alias is unique per input; short_urn is not), OR make it
# Dict[short_urn, List[str]] and fan out one structure/datapoints copy per alias
# in the run_sdmx conversion loop.

# ── Part 2: run_sdmx accepts data_structures and datapoints like run() ──────
def run_sdmx(
    script: Union[str, TransformationScheme, Path],
    datasets: Sequence[PandasDataset],
    mappings: Optional[...] = None,
    data_structures: Optional[Union[Dict, Path, str, ...]] = None,   # NEW, same type as run()
    datapoints: Optional[Union[Dict, List, str, Path]] = None,       # NEW, same type as run()
    value_domains: Optional[...] = None,
    external_routines: Optional[...] = None,
    time_period_output_format: str = "vtl",
    return_only_persistent: bool = True,
    output_folder: Optional[Union[str, Path]] = None,
    output_format: Literal["csv", "parquet"] = "csv",
) -> Dict[str, Union[Dataset, Scalar]]:
    ...
    # Merge SDMX-derived structures/datapoints with the explicit ones, then
    # forward everything to run(). Define precedence for name collisions
    # (recommend: raise a clear InputValidationException on duplicate names).


This is feasible with low risk: `run_sdmx()` already builds `data_structures_list` + `datapoints_dict` and forwards them to `run()`, and `run()` already accepts a **mixed list** of VTL-JSON dicts / paths / pysdmx objects for `data_structures` and a dict/list for `datapoints`. Part 2 is mostly plumbing + merge/precedence rules. Part 1 requires widening the accepted `mappings` types and changing the mapping structure from 1:1 to 1:many.
```

### Alternatives Considered

- **Pre-duplicating the `PandasDataset`** before calling `run_sdmx` (once per target alias): rejected — the alias still collapses to one `short_urn` key in `mapping_dict`, so the second entry overwrites the first; it does not actually work today.
- **Falling back to `run()` directly** with hand-built VTL-JSON: works but throws away the pysdmx `Schema` → VTL-JSON conversion and the `VtlDataflowMapping`/`VtlMappingScheme` typed inputs that `run_sdmx` exists to provide.
- **Keeping `Dict[short_urn, str]` and requiring distinct URNs**: rejected — it structurally forbids the "one Dataflow → many VTL datasets" use case that this issue is about.

### Additional Context

Affected code (all currently do `short_urn in sdmx_mappings` lookups, so any change to the mapping structure must update them consistently):

- `src/vtlengine/API/__init__.py` — `run_sdmx`, `run`, `semantic_analysis`
- `src/vtlengine/API/_sdmx_utils.py` — `_convert_sdmx_mappings`, `_convert_vtl_dataflow_mapping`, `_build_mapping_dict`
- `src/vtlengine/API/_InternalApi.py` — `load_datasets` / `_load_datastructure_single` mapping lookups
- `src/vtlengine/files/sdmx_handler.py` — `_extract_name_from_structure`, `extract_sdmx_dataset_name`, `load_sdmx_structure`

> [!NOTE]
> pysdmx already exposes `VtlMappingScheme` (an `ItemScheme` of `VtlMapping` items) as the canonical container for multiple `VtlDataflowMapping` objects — accepting it directly keeps the API aligned with the SDMX information model.

> [!IMPORTANT]
> Adding `data_structures`/`datapoints` to `run_sdmx` needs a defined precedence rule for name collisions between explicit inputs and SDMX-derived ones. Recommendation: raise `InputValidationException` on duplicate dataset names rather than silently overriding.

New error codes may be needed (next free in the `0-1-3-x` series is `0-1-3-9`) for: duplicate VTL alias across mappings, and duplicate dataset name between explicit and SDMX-derived inputs.

albertohernandez1995:
## Summary

`run_sdmx` had two rough edges when wiring SDMX inputs into a VTL script:

1. A single SDMX dataflow could only ever map to **one** VTL dataset. If a script
   referenced the same source under two names, there was no way to express it — the
   internal mapping was keyed by short-URN, so the second name was simply lost.
2. Unlike `run`, `run_sdmx` couldn't take extra inputs. If one dataset came from SDMX
   and another from a local CSV, you had to drop down to `run` and hand-build the
   VTL JSON yourself.

This PR fixes both.

**Mapping (one dataflow → many datasets).** `mappings` now accepts, on top of the
existing dict and single `VtlDataflowMapping`:
- a dict whose value is a list of names: `{"Dataflow=MD:DF(1.0)": ["DS_1", "DS_2"]}`
- a sequence of `VtlDataflowMapping` (same dataflow, different `dataflow_alias`)
- a pysdmx `VtlMappingScheme`

Internally the mapping became `{short_urn: [names...]}` and the conversion loop fans
out one structure + datapoints copy per alias. The widened type also applies to
`run`/`semantic_analysis`'s `sdmx_mappings`.

**Explicit inputs.** `run_sdmx` gained optional `data_structures` and `datapoints`
params mirroring `run`. They're merged with the SDMX-derived inputs before execution,
so you can mix an SDMX dataflow with a plain VTL JSON structure + DataFrame/CSV in one
call.

**Guardrails.** Conflicts that used to overwrite data silently now raise:
- `0-1-3-9` — a name is provided by both an SDMX dataset and an explicit input
- `0-1-3-10` — non-dict datapoints combined with SDMX datasets (no name to merge on)
- `0-1-3-11` — the same VTL name is mapped from more than one dataflow

## Checklist

- [x] Tests pass (`pytest` — full suite green)
- [ ] Documentation updated (if applicable) — autodoc picks up the new signatures; a
      worked example is a possible follow-up

## Impact / Risk

- **Breaking changes?** No. The new `run_sdmx` params are optional and default to the
  previous behaviour; the `mappings`/`sdmx_mappings` types are only widened, so existing
  dict / `VtlDataflowMapping` calls are unaffected.
- **Data/SDMX compatibility?** For the shared `run`/`semantic_analysis` structure-loading
  path, a many-name mapping still resolves one SDMX structure to a single name (first
  wins); the one-to-many fan-out is specific to `run_sdmx`.
- **Release notes:** new SDMX mapping capabilities + explicit inputs in `run_sdmx`; three
  new validation error codes (`0-1-3-9`/`10`/`11`).

## Notes

- Tests follow the suite convention: VTL scripts in `data/vtl/`, inputs from files, and
  results compared against committed reference outputs (codes `3-1`, `3-2`).
- Known follow-ups, out of scope here: one-to-many support in `run`/`semantic_analysis`
  for SDMX structure objects, honouring `ToVtlMapping`/`FromVtlMapping` sub-space filters,
  and a docs example.

Closes #874


src/vtlengine/API/_sdmx_utils.py
```diff
@@ -106,12 +177,94 @@ def _build_mapping_dict(
...
-    if isinstance(mappings, VtlDataflowMapping):
-        return _convert_vtl_dataflow_mapping(mappings)
+    # A VTL dataset name may come from exactly one dataflow; duplicates across (or within)
+    # dataflows would overwrite data silently.
+    all_names = [name for names in normalized.values() for name in names]
+    duplicates = sorted({name for name in all_names if all_names.count(name) > 1})
+    if duplicates:
+        raise InputValidationException(code="0-1-3-11", names=duplicates)
+
+    return normalized
```
javihern98, line 198: **Silent data loss when one dataflow feeds multiple datasets** (confidence: high — reproduced)

`mapping_dict` is keyed only by `short_urn`, with no check that multiple entries in `datasets` share a `short_urn`. In `run_sdmx`'s fan-out loop (https://github.com/Meaningful-Data/vtlengine/blob/6bf759813fd08ff200468847474684b8316eabd8/src/vtlengine/API/__init__.py#L622-L628), `datapoints_dict[dataset_name] = dataset.data` runs on every outer iteration, so two `PandasDataset`s sharing a `short_urn` (e.g. two reporting periods of the same dataflow — called out as "a common SDMX situation" in `docs/sdmx_inputs.rst`) silently overwrite each other's data for every aliased name. No exception, no warning.

Reproduced directly: two datasets sharing a urn, one real and one forced to `999.0`, mapped to `["DS_1", "DS_2"]` — both come back `999.0`. This is also the exact scenario issue #874's "Alternatives Considered" section flags as needing per-alias isolation, which this implementation doesn't achieve.

🤖 Generated by Claude Code

src/vtlengine/API/_sdmx_utils.py
```diff
@@ -55,45 +73,98 @@ def _convert_vtl_dataflow_mapping(mapping: VtlDataflowMapping) -> Dict[str, str]
...
-    if mappings is None:
+    normalized = _normalize_mappings(mappings)
+    if normalized is None:
         return None
-    if isinstance(mappings, dict):
-        return mappings
-    if isinstance(mappings, VtlDataflowMapping):
-        return _convert_vtl_dataflow_mapping(mappings)
-    raise InputValidationException("Expected dict or VtlDataflowMapping type for mappings.")
+    return {urn: names[0] for urn, names in normalized.items()}
```
javihern98, line 155: **One-to-many `sdmx_mappings` silently truncated in `run()`/`semantic_analysis()`** (confidence: 75/100)

This keeps only `names[0]` per urn (used by `run()`/`semantic_analysis()`, not `run_sdmx`'s own fan-out), silently dropping the rest — even though the public type hint on both functions (https://github.com/Meaningful-Data/vtlengine/blob/6bf759813fd08ff200468847474684b8316eabd8/src/vtlengine/API/__init__.py#L181 and https://github.com/Meaningful-Data/vtlengine/blob/6bf759813fd08ff200468847474684b8316eabd8/src/vtlengine/API/__init__.py#L280) was widened to accept one-to-many forms. Their docstrings (L219-220, L372-374) weren't updated to mention this. A caller passing a multi-name mapping directly to `run()` gets a confusing "Dataset not found" error instead of a clear signal that only the first alias was used.

🤖 Generated by Claude Code

src/vtlengine/API/_sdmx_utils.py
```diff
@@ -106,12 +177,94 @@ def _build_mapping_dict(
...
+    """Return the dataset/scalar names declared in a VTL JSON structure dict.
+
+    Only VTL JSON dicts expose their names cheaply; for other forms (Path, URL or
+    pysdmx objects) an empty list is returned and no collision check is performed.
+    """
+    if not isinstance(structure, dict):
+        return []
+    names = [ds["name"] for ds in structure.get("datasets", []) if "name" in ds]
+    names += [sc["name"] for sc in structure.get("scalars", []) if "name" in sc]
+    return names
```
javihern98, line 211: **Duplicate-name check bypassed for non-dict `data_structures`** (confidence: 75/100)

This returns `[]` for any non-dict `data_structures` item (`Path`, `str`, `Schema`, `DataStructureDefinition`, `Dataflow`), so `_merge_sdmx_data_structures`'s new duplicate check (`0-1-3-9`) never sees a name to compare for those forms, and a colliding name silently overwrites downstream via `{**ds_structures, **ds}`. Issue #874 (implemented by this PR) explicitly asks to "raise InputValidationException on duplicate dataset names rather than silently overriding" — this only holds for dict-form extras.

🤖 Generated by Claude Code

src/vtlengine/API/_sdmx_utils.py
```diff
@@ -55,45 +73,98 @@ def _convert_vtl_dataflow_mapping(mapping: VtlDataflowMapping) -> Dict[str, str]
...
+    Raises:
+        InputValidationException: If mappings (or one of its items) has an invalid type.
+    """
+    if mappings is None:
+        return None
+    if isinstance(mappings, dict):
+        return {
+            urn: [names] if isinstance(names, str) else list(names)
+            for urn, names in mappings.items()
+        }
```
javihern98: **`IndexError` instead of `InputValidationException` on empty mapping list** (confidence: 75/100)

No check that `list(names)` is non-empty here. `_convert_sdmx_mappings` a few lines below then unconditionally does `names[0]`, so `sdmx_mappings={"urn": []}` — allowed by the widened public type hint `Dict[str, Union[str, List[str]]]` — raises an unhandled `IndexError` instead of the library's normal `InputValidationException` pattern.

🤖 Generated by Claude Code

src/vtlengine/API/__init__.py
```diff
@@ -571,41 +602,46 @@ def run_sdmx(
...
-        datapoints_dict[dataset_name] = dataset.data
-
-    # Validate all script inputs are mapped
-    missing = [name for name in input_names if name not in mapping_dict.values()]
-    if missing:
-        raise InputValidationException(code="0-1-3-6", missing=missing)
+        for dataset_name in mapping_dict[schema.short_urn]:
+            data_structures_list.append(to_vtl_json(schema, dataset_name))
+            datapoints_dict[dataset_name] = dataset.data
+
```
javihern98, line 642: **Empty mapping list silently drops a whole dataset** (confidence: 75/100)

`if schema.short_urn not in mapping_dict` (L623) only checks the key exists, not that its value list is non-empty. `mappings={short_urn: []}` for a dataflow that IS present in `datasets` passes this check, then the `for dataset_name in mapping_dict[schema.short_urn]:` loop runs zero times — the dataset's structure/data vanish with no error. Not covered by any test in `tests/API/test_sdmx.py`.

🤖 Generated by Claude Code

src/vtlengine/Exceptions/messages.py
```diff
@@ -141,6 +141,23 @@
         "description": "Raised when the type of input Datasets is incorrect; "
         "a sequence of PandasDataset is expected.",
     },
+    "0-1-3-9": {
+        "message": "Duplicate dataset name(s) {names} provided both by SDMX datasets and "
+        "explicit inputs.",
+        "description": "Raised when an explicit data structure or datapoints entry uses a "
+        "dataset name that is already provided by one of the SDMX datasets.",
+    },
```
javihern98, line 148: **Message text misattributes the collision source** (confidence: 65/100)

This says duplicates are "provided both by SDMX datasets and explicit inputs," but the check in `_merge_sdmx_data_structures` (https://github.com/Meaningful-Data/vtlengine/blob/6bf759813fd08ff200468847474684b8316eabd8/src/vtlengine/API/_sdmx_utils.py#L232-L239) scans the whole combined list indiscriminately — two colliding names that are both from the `extra` side (no SDMX involvement at all) still raise this same message.

🤖 Generated by Claude Code

src/vtlengine/Exceptions/messages.py
```diff
@@ -141,6 +141,23 @@
...
+        "message": "Explicit datapoints must be a dict of dataset name to data when SDMX "
+        "datasets are provided.",
+        "description": "Raised when non-dict datapoints (list, path or URL) are combined with "
+        "SDMX datasets, which carry no dataset name to merge on.",
+    },
+    "0-1-3-11": {
+        "message": "VTL dataset name(s) {names} are mapped from more than one SDMX dataflow.",
+        "description": "Raised when the mappings assign the same VTL dataset name to more than "
+        "one SDMX dataflow, which would overwrite data silently.",
+    },
```
javihern98, line 178: **Message says "more than one dataflow" but also fires within one dataflow** (confidence: 65/100)

The code comment in `_build_mapping_dict` (https://github.com/Meaningful-Data/vtlengine/blob/6bf759813fd08ff200468847474684b8316eabd8/src/vtlengine/API/_sdmx_utils.py#L183-L190) correctly notes duplicates "across (or within) dataflows" trigger this check, but this message narrows it to "mapped from more than one SDMX dataflow" — misleading when the duplicate is a repeated alias within a single dataflow's own list (e.g. a typo).

🤖 Generated by Claude Code

src/vtlengine/API/__init__.py
```diff
@@ -528,7 +550,16 @@ def run_sdmx(
...
+        data_structures: Optional extra data structures merged with the SDMX-derived ones, \
+        using the same forms accepted by :obj:`run <vtlengine.API>`. (default: None)
+
+        datapoints: Optional extra datapoints merged with the SDMX-derived ones, using the \
+        same forms accepted by :obj:`run <vtlengine.API>`. Must be a dict keyed by dataset \
+        name when SDMX datasets are also provided. (default: None)
 
         value_domains: Dict or Path, or List of Dicts or Paths of the \
         value domains JSON files. (default:None) It is passed as an object, that can be read from \
```
javihern98, line 570: **`Raises:` section not updated alongside this docstring edit** (confidence: 50/100)

The `Args:` section was extended here with `data_structures`/`datapoints`, but the `Raises:` section further down (L590-592, not touched by this diff) still only mentions `SemanticError` and doesn't list the three new codes (`0-1-3-9`, `0-1-3-10`, `0-1-3-11`) this function now raises.

🤖 Generated by Claude Code

javihern98 (commented):
### Code review

This review was generated by Claude Code. It surfaces 10 potential issues found while reviewing the one-to-many SDMX mapping and `run_sdmx` merge logic, each posted as a separate inline comment with a confidence score (0-100).

Only the first (silent data loss when one dataflow feeds multiple datasets via distinct `PandasDataset` objects) was independently reproduced and verified as a real, high-confidence bug. The remaining nine are lower-confidence signals included for completeness at the requester's request — treat anything below ~80 as worth a look rather than confirmed, and the two at 25-50 as likely non-issues on closer inspection.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/vtlengine/API/_sdmx_utils.py
albertohernandez1995: Good catch — this was a real silent overwrite. Two `PandasDataset`s resolving to the same `short_urn` now raise `0-1-3-12` in the fan-out loop instead of clobbering each other's data. Added a test with two datasets of the same dataflow.

src/vtlengine/API/_sdmx_utils.py
albertohernandez1995: Agreed. `_convert_sdmx_mappings` now raises `0-1-3-13` when a single structure is mapped to more than one name, so `run()`/`semantic_analysis()` fail loudly instead of dropping the extra aliases. `run_sdmx` forwards no mapping (its inputs are already resolved to named VTL JSON), and both docstrings now state the one-name-per-structure rule and point to `run_sdmx` for one-to-many.

src/vtlengine/API/_sdmx_utils.py
albertohernandez1995: Fixed — an empty name list now raises `0-1-3-14` in `_normalize_mappings`, before `_convert_sdmx_mappings` ever does `names[0]`.

src/vtlengine/API/__init__.py
albertohernandez1995: Same root cause as the empty-list case. Rejecting empty lists at `_normalize_mappings` (`0-1-3-14`) also prevents the zero-iteration silent drop here. Added a test.

src/vtlengine/Exceptions/messages.py
albertohernandez1995: Reworded to "Duplicate dataset name(s) {names} in the combined SDMX and explicit inputs" — accurate whether the clash is SDMX-vs-explicit or between two explicit entries.

src/vtlengine/Exceptions/messages.py
albertohernandez1995: Reworded to "VTL dataset name(s) {names} are mapped more than once" so it also fits a repeated alias within a single dataflow's list.

src/vtlengine/API/__init__.py
albertohernandez1995: Updated the `Raises:` section to list `InputValidationException` with the new codes (`0-1-3-9`/`10`/`11`/`12`/`14`).

src/vtlengine/API/_sdmx_utils.py
albertohernandez1995: Deliberately keeping the name check to dict-form extras here, for two reasons:

1. The realistic collision is already caught. A duplicate dataset needs both a structure *and* data. Its datapoints are keyed by name, so a `Path`/JSON structure colliding with an SDMX-mapped name (e.g. both `DS_1`) trips the datapoint-level `0-1-3-9` check. A structure-only collision with no matching datapoints is a broken input that fails downstream regardless.
2. Consistency with `run()`. `run()` itself merges its `data_structures` list last-wins and never raises on duplicates for *any* form. Introspecting `Path`/`str`/`Schema`/`DSD`/`Dataflow` here would mean file I/O + mapping resolution inside a merge helper, and would make `run_sdmx` stricter than `run()` for identical inputs.

The docstring documents that only dict-form extras are name-checked. Happy to file a follow-up if we want full-coverage duplicate detection unified across `run()` and `run_sdmx` — it'd belong post-load, on resolved names, not in the merge helper.

PR #878, opened 2026-07-08
Fix #877: viral attribute propagation and rule execution across operators

### LINKED ISSUE #877: Viral attributes dropped by hierarchy, check_hierarchy, check_datapoint, unpivot, period_indicator
### Summary

By the VTL Attribute propagation rule, a viral Attribute (role `Viral Attribute`) must propagate from the operand Data Set(s) to the result Data Set (an Attribute that is viral in at least one operand is viral in the result). There are two related defects.

**1. Five operators drop the viral attribute entirely:**

- `hierarchy` — deletes viral attributes; computed nodes lose them entirely
- `check_hierarchy` — result never carries viral attributes
- `check_datapoint` — result never carries viral attributes
- `unpivot` — viral attribute is dropped by the melt
- `period_indicator` — result keeps only identifiers + `duration_var`

**2. The viral propagation rule is not executed by every dataset-level operator.**

The `define viral propagation` rule must be **executed** by every operator that produces a viral attribute at dataset level, applying the rule's resolution (enumerated mapping / aggregate function / default value) to the value(s) that contribute to each result row:

- Unary operators (`round`, `abs`, `ceil`, …) and dataset-scalar operators (`DS_1 * 10`, …) must execute the rule instead of passing the viral value through unchanged.
- String parameterized operators (`substr`, `replace`, `instr`) must execute the rule too (this was missing in the pandas backend, so the two backends disagreed).
- `check_datapoint` and `check_hierarchy` must likewise execute the rule on the viral value(s) contributing to each validation row.

For row-preserving operators an **aggregate** rule collapses the whole dataset's viral values to a single value applied to every result row; an **enumerated** rule maps each row's value (matching unary clause, else the default). For grouping/combining operators the rule collapses per group / per matched pair.

Today only multi-operand operators (binary, aggregation, analytic, join) execute the rule; single-source / row-preserving operators skip it.

**3. Aggregate viral propagation rules are not type-checked.**

An `aggregate sum` / `aggregate avg` rule targeting a **non-numeric** viral attribute (or value domain) is nonsensical and must raise a `SemanticError`. Today it slips through `semantic_analysis()` and crashes at execution with a cryptic, backend-specific error:

- pandas: `TypeError: unsupported operand type(s) for +: 'int' and 'str'`
- DuckDB: `BinderException: No function matches ... 'sum(VARCHAR)'`

`sum`/`avg` require a `Number`/`Integer` viral attribute; `min`/`max` remain valid for any comparable type (String, Date, Duration, …). The check must apply to both `variable` and `valuedomain` rule signatures.

**4. Every viral attribute must declare a viral propagation rule (`1-3-3-6`).**

A viral attribute that reaches a result Data Set without a `define viral propagation` rule must raise a `SemanticError` (`1-3-3-6`) — a rule is **mandatory, not optional**. This applies to **every** operator, including pure passthrough:

- identity assignment (`DS_r <- DS_1`), `keep` / `drop` / `filter`,
- unary and dataset-scalar operators,
- `calc` that creates a viral attribute,
- as well as the combining operators (aggregation, analytic, hierarchy nodes, binary/join where the attribute is in ≥2 operands).

The rule is enforced **centrally**, once per statement result, so it fires for both backends and at `semantic_analysis()` time (no execution needed). Previously the engine either silently produced `NULL` when combining without a rule, or passed values through unchanged for single-operand cases; both are now rejected. When multiple viral attributes are present, each must have its own rule — the first one lacking a rule is reported by name.

This affects **both** execution backends (`use_duckdb=False` and `use_duckdb=True`).

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = """
    define viral propagation VP1 (variable VAt_1) is aggregate max end viral propagation;
    define hierarchical ruleset H ( valuedomain rule Id_2 ) is
        UGO = AUD + BE
    end hierarchical ruleset;
    DS_r <- hierarchy(DS_1, H rule Id_2 non_null);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

data_df = pd.DataFrame(
    {"Id_1": [1, 1], "Id_2": ["AUD", "BE"], "Me_1": [1.0, 1.0], "VAt_1": [100, 200]}
)

run_result = run(
    script=script,
    datapoints=datapoints,
    return_only_persistent=False,
)

```

### Expected vs Actual

**Actual** — `DS_r` drops `VAt_1` entirely:

| Id_1 | Me_1 | Id_2 |
| ---- | ---- | ---- |
| 1    | 2.0  | UGO  |

**Expected** — `DS_r` carries `VAt_1` as a `Viral Attribute`, and the computed node `UGO` combines its child code items via the rule (`aggregate max`), so `VAt_1 = max(100, 200) = 200`:

| Id_1 | Id_2 | Me_1 | VAt_1 |
| ---- | ---- | ---- | ----- |
| 1    | UGO  | 2.0  | 200   |

The same propagation must hold for `check_hierarchy`, `check_datapoint`, `unpivot` and `period_indicator`, in both the pandas and DuckDB backends. In addition, dataset-level unary/string operators (e.g. `round(DS_1)`, `substr(DS_1, 2)`) and the validation operators `check_datapoint` / `check_hierarchy` must **execute** the viral propagation rule on the contributing viral attribute value(s), not pass them through unchanged.

Finally, an `aggregate sum`/`avg` rule over a non-numeric viral attribute (variable or value domain) must raise a `SemanticError` at semantic-analysis time (currently it reaches a cryptic runtime crash); and **every** viral attribute must declare a `define viral propagation` rule — a viral attribute reaching a result without one raises a `SemanticError` (`1-3-3-6`), for both backends and at semantic-analysis time, regardless of whether the operator combines values or merely passes them through.

### vtlengine version

1.9.0

### Python version

3.12.3

### OS

Ubuntu 24.04

javihern98:
## Summary

Fixes #877. Viral attributes (role `Viral Attribute`) now propagate to the result of every dataset-level operator, and the `define viral propagation` rule is **executed** on the contributing value(s), in both the pandas (default) and DuckDB backends.

- **Structural propagation** for the operators that dropped viral attributes: `hierarchy`, `check_hierarchy`, `check_datapoint`, `unpivot`, `period_indicator`.
- **Rule execution** on every dataset-level operator: aggregate rules (`min|max|sum|avg`) collapse over the operand's values — dataset-wide for row-preserving ops (unary/`round`/`abs`/dataset-scalar/`period_indicator`/`unpivot`), per group/node for aggregating ops (`hierarchy`); enumerated rules map per value. `check_datapoint`/`check_hierarchy` re-attach the datapoint's viral value and apply the rule.
- **Core:** `ViralPropagationRegistry.resolve_single` + `apply_row_preserving` (pandas) and `vp_dataset_wide_sql` (DuckDB); `resolve_group` now applies the rule to single-element groups.
- **Bug fix (DAG):** `DAGAnalyzer.sort_ast` was dropping `ViralPropagationDef` nodes, so the rule was never registered in any multi-statement script and combinations silently produced NULL. The definition is now preserved.

Viral logic stays **out of** the datapoint/hierarchy rule handling: values are captured before the rule computation and re-attached/combined at the final `evaluate` (pandas) or SQL (DuckDB).

Two hierarchy test references (`GL_cs_22`, `GL_463_1`) were corrected: their `filter` operands were previously corrupted by an in-place hierarchy strip (now prevented by the unconditional deep-copy of the hierarchy operand), so `filter` correctly keeps its attributes.

## Checklist

- [x] Tests pass (`pytest`) — full suite on both `pandas` and `duckdb` backends

## Impact / Risk

- **Behavior change:** operators that previously dropped viral attributes now keep them; the propagation rule now executes on row-preserving operators. No public API change.
- The DAG fix changes behavior of any multi-statement script that uses `define viral propagation` (previously the rule was silently not applied — combinations returned NULL).
- Both backends verified and produce matching results.

## Notes

- `filter` and `round` already propagated viral attributes structurally; the rule-execution layer is what's new for row-preserving operators.
- Non-viral attribute handling is unchanged (non-viral attributes are still dropped by combining/aggregating operators).


mla2001 (approved):
Looks fine! 😊

PR #880, opened 2026-07-08
Fix #879: strict, cross-backend-consistent Date/datetime validation

### LINKED ISSUE #879: Date validation accepts partial time components and reports misleading errors for malformed times
### Summary

When a `Date` value carries a time component, the engine accepts an **incomplete** time (`HH` or `HH:MM`) and silently zero-fills the missing parts instead of rejecting it. Per ISO 8601 / VTL, a time component must be a complete `HH:MM:SS` (optionally with fractional seconds and/or a timezone suffix).

Separately, when a datetime is genuinely malformed (out-of-range hour, bad separator), the load error misleadingly says the value is **"out of range for the month"** even when the month and date are perfectly valid — the real problem is the time component.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = "DS_A <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

# 1) Truncated time (missing seconds) is silently accepted and zero-filled
dp = {"DS_1": pd.DataFrame({"Id_1": [1], "Me_1": ["2020-01-01T12:30"]})}
res = run(script=script, data_structures=data_structures, datapoints=dp)
print("Case 1 (missing seconds) ->", res["DS_A"].data["Me_1"].tolist())

# 2) Malformed time reports a misleading "out of range for the month" error
dp = {"DS_1": pd.DataFrame({"Id_1": [1], "Me_1": ["2020-01-01T25:00:00"]})}
try:
    run(script=script, data_structures=data_structures, datapoints=dp)
except Exception as e:
    print("Case 2 (bad hour)        ->", type(e).__name__, "|", e.args[0])
```

### Expected vs Actual

**Expected:**

| Input | Expected |
| ----- | -------- |
| `2020-01-01T12:30` (missing seconds) | Rejected with a clear "incomplete time" error |
| `2020-01-01T25:00:00` (hour out of range) | Error that identifies the **time** as the problem |

**Actual (vtlengine 1.9.0):**

| Input | Actual |
| ----- | ------ |
| `2020-01-01T12:30` | Silently loaded as `2020-01-01T12:30:00` |
| `2020-01-01T25:00:00` | `DataLoadError: ... "Date 2020-01-01T25:00:00 is out of range for the month."` (misleading — the month is valid; the hour `25` is the problem) |

### vtlengine version

1.9.0

### Python version

3.12.3

### OS

Ubuntu 24.04 (Linux 6.17)

### Additional Context

- Affects the pandas load path (`check_date`), the cast path (`check_max_date`), and the DuckDB DataFrame-register path.
- On the DuckDB DataFrame path, a bad-separator value such as `2020-01-01X12:30:45` was additionally **silently truncated** to `2020-01-01`, and out-of-range times surfaced as `Date unknown is out of range for the month`.
- Full datetimes (`2020-01-01T12:30:45`, with optional fractional seconds / timezone) and date-only values must remain valid.

javihern98:
## Summary

Makes Date/datetime validation strict and **consistent across the pandas and DuckDB backends**, and documents the accepted input formats.

**1. Reject partial/truncated times (all paths).** A time component, when present, must be a complete `HH:MM:SS` (fractional seconds and a timezone suffix stay valid); date-only stays valid. Previously the engine delegated to Python's lenient `datetime.fromisoformat`, silently accepting `2020-01-01T12` / `2020-01-01T12:30` and zero-filling. Now enforced on the pandas load (`check_date`) and cast (`check_max_date`) paths via a shared `normalize_datetime()`, and on both DuckDB paths (CSV loader + DataFrame register).

**2. Accurate error messages.** A `_build_date_error()` classifier reports the real fault — incomplete/invalid time vs. out-of-range month/day vs. bad format — instead of always saying "out of range for the month"; the year-range message now says `1800`. On DuckDB, a bad separator (`2020-01-01X12:30:45`, previously silently truncated to `2020-01-01`) and out-of-range times (`2020-01-01T25:00:00`, previously `Date unknown is out of range for the month`) are now rejected with a clear message.

**3. Cross-backend format parity.** All three input paths (pandas, DuckDB CSV, DuckDB DataFrame) now accept the same set: `YYYY-MM-DD`, or `YYYY-MM-DD` + (`T` | space) + complete `HH:MM:SS` + optional `.ffffff` + optional timezone (`±HH:MM` | `Z`). The DuckDB CSV loader previously rejected the `T` separator and any timezone; both DuckDB builders now share one `VALID_DATE_REGEX`. Timezones are normalized identically on both backends — the offset is dropped, keeping wall-clock (naive) — so the same input yields the same output. A trailing `Z` is accepted on all supported Python versions (3.9+).

**4. Docs.** `docs/data_types.rst` now documents the `T` separator, the required complete `HH:MM:SS`, and the accepted `Z`/`±HH:MM` timezone suffix (offset discarded).

## Checklist

- [x] Tests pass (`pytest`) — full suite green on both backends (DuckDB: 5349 passed; pandas: 4777 passed)
- [x] Documentation updated (`docs/data_types.rst`)

## Impact / Risk

- **Behavior change:** partial/truncated times are rejected instead of silently zero-filled; malformed datetimes raise clearer, accurate errors; timezone offsets are now dropped (naive) on the pandas path too (previously preserved), making both backends consistent.
- Full datetimes (with `T`/space, fractional seconds, `±HH:MM`/`Z`) and date-only values remain valid. No CSV/JSON fixture had a `T`/tz value in a `Date` column, so no reference data changed.
- No public API/signature changes; no SDMX compatibility concerns.
- Release note: stricter and cross-backend-consistent Date/datetime parsing; clearer validation messages; timezone offsets are normalized away.

## Notes

- Pre-existing, still out of scope (neither involves `T`/`Z`): pandas requires zero-padded `\d{2}` month/day while DuckDB accepts `\d{1,2}` (`2020-1-1`); DuckDB promotes a date-only value to `...T00:00:00` when it shares a column with datetimes (pandas keeps it date-only).

Fixes #879

### DIFF
--- a/docs/data_types.rst
+++ b/docs/data_types.rst
@@ -145,6 +145,11 @@ Date
       - ISO 8601 date: ``"2020-01-15"``,
         ``"2020-01-15 10:30:00"``,
         ``"2020-01-15T10:30:00"``.
+        A time component, when present, must be a
+        complete ``HH:MM:SS`` (``T`` or space
+        separator). An optional timezone suffix
+        (``Z`` or ``±HH:MM``) is accepted; the offset
+        is discarded and the wall-clock time is kept.
         Nanosecond precision is truncated to
         microseconds. Year range: 1800–9999.
     * - **Input (DataFrame)**
@@ -157,6 +162,7 @@ Date
     * - **Output format**
       - ISO 8601 with ``T`` separator:
         ``"YYYY-MM-DD"`` or ``"YYYY-MM-DDThh:mm:ss"``.
+        Timezone offsets are not included in the output.
     * - **Output dtype**
       - ``string[pyarrow]``
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "vtlengine"
-version = "1.9.0"
+version = "1.9.1rc1"
 description = "Run and Validate VTL Scripts"
 license = "AGPL-3.0"
 readme = "README.md"
--- a/src/vtlengine/DataTypes/TimeHandling.py
+++ b/src/vtlengine/DataTypes/TimeHandling.py
@@ -598,12 +598,15 @@ def check_max_date(str_: Optional[str]) -> Optional[str]:
     if len(str_) == 9 and str_[7] == "-":
         str_ = str_[:-1] + "0" + str_[-1]
 
-    # Accept YYYY-MM-DD (len 10) or YYYY-MM-DD[T| ]HH:MM:SS[.fffffffff]
-    has_time = len(str_) > 10 and str_[10] in ("T", " ")
-    if has_time:
-        from vtlengine.DataTypes._time_checking import _truncate_nanoseconds
+    # Accept YYYY-MM-DD (len 10) or a full YYYY-MM-DD[T| ]HH:MM:SS[.ffffff][+HH:MM|Z].
+    # A time component, when present, must be a COMPLETE HH:MM:SS.
+    from vtlengine.DataTypes._time_checking import _has_time_component, normalize_datetime
 
-        return dt.fromisoformat(_truncate_nanoseconds(str_)).isoformat(sep=" ")
+    if _has_time_component(str_):
+        try:
+            return normalize_datetime(str_)
+        except ValueError:
+            raise RunTimeError("2-1-19-8", date=str_) from None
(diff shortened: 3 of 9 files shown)


mla2001 (approved):
Looks fine! 😊

PR #882, opened 2026-07-09
Maintain attributes inside joins

### LINKED ISSUE #854: Fix several join errors
### Initial Checks

### Summary

- ASTString set the drop (inside join) statement outside the join as [drop ...] clause.
- Inner_join keeps non viral attributes unlike the other join clauses.
- Filter (inside join) clause checks for duplicates measure even when its not the last regular aggr clause inside the join.

### vtlengine version

1.9.0rc7

### Python version

Any

### OS

Any

### LINKED ISSUE #881: Joins silently drop Attribute components; attributes should be maintained
### Initial Checks

- [x] I have searched existing issues for duplicates
- [x] I have read the documentation

### Summary

Join operators (`inner_join`, `left_join`, `full_join`, `cross_join`) over two or more datasets silently drop every `Attribute`-role component from the result. This happens both to attributes coming from the operands and to attributes explicitly materialized inside the join body via `calc`/`apply`. Measures created the same way are correctly kept, so this is specific to the attribute role.

Attributes should be maintained in the join result. Dropping them is data loss and is not what the user asked the join to produce.

### Scope

_No response_

### Testing

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Join operators (`inner_join`, `left_join`, `full_join`, `cross_join`) silently dropped every `Attribute`-role component from the result — both attributes coming from the operands and attributes explicitly created inside the join body via `calc`/`apply`. This was introduced in 1.9.0 (`5f6b3b04`, "Fix #854") to implement the VTL 2.2 non-propagation rule, but it is data loss and not the intended behavior.

This reverts only the attribute-drop: it was bolted onto `_strip_join_prefixes` via a `join_drops_attributes` flag, kept separate from that method's original job (stripping `ds#comp` prefixes). Removing it makes attributes behave like measures in a join — maintained when unambiguous, with shared duplicates still requiring `keep`/`rename`. Viral-attribute merging is untouched.

## Checklist

- [x] Tests pass (`pytest` — 4974 passed, 11 skipped)
- [ ] Documentation updated (if applicable) — no docs reference this behavior

## Impact / Risk

- **Breaking change (behavior):** join results now include attribute components again. Consumers relying on the 1.9.0 drop will see extra columns.
- **SDMX:** positive — attributes are preserved through joins, matching real-world mart use cases.
- **Release note:** "Attributes are now maintained through join operators (reverts the VTL 2.2 non-propagation drop added in 1.9.0)."

## Notes

- Consciously diverges from the VTL 2.2 non-propagation rule; flag if strict conformance is ever required.
- Reverted the tests that encoded the drop: `test_GH_847_9` (Joins) and `AnaMart_AnaMart_1` (now expects `1-1-13-4`, the real `using` error, instead of `1-1-1-10` which only appeared because the attribute was dropped).
- 82 reference data files regenerated; each verified to add **only** attribute columns, with identifiers, measures, and row counts unchanged.

Fixes #881


javihern98 (approved):
Looks good, thanks! 😊

PR #883, opened 2026-07-09
Maintain attributes inside joins (1.9.X)

### LINKED ISSUE #854: Fix several join errors

albertohernandez1995:
## Summary

Join operators (`inner_join`, `left_join`, `full_join`, `cross_join`) silently dropped every `Attribute`-role component from the result — both attributes coming from the operands and attributes explicitly created inside the join body via `calc`/`apply`. This was introduced in 1.9.0 (`5f6b3b04`, "Fix #854") to implement the VTL 2.2 non-propagation rule, but it is data loss and not the intended behavior.

This reverts only the attribute-drop: it was bolted onto `_strip_join_prefixes` via a `join_drops_attributes` flag, kept separate from that method's original job (stripping `ds#comp` prefixes). Removing it makes attributes behave like measures in a join — maintained when unambiguous, with shared duplicates still requiring `keep`/`rename`. Viral-attribute merging is untouched.

## Checklist

- [x] Tests pass (`pytest` — 4974 passed, 11 skipped)
- [ ] Documentation updated (if applicable) — no docs reference this behavior


javihern98 (approved):
Looks good, thanks! 😊

PR #885, opened 2026-07-09
Update coverage to 7.15.0 (supersedes #876)

javihern98:
## Summary

Bumps `coverage` 7.14.3 → 7.15.0 (dev group floor raised to `>=7.15.0,<8.0`), with the lock regenerated by Poetry itself. This supersedes Dependabot PR #876, which cannot be merged as-is: its machine-generated `poetry.lock` is invalid.

**Why #876 fails CI** — the `testing (windows-latest, 3.14)` job dies with `xdist: maximum crashed workers reached`. The crash has nothing to do with coverage, pytest-cov, or pytest-xdist:

- Dependabot's lock regeneration dropped the marker-split `numpy 2.4.6` entry and stripped the `python_version == "3.10"` marker from the `numpy 2.2.6` entry, collapsing numpy to 2.2.6 for **all** Python versions — violating the `numpy>=2.4.4,<2.5; python_version >= '3.11'` constraint introduced in #871. This is the known Dependabot bug [dependabot-core#5862](https://github.com/dependabot/dependabot-core/issues/5862) (Poetry multiple-constraint dependencies broken by unrelated updates: only the first lock entry is kept).
- numpy 2.2.6 predates Python 3.14 and ships no cp314 wheels. On windows/3.14 the resulting binary triggers `Windows fatal exception: access violation` inside `numpy._core._multiarray_umath` import during collection, killing all 20 xdist workers instantly (`node down: Not properly terminated`). On ubuntu/macos 3.14 the 2.2.6 binary happens to import, so only the Windows job failed.
- Isolation: the passing cr-881 run (2026-07-09) used the same CPython 3.14.6 on windows with coverage 7.14.3 + numpy 2.4.6; #876's failing run had coverage 7.15.0 + numpy 2.2.6, and the fault address is in the numpy import.

In this PR the lock diff touches **only** coverage; both numpy entries survive with their markers intact.

## Checklist

- [x] Tests pass (`pytest`) — full suite in CI; `tests/API` smoke run locally with `-n auto --cov=vtlengine` (239 passed)

## Impact / Risk

- **Breaking changes?** None. Dev-group dependency bump; no runtime dependency or API changes.
- **Data/SDMX compatibility concerns?** None.
- **Notes for release/changelog?** Routine dependency maintenance.

## Notes

- #876 should be closed without merging (Dependabot should also auto-close it once this lands, as coverage will no longer be outdated).
- The green windows-3.14 job on this PR is the decisive evidence that coverage 7.15.0 is compatible with the pytest-xdist/pytest-cov stack.
- The #871 constraint alone does **not** protect us from dependabot-core#5862 — Dependabot writes a lock that violates `pyproject.toml` markers, and `poetry install` trusts the lock. Any future Dependabot PR here needs its lock diff checked for the numpy entries (or a CI guard added; happy to file a follow-up Task for that).
- No associated issue, so no closing keyword.


javihern98:
No longer need for this PR, we will update main with a merge from 1.9.X, which already includes this

PR #862, opened 2026-06-22
Aligned Viral Atributes tests with the main test suite.

albertohernandez1995:
## Summary

Match the viral attribute test suite to the standard vtlengine pattern:

## Checklist

- [x] Code quality checks pass (`ruff format`, `ruff check`, `mypy`)
- [x] Tests pass (`pytest`)
- [ ] Documentation updated (if applicable) — N/A (tests only)

---

Closes #810

PR #886, opened 2026-07-10
Merge 1.9.1 into main (excluding pandas backend changes)

javihern98:
## Summary

Forward-ports the v1.9.1 release content (`v1.9.0..v1.9.1` on `1.9.X`) to main, excluding everything related to the pandas execution backend, which main no longer has. Scope per the release notes:

| 1.9.X commit | PR | Status here |
| --- | --- | --- |
| cd3afc33 | #878 — viral attribute propagation and rule execution across operators (#877) | **Ported** (DuckDB + semantic layers only) |
| 8d429dd9 | #880 — strict, cross-backend-consistent Date/datetime validation (#879) | **Ported** |
| 988017df | #883 — maintain attributes inside joins | Already on main as #882 (byte-identical src change) — nothing to do |
| 818fa684 | #884 — bump version to 1.9.1 | Not ported (main stays `2.0.0rc3`) |

References #878 / #880 (issues #877 and #879 are already closed by the 1.9.X releases).

### What lands from #878

- New semantic errors **1-3-3-5** (`aggregate sum/avg` viral rule requires a numeric attribute) and **1-3-3-6** (every viral attribute reaching a result requires a `define viral propagation` rule), enforced once per statement result so they fire for both `run()` and `semantic_analysis()`.
- `DAGAnalyzer.sort_ast` keeps `ViralPropagationDef` nodes in multi-statement scripts.
- Viral attributes propagate through `unpivot`, `period_indicator`, `check_datapoint`, `check_hierarchy` and `hierarchy`: structure inference in the Operators' `validate()` methods, execution in the SQL transpiler (`vp_dataset_wide_sql`, hierarchy viral CTE chain, validation re-attach joins).
- Hierarchy operands are always deep-copied (`validate_hr_dataset` strips attributes in place and corrupted the shared operand's structure for later statements) — `GL_463`/`GL_cs_22` reference outputs updated, identically to 1.9.X.

**Excluded as pandas-backend changes:** the `evaluate()` hunks (`Operators/__init__.py`, `Numeric.py`, `String.py`, and the data parts of `Clause/Time/Validation/HROperators`), the pandas resolution methods (`resolve_single`/`apply_row_preserving`/`resolve_group`) that main already deleted, the `node_children`/`viral_values` threading that only feeds pandas execution, and #878's CI workflow fix (already on main).

### What lands from #880

- Shared strict `normalize_datetime()` (`DataTypes/_time_checking.py`): partial times (`2020-01-01T12:30`) are rejected instead of zero-filled; malformed times are reported as time errors instead of *"out of range for the month"*; `Z`/offsets are accepted on all supported Pythons and normalized to naive datetimes.
- One canonical `VALID_DATE_REGEX` shared by the DuckDB CSV and DataFrame loaders (T or space separator, complete in-range `HH:MM:SS`, optional fraction/timezone).
- `docs/data_types.rst` documents the accepted Date input formats.

### Main-only adaptation (candidate to back-port to 1.9.X)

The ported DataFrame Date guard is applied **only to string-like source columns** (VARCHAR/ENUM). As ported verbatim it also ran on typed temporal columns, rejecting tz-aware `datetime64` DataFrames that loaded fine before (DuckDB renders `TIMESTAMPTZ` with a `+01` offset the regex doesn't match) and paying a per-row regex on already-typed data. Source types come from a `DESCRIBE` of the registered view / the zero-row parquet scan. Regression tests included.

### Tests

- #878's new coverage lands as `tests/ViralAttributes/test_viral_rule_execution.py`, adapted to the DuckDB-only API (no `use_duckdb` parametrization, no pandas-registry unit tests). The inline test files that #862 removed are **not** reinstated.
- The file-based viral suite now declares propagation rules on its vtl cases (1-3-3-6 fallout): identity-enumerated rules for row-preserving cases, `aggregate max` for combining ones; `2-3`/`11-1`/`11-2`/`11-3` references updated to rule-execution semantics; `1-3`/`3-2`/`4-9` converted to 1-3-3-6 error assertions, mirroring the 1.9.X treatment.
- #880's tests merged into `tests/DateTime/test_datetime.py` and `tests/duckdb_transpiler/test_efficient_io.py`.

## Checklist

- [x] Tests pass (`pytest` — 5037 passed, 11 skipped)
- [x] Documentation updated (`docs/data_types.rst`; error docs are generated at build time)

## Impact / Risk

- **Breaking (as released in 1.9.1):** scripts with viral attributes but no `define viral propagation` rule now fail semantic analysis with 1-3-3-6 (previously silent NULL combines / passthrough). `aggregate sum/avg` rules over non-numeric viral attributes now raise 1-3-3-5 at semantic time.
- **Breaking (as released in 1.9.1):** stricter Date parsing on load — partial times, bad separators and out-of-range times are rejected on the CSV, DataFrame and pandas validation paths; `T` separator and timezones are accepted uniformly.
- Hierarchy no longer mutates its operand in place; scripts that reused an operand after a hierarchy/check_hierarchy statement get correct (previously corrupted) structures — `GL_463`/`GL_cs_22` reference outputs changed accordingly.
- Verified end-to-end with the reproduction scripts from #877 and #879.

## Notes

An adversarial review of this port surfaced six quirks that are **byte-identical to released 1.9.1** (kept here for parity). Each is now tracked as its own Bug issue with a minimal reproducible example:

1. #887 — viral propagation definition placed after the transformations raises 1-3-3-6 in scripts without inter-statement dependencies (`create_dag` skips `sort_ast` when the DAG has no edges).
2. #888 — `check_datapoint` executes aggregate viral rules per rule arm instead of over the whole validation result.
3. #889 — `check_datapoint` over a ruleset defined on a viral attribute raises a misleading 1-1-1-10 "Component not found".
4. #890 — `unpivot` into a name colliding with a viral attribute raises a bare `ValueError` instead of a `SemanticError`.
5. #891 — Date values with 1-2 fractional-second digits fail on Python 3.10 (validator regex accepts, `fromisoformat` rejects).
6. #892 — viral propagation SQL re-executes the hierarchy operand per leaf child and recomputes the unpivot dataset-wide window per measure (`optimization`).

Pre-existing main-side gaps observed while testing (not introduced here, not yet filed): the DataFrame path's `_detect_date_type_overrides` misses 1-digit-month datetimes (position-10 check) and the parquet loader has no Date→TIMESTAMP override detection, so datetimes that pass validation can still be cast-truncated to dates on those paths.


albertohernandez1995 (approved):
Good effort!

mla2001 (approved):

Issue #891, opened 2026-07-10
Date values with 1-2 fractional-second digits fail on Python 3.9/3.10 (validator regex accepts, fromisoformat cannot parse)


javihern98:
### Summary

The strict datetime validator (`DataTypes/_time_checking.py`) accepts fractional seconds of **any length** via `_STRICT_DATETIME_RE` (`(\.\d+)?`), then `normalize_datetime` delegates the actual parsing to `datetime.fromisoformat`. Before Python 3.11, `fromisoformat` only parses fractions of **exactly 3 or 6 digits**, so a value like `2020-01-01 12:30:45.12`:

- is accepted and normalized on Python ≥ 3.11,
- raises inside `fromisoformat` on Python 3.9 and 3.10 (surfacing as a date-format engine error),
- while the DuckDB CSV load path (regex-only validation) accepts it on every interpreter.

The same input therefore validates differently depending on the Python version and on the input path. Since `requires-python = ">=3.9"` is the support floor for clients installing via pip, both 3.9 and 3.10 installs are affected.

> [!NOTE]
> `requires-python = ">=3.9"` (pip-client floor) intentionally differs from the Poetry `python = ">=3.10,<4.0"` constraint, which only governs the development environment — that difference is by design, not part of this bug.
>
> A straightforward fix is to right-pad/truncate the fractional part to 6 digits before calling `fromisoformat`, keeping the regex as the single source of truth for what is accepted. Note that any fix must stay 3.9-compatible.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r <- DS_1[calc Me_2 := cast("2020-01-01 12:30:45.12", date)];
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Number", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

data_df = pd.DataFrame({"Id_1": [1], "Me_1": [10.0]})

run_result = run(script=script, data_structures=data_structures, datapoints={"DS_1": data_df})

print(run_result["DS_r"].data.to_string(index=False))
```

The cast provably routes through the strict validator — an incomplete time is rejected by it:

```python
# 'DS_r <- DS_1[calc Me_2 := cast("2020-01-01 12:30", date)];' raises:
[...]

PR #893, opened 2026-07-10
Align Python floor declarations to 3.10

javihern98:
## Summary

The project metadata declared `requires-python = ">=3.9"` while everything else already required 3.10: the Poetry constraint (`python = ">=3.10,<4.0"`), the CI test matrix (3.10–3.14) and the resolved lock file. Python 3.9 was therefore never actually supported — pip on 3.9 would accept the wheel and then fail against the locked dependencies. This aligns every declaration to **Python 3.10+**:

- `pyproject.toml`: `requires-python = ">=3.10"`, drop the `3.9` trove classifier, and remove the unreachable `python_version < '3.10'` dependency branches (pyarrow, numpy, pandas-stubs). The numpy 3.10 / 3.11+ marker split is untouched.
- `README.md` / `CONTRIBUTING.md`: state Python 3.10+.
- `[tool.ruff] target-version` moves to `py310`; the newly-activated B905 rule is satisfied by making the ten existing `zip()` call sites explicit with `strict=False` — exactly the previous implicit behavior, so the change stays behavior-neutral (tightening provably-safe sites to `strict=True` is left for a separate change).
- `[tool.cibuildwheel]` stops building `cp39-*` wheels: a cp39 wheel whose metadata says `Requires-Python >=3.10` cannot pass cibuildwheel's install-and-test step, so keeping it would break the wheels workflow.
- `poetry.lock`: content-hash regeneration only — the diff is a single line, with **no resolved version or marker changes** (Poetry already resolved for >=3.10).

The mismatch was noted in #891 (which remains open for the unrelated `fromisoformat` fractional-seconds bug).

## Checklist

- [x] Code quality checks pass (`ruff format`, `ruff check`, `mypy`) under the new `py310` target
- [x] Tests pass (`pytest` — 4974 passed, 11 skipped); `poetry check` passes and the lock diff is hash-only
- [x] Documentation updated (README / CONTRIBUTING)

## Impact / Risk

- Wheel metadata now advertises `Requires-Python: >=3.10`, so pip on Python 3.9 will refuse to install instead of failing later during dependency resolution — this is the honest behavior, not a support change.
- No dependency version changes; Dependabot-sensitive numpy marker split preserved verbatim.

## Notes

Release notes: packaging metadata now correctly states the minimum supported Python (3.10); 3.9 was already unsupported in practice.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 Thanks for your interest in contributing! This guide covers the basics to get you productive quickly.
 
 ## Getting Started
-- Use Python 3.9+.
+- Use Python 3.10+.
 - Install dependencies with poetry:
   ```bash
   poetry install
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ the [VTL Language 2.1](https://sdmx-twg.github.io/vtl/2.1/html/index.html).
 
 ### Requirements
 
-The VTL Engine requires Python 3.9 or higher.
+The VTL Engine requires Python 3.10 or higher.
 
 ### Install with pip
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -2399,4 +2399,4 @@ test = ["pytest", "pytest-cov"]
 [metadata]
 lock-version = "2.1"
 python-versions = ">=3.10,<4.0"
-content-hash = "16cc603ae77e4ed710171a747fb016623b6517f1951af5857757754952a06224"
+content-hash = "cf6b8ed25be03528affcac78d0eab090a16a3194539edaf7462be0481252f9d7"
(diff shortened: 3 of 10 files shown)


javihern98:
Closing: this was opened on a misreading. `requires-python = ">=3.9"` is the intentional support floor for clients installing the engine via pip (cp39 wheels are built and smoke-tested by cibuildwheel, and the `python_version < '3.10'` dependency branches serve exactly those installs). The Poetry `python = ">=3.10,<4.0"` constraint only governs the development environment, so there is no mismatch to fix — and this PR would have dropped real 3.9 client support (its `zip(..., strict=)` additions do not even exist on 3.9). No dependency or packaging changes needed.

PR #896, opened 2026-07-13
Wrong semantic error raised in viral propagation with single expresion inside DAG.

### LINKED ISSUE #887: Viral propagation definition placed after transformations raises 1-3-3-6 in scripts without inter-statement dependencies
### Summary

`DAGAnalyzer.create_dag` only calls `sort_ast` (which hoists `ViralPropagationDef` statements before the transformations) when the statement DAG has inter-statement edges. In a script whose statements only read input datasets — so the DAG has no edges — the statements keep their source order, the assignment is visited before `visit_ViralPropagationDef` registers the rule, and the per-statement rule-requirement check rejects the script with `1-3-3-6` even though the rule is declared.

Whether the same script is valid therefore depends on whether *unrelated* statements happen to create DAG edges:

- definition first → OK
- definition last, no dependencies between statements → **SemanticError 1-3-3-6**
- definition last, plus a statement that depends on another → OK (an edge exists, `sort_ast` runs and hoists the definition)

The definition-language statements (`define viral propagation`, rulesets, operators) should be registered independently of the statement ordering, e.g. by hoisting them in the no-edges branch of `create_dag` as well, or by registering `ViralPropagationDef` nodes before the statement loop in `Interpreter.visit_Start`.

### Reproducible Example

```python
from vtlengine import semantic_analysis

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

VP = 'define viral propagation VP (variable VAt_1) is when "A" then "A" end viral propagation;'

cases = {
    "definition first, no dependencies": f"{VP} DS_r <- DS_1;",
    "definition last, no dependencies": f"DS_r <- DS_1; {VP}",
    "definition last, with dependencies": f"DS_A <- DS_1; DS_r <- DS_A; {VP}",
}
for name, script in cases.items():
    try:
        semantic_analysis(script=script, data_structures=data_structures)
        print(f"{name:35s} -> OK")
    except Exception as e:
        print(f"{name:35s} -> {type(e).__name__}: {e.args[0]}")
```

### Expected vs Actual

**Actual:**

```
definition first, no dependencies   -> OK
definition last, no dependencies    -> SemanticError: Viral attribute VAt_1 has no viral propagation rule; declare a 'define viral propagation' rule for it. Please check transformation with output Dataset DS_r
definition last, with dependencies  -> OK
```

**Expected:** all three cases pass — the position of a `define viral propagation` statement in the script must not affect the result, exactly as it already does not when the DAG has edges.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

albertohernandez1995:
## Summary

`DAGAnalyzer.create_dag` now always hoists definition-language statements
(`define viral propagation`, rulesets, operators) ahead of the transformations.
Previously `sort_ast` only ran when the statement DAG had edges, so an edgeless
script kept its source order and a trailing `define viral propagation` was
registered after the assignment that produced the viral attribute — rejecting a
valid script with a spurious `1-3-3-6`. Whether the same script validated
depended on whether unrelated statements happened to create DAG edges.

Fixes #887

## Checklist

- [x] Tests pass (`pytest`)


src/vtlengine/AST/DAG/__init__.py
```diff
@@ -129,15 +129,7 @@ def create_dag(cls, ast: Start) -> "DAGAnalyzer":
...
-            if len(dag.edges) != 0:
-                dag.sort_ast(ast)
-            else:
-                ml_statements: list = [
-                    ml
-                    for ml in ast.children
-                    if not isinstance(ml, (HRuleset, DPRuleset, Operator, ViralPropagationDef))
-                ]
-                dag.check_overwriting(ml_statements)
+            dag.sort_ast(ast)
```
javihern98, line 132: This solution is wrong as the check_overwriting will ensure that an output dataset is not declared twice. Ensure that if this is deleted a test is added to check an error is raised in this case

javihern98 (changes requested):

src/vtlengine/AST/DAG/__init__.py
albertohernandez1995: Good catch. The overwriting check is still enforced after removing the edgeless branch: `create_dag` now always calls `sort_ast`, which runs `check_overwriting` on the sorted statements. I verified both DAG shapes still raise `1-2-2`:

- edgeless: `DS_r <- DS_1; DS_r <- DS_2;`
- with edges: `DS_1b <- DS_1; DS_r <- DS_1b; DS_r <- DS_2;`

I've added a regression test (`TestCreateDagOverwriting` in `tests/DAG/test_topological_sort.py`) covering both cases to lock this in.

javihern98: Ok, is check_overwriting used anywhere else? Should we remove that code?

albertohernandez1995: No, it is not dead code. check_overwriting is called in exactly one place — inside sort_ast — and sort_ast runs only from create_dag, which now calls it unconditionally on every analysis. So it's not dead code: it's the sole enforcement point for the 1-2-2 error, so we should keep it. Removing it would drop the duplicate-output check entirely.

javihern98: Ok, will finish the review then, thanks! :blush:

javihern98 (approved):
Looks good now, thanks! 😊

PR #897, opened 2026-07-13
Window check_datapoint viral aggregate over the whole validation result

### LINKED ISSUE #888: check_datapoint executes aggregate viral propagation rules per rule instead of over the whole validation result
### Summary

In the SQL generated for `check_datapoint`, an **aggregate** viral propagation rule is evaluated as `AGG(col) OVER ()` **inside each per-rule SELECT, after that rule's `WHERE` condition**, and the rule arms are then combined with `UNION ALL` (`_build_dp_rule_sql`). The aggregate therefore collapses over each rule's own failing/qualifying rows instead of over the whole validation result.

This is also internally inconsistent: the SQL generated for `check_hierarchy` windows the same kind of rule over the entire joined union of validation rows.

With a ruleset of a single rule the two interpretations coincide, which is why the existing tests do not catch it; with two or more rules the per-row values diverge (and `sum`/`avg` rules in `all`/`all_measures` modes diverge by construction).

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    define datapoint ruleset R (variable Me_1) is
        r1: Me_1 < 15 errorcode "e1" errorlevel 1;
        r2: Me_1 > 5 errorcode "e2" errorlevel 1
    end datapoint ruleset;
    DS_r <- check_datapoint(DS_1, R invalid);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

data_df = pd.DataFrame({"Id_1": [1, 2], "Me_1": [1.0, 20.0], "VAt_1": [300, 100]})

print(run_result["DS_r"].data.sort_values("ruleid").to_string(index=False))
```

### Expected vs Actual

Rule `r1` fails only on the datapoint carrying `VAt_1 = 100`, rule `r2` fails only on the one carrying `VAt_1 = 300`.

**Actual** — the aggregate collapses per rule arm:

| Id_1 | ruleid | Me_1 | errorcode | errorlevel | VAt_1 |
| ---- | ------ | ---- | --------- | ---------- | ----- |
| 2 | r1 | 20.0 | e1 | 1 | 100.0 |
| 1 | r2 | 1.0 | e2 | 1 | 300.0 |

**Expected** — the `aggregate max` rule collapses over the viral values contributing to the whole `check_datapoint` result, i.e. `max(300, 100) = 300` on every validation row (matching how `check_hierarchy` windows over its entire result):

| Id_1 | ruleid | Me_1 | errorcode | errorlevel | VAt_1 |
| ---- | ------ | ---- | --------- | ---------- | ----- |
| 2 | r1 | 20.0 | e1 | 1 | 300.0 |
| 1 | r2 | 1.0 | e2 | 1 | 300.0 |

> [!NOTE]
> A fix can move the viral window to a wrapper SELECT around the `UNION ALL` of the rule arms, so the rule executes once over the complete validation result.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

albertohernandez1995:
## Summary

In the SQL generated for `check_datapoint`, an **aggregate** viral propagation rule was evaluated as `AGG(col) OVER ()` *inside each per-rule SELECT, after that rule's `WHERE`*, and the rule arms were then combined with `UNION ALL`. The aggregate therefore collapsed over each rule arm's own rows instead of over the whole validation result — inconsistent with `check_hierarchy`, which windows the same kind of rule over its entire result. With a single-rule ruleset the two interpretations coincide, which is why existing tests did not catch it.

The fix carries the raw viral value per rule arm and moves the viral window to a wrapper SELECT around the `UNION ALL`, so each propagation rule executes once over the complete validation result. This reuses the existing `unpivot` / `check_hierarchy` pattern (`SELECT * EXCLUDE (viral…), <windowed viral> FROM (union)`).

For the issue's example, `aggregate max` now yields `max(300, 100) = 300` on every validation row.

## Checklist

- [x] Tests pass (`pytest` — 5038 passed, 11 skipped)

Fixes #888


javihern98 (approved):
Looks good, thanks! 😊

PR #898, opened 2026-07-13
Resolve datapoint rulesets whose signature variable is a viral attribute

### LINKED ISSUE #889: check_datapoint over a ruleset defined on a viral attribute raises misleading 1-1-1-10 "Component not found"
### Summary

Viral attributes are deliberately kept out of datapoint-rule handling: they are stripped from the internal `ruleset_dataset` before the rules are visited and re-attached to the validation result afterwards. As a side effect, a datapoint ruleset whose **signature variable is itself a viral attribute** can no longer resolve that variable, and the script fails with `1-1-1-10 "Component VAt_1 not found in Dataset DS_1"` — although the component clearly exists in the dataset.

> [!NOTE]
> If rules over viral attributes are intentionally unsupported (consistent with the "viral stays out of rule handling" model), the failure should be a dedicated semantic error stating exactly that, not a "Component not found" message that contradicts the input structure. If they should be supported, the rule visitation needs access to the un-stripped components while the result re-attachment keeps working as today.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    define viral propagation VP (variable VAt_1) is
        when "A" then "A"; when "B" then "B"
    end viral propagation;
    define datapoint ruleset R (variable VAt_1) is
        r1: VAt_1 = "A" errorcode "e1" errorlevel 1
    end datapoint ruleset;
    DS_r <- check_datapoint(DS_1, R);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

data_df = pd.DataFrame({"Id_1": [1], "Me_1": [10.0], "VAt_1": ["A"]})

print(run_result)
```

### Expected vs Actual

**Actual:**

```
vtlengine.Exceptions.SemanticError: ('Component VAt_1 not found in Dataset DS_1. Please check transformation with output Dataset DS_r', '1-1-1-10')
```

**Expected:** either the ruleset validates the viral attribute's values (returning the usual `check_datapoint` result structure), or a dedicated semantic error explains that datapoint rulesets cannot be defined on viral attributes. Reporting the component as *not found* is wrong in both readings — `VAt_1` exists in `DS_1`.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

albertohernandez1995:
## Summary

`check_datapoint` deliberately keeps viral attributes out of datapoint-rule
handling: they are stripped from the internal `ruleset_dataset` before the rules
are visited and re-attached to the validation result afterwards (issue #877).
As a side effect, a datapoint ruleset whose **signature variable is itself a
viral attribute** could no longer resolve that variable during rule visitation,
and the script failed with `1-1-1-10 "Component VAt_1 not found in Dataset DS_1"`
— even though the component clearly exists.

The fix strips only the viral attributes **not** named by the ruleset signature,
keeping the signature ones available for rule resolution. Non-signature viral
attributes stay out of rule handling as before, and re-attachment is unchanged
(it already uses the full operand, not the stripped dataset).

The result now validates the viral attribute's values and re-attaches it as a
`Viral Attribute`, e.g. for `check_datapoint(DS_1, R all)`:

| Id_1 | ruleid | bool_var | errorcode | errorlevel | VAt_1 | VAt_2 |
| ---- | ------ | -------- | --------- | ---------- | ----- | ----- |
| 1    | r1     | True     |           |            | A     | X     |
| 2    | r1     | False    | e1        | 1          | B     | Y     |

## Checklist

- [x] Tests pass (`pytest` — 5038 passed, 11 skipped)

## Impact / Risk

- **Breaking changes?** None. Rulesets over non-viral components behave exactly
  as before; this only widens what resolves.
- **Data/SDMX compatibility concerns?** None.
- **Notes for release/changelog?** `check_datapoint` now supports datapoint
  rulesets defined on a viral attribute.

## Notes

Regression test added under `tests/ViralAttributes` (code `13-1`) following the
suite's data-file pattern: it covers a viral attribute in the signature plus a
non-signature viral attribute, and fails on `1-1-1-10` without the fix.

Fixes #889


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):
Reapproved after merge

PR #899, opened 2026-07-13
Fix #890: raise SemanticError on unpivot name collision with a viral attribute

### LINKED ISSUE #890: unpivot into a component name colliding with a viral attribute raises a bare ValueError instead of a SemanticError
### Summary

`Unpivot.validate` re-attaches the operand's viral attributes to the result via `Dataset.add_component`. When the unpivot's new identifier or measure name collides with a viral attribute name, `add_component` raises a bare `ValueError("Component with name VAt_1 already exists")` that escapes the VTL error model entirely: no error code, no output-dataset context, and it is not a `SemanticError`/`VTLEngineException` subclass, so callers that catch engine errors will not catch it.

Every comparable structural clash in the engine surfaces as a `SemanticError` (e.g. `1-1-6-2` when the unpivot identifier already exists in the dataset); this collision should do the same.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r <- DS_1[unpivot Id_2, VAt_1];
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Me_2", "type": "Number", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

data_df = pd.DataFrame({"Id_1": [1], "Me_1": [10.0], "Me_2": [20.0], "VAt_1": ["A"]})

print(run_result)
```

### Expected vs Actual

**Actual:**

```
  File ".../vtlengine/Operators/Clause.py", line 238, in validate
    result_dataset.add_component(copy(viral_comp))
  File ".../vtlengine/Model/__init__.py", line 338, in add_component
    raise ValueError(f"Component with name {component.name} already exists")
ValueError: Component with name VAt_1 already exists
```

**Expected:** a `SemanticError` with a proper error code (like the existing `1-1-6-2` raised when the unpivot identifier collides with an existing component), naming the operator, the colliding component and the output dataset.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

albertohernandez1995:
## Summary

`Unpivot.validate` validated the new identifier name but never the new measure name.
When the measure collided with a component retained in the result (a viral attribute,
a kept identifier, or the new identifier), `Dataset.add_component` raised a bare
`ValueError` that escaped the VTL error model — no code, no dataset context, not a
`SemanticError`.

This adds a collision check that raises the new `SemanticError` `1-1-6-14`, naming the
operator, the colliding component, and the output dataset. The measure may still reuse a
collapsed (removed) measure's name (`unpivot Id_2, Me_1`), which stays legal. Also set
`op = UNPIVOT` on the `Unpivot` class so the operator is named in this and the existing
`1-1-6-2` message (previously "At op None").

## Checklist

- [x] Tests pass (`pytest`)
- [ ] Documentation updated (if applicable) — `error_messages.rst` is build-time generated; new code appears automatically

## Impact / Risk

- Breaking changes? A script that previously crashed with a bare `ValueError` now raises a catchable `SemanticError` — strictly an improvement; valid scripts are unaffected.
- Notes for release/changelog? New error code `1-1-6-14`.

## Notes

New suite-pattern semantic test `CC_59` covers the collision. Full regression run: 1108 passed, 4 skipped.

Fixes #890

### DIFF
--- a/src/vtlengine/Exceptions/messages.py
+++ b/src/vtlengine/Exceptions/messages.py
@@ -438,6 +438,12 @@ centralised_messages = {
         "message": "At op {op}: Not allowed to overwrite an Identifier: {comp_name}",
         "description": "Raised when an operation attempts to overwrite an existing Identifier.",
     },
+    "1-1-6-14": {
+        "message": "At op {op}: The component {name} in Dataset {dataset} could not be "
+        "included in the {op} op.",
+        "description": "Raised when a component cannot be included in the specified operation "
+        "because its name collides with a component already present in the result.",
+    },
     # Comparison errors
     "1-1-7-1": {
         "message": "At op {op}: Value in {left_name} of type {left_type} is not comparable "
--- a/src/vtlengine/Operators/Clause.py
+++ b/src/vtlengine/Operators/Clause.py
@@ -3,7 +3,7 @@ from copy import copy
 from typing import List, Type, Union
 
 from vtlengine.AST import RenameNode
-from vtlengine.AST.Grammar.tokens import AGGREGATE, CALC, DROP, KEEP, RENAME, SUBSPACE
+from vtlengine.AST.Grammar.tokens import AGGREGATE, CALC, DROP, KEEP, RENAME, SUBSPACE, UNPIVOT
 from vtlengine.DataTypes import (
     Boolean,
     ScalarType,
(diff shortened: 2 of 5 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #900, opened 2026-07-13
Fix #891: normalize fractional seconds to microseconds before fromisoformat

### LINKED ISSUE #891: Date values with 1-2 fractional-second digits fail on Python 3.9/3.10 (validator regex accepts, fromisoformat cannot parse)

albertohernandez1995:
## Summary

The strict datetime validator (`DataTypes/_time_checking.py`) accepts fractional
seconds of any length via `_STRICT_DATETIME_RE`, then delegates parsing to
`datetime.fromisoformat`. Before Python 3.11, `fromisoformat` only parses
fractions of exactly 3 or 6 digits, so a value like `2020-01-01 12:30:45.12`
validated on 3.11+ but raised a date-format error on 3.9/3.10 — while the
regex-only DuckDB load path accepted it on every interpreter.

This replaces `_truncate_nanoseconds` (which only touched fractions longer than
6 digits) with `_normalize_fractional_seconds`, which pads short fractions and
truncates long ones to exactly 6 digits before `fromisoformat`. The regex stays
the single source of truth for what is accepted, and parsing is now identical on
every supported interpreter (3.9+) and input path.

## Checklist

- [x] Tests pass (`pytest`) — 5045 passed, 11 skipped

## Impact / Risk

- Breaking changes? No. Short fractions are now padded (e.g. `.12` → `.120000`);
  values already handled (6+ digits, no fraction, `Z`/`+HH:MM` suffixes) are
  unchanged.
- Data/SDMX compatibility concerns? None — this aligns the strict validator with
  the regex-only path, removing a Python-version-dependent discrepancy.
- Notes for release/changelog? Fixes date values with 1–2 (and 4–5) fractional-
  second digits failing on Python 3.9/3.10.

## Notes

- Added regression tests in `tests/DateTime/test_datetime.py` for 1/2/4/5-digit
  fractions, plus `Z` and `+02:00` variants, covering both `check_date` and
  `check_max_date`.
- Only Python 3.12 is available in the dev env, so the 3.9/3.10 failure can't be
  triggered directly here; the fix is version-agnostic since `fromisoformat`
  only ever receives a 6-digit (or no) fraction after normalization.

Fixes #891

### DIFF
--- a/src/vtlengine/DataTypes/_time_checking.py
+++ b/src/vtlengine/DataTypes/_time_checking.py
@@ -14,19 +14,26 @@ def _has_time_component(value: str) -> bool:
     return len(value) > 10 and value[10] in ("T", " ")
 
 
-def _truncate_nanoseconds(value: str) -> str:
-    """Truncate sub-second precision beyond 6 digits (microseconds) for Python compatibility."""
+def _normalize_fractional_seconds(value: str) -> str:
+    """Normalize sub-second precision to exactly 6 digits (microseconds).
+
+    ``_STRICT_DATETIME_RE`` accepts a fractional part of any length, but
+    ``datetime.fromisoformat`` before Python 3.11 only parses fractions of
+    exactly 3 or 6 digits. Padding short fractions and truncating long ones to 6
+    digits keeps the regex the single source of truth for what is accepted and
+    yields identical parsing on every supported interpreter (3.9+).
+    """
     dot_idx = value.find(".")
     if dot_idx == -1:
         return value
-    # Keep at most 6 decimal digits after the dot
+    # Span the run of digits forming the fractional part (a timezone suffix,
+    # if any, follows it and must be preserved).
     frac_end = dot_idx + 1
     while frac_end < len(value) and value[frac_end].isdigit():
         frac_end += 1
     frac_digits = value[dot_idx + 1 : frac_end]
-    if len(frac_digits) > 6:
-        return value[:dot_idx] + "." + frac_digits[:6] + value[frac_end:]
-    return value
+    normalized_frac = (frac_digits + "000000")[:6]
+    return value[:dot_idx] + "." + normalized_frac + value[frac_end:]
 
 
 # A full date followed by a COMPLETE HH:MM:SS time. Hours, minutes AND seconds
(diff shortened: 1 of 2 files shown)


javihern98 (approved):
Looks good, thanks! 😊

Issue #626, opened 2026-03-24
Ensure duckdb queries are thread-safe


javihern98:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

`Aggregation._agg_func()` and `Analytic.analyticfunc()` use `duckdb.query(query).to_df()`, which operates on DuckDB's **default in-process connection**. This shared default connection is **not thread-safe** — when two threads execute aggregation or analytic operations concurrently within the same process, they interfere with each other's pending query results.

The fix is to replace `duckdb.query()` calls with explicit per-call `duckdb.connect(database=":memory:")` connections (the same pattern `Eval._execute_query()` already uses successfully).

> [!NOTE]
> `Eval._execute_query()` in `General.py` already creates isolated `:memory:` connections per call, so External Routines are not affected by this issue.

**Affected files:**
- `src/vtlengine/Operators/Aggregation.py` — lines 201 and 230
- `src/vtlengine/Operators/Analytic.py` — line 301

### Reproducible Example

```python
import pandas as pd
from concurrent.futures import ThreadPoolExecutor, as_completed

from vtlengine import run

script = """
    DS_r <- sum(DS_1 group by Id_1);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Id_2", "type": "String", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Number", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

data_df = pd.DataFrame(
    {
        "Id_1": [1, 1, 2, 2, 3],
        "Id_2": ["A", "B", "A", "B", "A"],
        "Me_1": [10.0, 20.0, 30.0, 40.0, 50.0],
    }
)

datapoints = {"DS_1": data_df}


def run_vtl(thread_id):
    result = run(script=script, data_structures=data_structures, datapoints=datapoints)
    return thread_id, result


with ThreadPoolExecutor(max_workers=2) as executor:
    futures = [executor.submit(run_vtl, i) for i in range(2)]
    for future in as_completed(futures):
        tid, result = future.result()
        print(f"Thread {tid}: {result}")
```

[...]

PR #902, opened 2026-07-13
Fix #901: serialize C++ parser access to prevent parse-tree use-after-free

### LINKED ISSUE #901: create_ast is not thread-safe: C++ parser keeps the last parse tree in process-global state
### Initial Checks

### Summary

The compiled C++ parser (`src/vtlengine/AST/Grammar/_cpp_parser/bindings.cpp`) stores the ANTLR parser and parse tree of the **last** `parse()` call in a single process-global buffer (`static ParserState g_state;`). The `ParseNode`/`TerminalNode` objects it returns to Python hold **raw pointers** into that buffer (the binding comment even notes: *"raw pointer - kept alive by g_state"*), and `get_syntax_error()` / `get_comments()` read from the same global.

`create_ast()` (in `API/__init__.py`) calls `parse()` and then lazily walks the returned tree during `visitStart()`. Because the GIL is released between Python bytecodes, a **second thread calling `parse()` frees the first thread's tree mid-traversal** — a use-after-free. The result is non-deterministic corruption: a malformed AST (a persistent assignment silently lost → `KeyError: 'DS_r'` downstream), a `NotImplementedError` from the AST constructor walking a clobbered node, wrong results, or a hard segfault (core dump).

> [!IMPORTANT]
> This is a **pre-existing, change-independent** flake: it has intermittently failed CI on unrelated PRs (#890, #891, #896 — e.g. `tests/Concurrency/test_thread_safety.py::TestConcurrentAggregation::test_concurrent_sum` failing with `KeyError: 'DS_r'`). It is distinct from #626, which fixed thread-safety of the **DuckDB** query layer; this one is in the **parser** layer.

### Reproducible Example

Deterministic, single-threaded demonstration of the shared global state (a second `parse()` invalidates the first tree that Python is still holding):

```python
from vtlengine.AST.Grammar._cpp_parser import vtl_cpp_parser

first = vtl_cpp_parser.parse("DS_result <- DS_1;")
before = first.text
vtl_cpp_parser.parse("DS_other <- DS_2 + DS_3 + DS_4;")  # overwrites the global buffer
after = first.text

print("first tree BEFORE second parse:", before)
print("first tree AFTER  second parse:", after)
print("same object, content changed:", before != after)
```

Real-world impact — concurrent `run()` calls (timing-dependent; a tiny GIL switch interval makes it fail reliably):

```python
import sys
from concurrent.futures import ThreadPoolExecutor, as_completed

import pandas as pd

from vtlengine import run

sys.setswitchinterval(1e-6)  # widen the race window; the bug is otherwise rare

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}
datapoints = {"DS_1": pd.DataFrame({"Id_1": [1, 2, 3], "Me_1": [10.0, 20.0, 30.0]})}
scripts = ["DS_r <- DS_1 * 10;", "DS_r <- DS_1 + 1;", "DS_r <- DS_1[filter Me_1 > 10];"]


def task(i):
    r = run(script=scripts[i % 3], data_structures=data_structures, datapoints=datapoints)
    return r["DS_r"]  # raises KeyError: 'DS_r' when the parse tree was corrupted


for _ in range(60):
    with ThreadPoolExecutor(max_workers=6) as ex:
        futures = [ex.submit(task, i) for i in range(6)]
        for f in as_completed(futures):
            f.result()
print("completed without error")
```

### Expected vs Actual

**Deterministic example — Actual:**

```
first tree BEFORE second parse: DS_result<-DS_1;<EOF>
first tree AFTER  second parse: DS_other<-DS_2+DS_3+DS_4;<EOF>
same object, content changed: True
```

The `first` node's content mutated because its raw pointer now reads the second parse's tree.

**Threaded example — Actual:** intermittently raises `KeyError: 'DS_r'`, `NotImplementedError` (from the AST constructor), returns wrong results, or segfaults (core dump).

**Expected:** each `parse()` / `create_ast()` result is independent of concurrent parses; concurrent `run()` calls always produce correct, isolated results (and never corrupt memory).

### vtlengine version

2.0.0rc3 (main)

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

Root cause is the process-global `g_state` in `bindings.cpp` plus the lazy raw-pointer `LazyParseNode`/`LazyTerminalNode` handed to Python.

Two possible fixes:

- **Python-level (small, no recompile):** serialize `parse()` **and** the full parse-tree traversal under a re-entrant lock in `create_ast()` / `create_ast_with_comments()`. Correct because once `visitStart()` returns, the Python AST holds no C++ pointers. Parsing becomes serial across threads, but execution (DuckDB/pandas) still runs concurrently, so the throughput cost is small. *(This is the minimal fix applied to unblock CI.)*
- **C++-level (larger, needs wheel rebuild):** make `parse()` return a self-contained tree (own the ANTLR objects per call / eagerly materialize) so no raw pointers dangle and true parallel parsing is possible.

Related: #626 (DuckDB-layer thread-safety, already fixed).

albertohernandez1995:
## Summary

Concurrent `run()`/`create_ast()` calls intermittently failed with `KeyError: 'DS_r'`,
`NotImplementedError`, wrong results, or a segfault — and flaked CI on unrelated PRs
(#890, #891, #896). Root cause: the compiled parser stores the last parse tree in a
single process-global buffer (`g_state` in `bindings.cpp`) and returns raw pointers into
it, so a second thread's `parse()` frees a tree the first thread is still traversing in
`visitStart()` (use-after-free).

This serializes `parse()` and the full parse-tree traversal under a re-entrant
`parser_lock` in `create_ast()` and `create_ast_with_comments()`. Once `visitStart()`
returns, the Python AST holds no C++ pointers, so only parsing is serialized —
DuckDB/pandas execution still runs concurrently. Adds a regression test that reproduces
the corruption reliably (mixed scripts across threads under a tiny GIL switch interval).

## Checklist

- [x] Tests pass (`pytest`)
- [ ] Documentation updated (if applicable) — n/a

## Impact / Risk

- Breaking changes? No. Parsing is now serialized across threads, but it was never safe
  concurrently; execution stays parallel.
- Notes for release/changelog? Fixes a concurrency crash/corruption in the parser.

## Notes

Follow-up option (own issue/PR): make the C++ `parse()` return a self-contained tree so
parsing can also run in parallel. Related: #626 (DuckDB-layer thread-safety, already fixed).

Fixes #901

### DIFF
--- a/src/vtlengine/API/__init__.py
+++ b/src/vtlengine/API/__init__.py
@@ -24,7 +24,7 @@ from vtlengine.AST import Start
 from vtlengine.AST.ASTConstructor import ASTVisitor
 from vtlengine.AST.ASTString import ASTString
 from vtlengine.AST.DAG import DAGAnalyzer
-from vtlengine.AST.Grammar._cpp_parser import vtl_cpp_parser
+from vtlengine.AST.Grammar._cpp_parser import parser_lock, vtl_cpp_parser
 from vtlengine.duckdb_transpiler.Config.config import configured_connection
 from vtlengine.duckdb_transpiler.io import execute_queries, extract_datapoint_paths
 from vtlengine.duckdb_transpiler.Transpiler import SQLTranspiler
@@ -87,18 +87,22 @@ def create_ast(text: str) -> Start:
         Exception: When the vtl syntax expression is wrong.
     """
     text = text + "\n"
-    cst = vtl_cpp_parser.parse(text)
-    error = vtl_cpp_parser.get_syntax_error()
-    if error is not None:
-        raise VTLSyntaxError(
-            line=error["line"],
-            column=error["column"] + 1,
-            detail=error["message"],
-            source_line=error["source_line"],
-            underline_length=error["underline_length"],
-        )
-    visitor = ASTVisitor()
-    ast = visitor.visitStart(cst)
+    # The C++ parser holds the parse tree in process-global state and hands Python
+    # raw pointers into it, so parse() and the lazy tree traversal in visitStart()
+    # must run under parser_lock to stay safe across threads (see parser_lock docs).
+    with parser_lock:
+        cst = vtl_cpp_parser.parse(text)
+        error = vtl_cpp_parser.get_syntax_error()
+        if error is not None:
+            raise VTLSyntaxError(
+                line=error["line"],
+                column=error["column"] + 1,
+                detail=error["message"],
+                source_line=error["source_line"],
+                underline_length=error["underline_length"],
+            )
+        visitor = ASTVisitor()
+        ast = visitor.visitStart(cst)
     DAGAnalyzer.create_dag(ast)
     return ast
 
--- a/src/vtlengine/AST/ASTComment.py
+++ b/src/vtlengine/AST/ASTComment.py
@@ -2,7 +2,7 @@ from typing import Any, Dict
 
 from vtlengine.API import create_ast
 from vtlengine.AST import Comment, Start
-from vtlengine.AST.Grammar._cpp_parser import vtl_cpp_parser
+from vtlengine.AST.Grammar._cpp_parser import parser_lock, vtl_cpp_parser
 
 
 def generate_ast_comment(comment: Dict[str, Any]) -> Comment:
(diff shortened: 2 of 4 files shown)


javihern98 (approved):
Looks good, thanks for spotting it!😊

PR #908, opened 2026-07-16
Fix #906: execute viral propagation only at combination points (main, DuckDB-only)

### LINKED ISSUE #906: Viral propagation applied on non-combining operators (aggregate rule collapses viral column on abs)
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues) for duplicates

### Summary

Per the VTL 2.2 [*Attribute propagation rule*](https://sdmx-twg.github.io/vtl/2.2/), the default propagation algorithm must run **only** when input Data Points are combined into one output Data Point. When Data Points are **not** combined, "the values of the viral Attributes are simply copied from the input Data Point to the output Data Point".

The engine currently executes the propagation rule on essentially every dataset-producing operator (via shared row-preserving hooks in both backends) and requires every viral attribute in every statement result to declare a rule. This yields wrong results and spurious errors on operations that do **not** combine Data Points:

- An **aggregate** rule (`min`/`max`/`sum`/`avg`) collapses a viral column to a single dataset-wide value on a row-preserving op such as `abs(DS_1)` or `DS_1 + 5`.
- An **enumerated** rule can overwrite each value with the clause default on a row-preserving op.
- A viral attribute that is **never combined** still raises `1-3-3-6`, even though no rule is needed.

The rule should be executed (and required) only at the three combination points: an operation over two or more datasets; an aggregation/analytic group-by; and a hierarchy roll-up. Everywhere else — unary ops, dataset⊕scalar, `calc`/`keep`/`drop`/`rename`/`unpivot`, `cast`, time shifts, `nvl`, set operators, `check_datapoint`/`check_hierarchy` — viral attributes must be copied through unchanged.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}
datapoints = {
    "DS_1": pd.DataFrame(
        {
            "Id_1": [1, 1, 2],
            "Id_2": ["A", "B", "A"],
            "Me_1": [-1.0, -2.0, -3.0],
            "VAt_1": [10.0, None, 30.0],
        }
    )
}

script = """
    define viral propagation VP1 (variable VAt_1) is
        aggregate max
    end viral propagation;
    DS_r <- abs(DS_1);
"""

```

### Expected vs Actual

`abs(DS_1)` is row-preserving (it does not combine Data Points), so `VAt_1` should be copied. Instead the aggregate `max` rule is applied dataset-wide. Identical on both backends (`use_duckdb=False` and `use_duckdb=True`).

**Actual**

| Id_1 | Id_2 | Me_1 | VAt_1 |
| ---- | ---- | ---- | ----- |
| 1 | A | 1.0 | 30.0 |
| 1 | B | 2.0 | 30.0 |
| 2 | A | 3.0 | 30.0 |

**Expected**

| Id_1 | Id_2 | Me_1 | VAt_1 |
| ---- | ---- | ---- | ----- |
| 1 | A | 1.0 | 10.0 |
| 1 | B | 2.0 | (null) |
| 2 | A | 3.0 | 30.0 |

Additionally, removing the `define viral propagation` statement (so `VAt_1` has no rule and is never combined) currently raises:

```
SemanticError: Viral attribute VAt_1 has no viral propagation rule; declare a
'define viral propagation' rule for it. ... ('1-3-3-6')
```

Expected: the statement succeeds and `VAt_1` is copied unchanged, because `VAt_1` is never combined.

> [!NOTE]
> A rule is still required — and `1-3-3-6` must still be raised — when a viral attribute **is** combined at one of the three combination points (two-or-more datasets, aggregation/analytic group-by, hierarchy roll-up).

### vtlengine version

1.9.1 (branch `1.9.X`)

### Python version

3.12.3

### OS

Linux

### Additional Context

Follow-up to #878 (issue #877). The fix targets `1.9.X` first; a DuckDB-only adaptation for `main` will follow separately (main is DuckDB-only and its viral code has diverged).

javihern98:
## Summary

DuckDB-only adaptation for `main` of the 1.9.X change in #907 (issue #906). Viral attribute propagation was executed (and required) on essentially every dataset operator; per the VTL 2.2 *Attribute propagation rule* it must run **only when input data points are combined**, and otherwise **copy** viral attributes unchanged.

- **Combination points** (rule executed, rule required): an operation over two or more datasets (dataset–dataset binary, joins, `if`/`case` with ≥2 dataset branches); an aggregation/analytic group-by/partition; the `hierarchy` roll-up.
- **Everywhere else** (viral copied unchanged, no rule required): unary ops, dataset⊕scalar, `calc`/`keep`/`drop`/`rename`/`unpivot`, `cast`, time shifts, `nvl`, set operators, `check_datapoint`/`check_hierarchy`.

The `SemanticError 1-3-3-6` moves from a global per-statement check into the combining operators' `validate` methods (still fires during `semantic_analysis()`), flagging only the viral attributes actually combined. The row-preserving DuckDB SQL emissions become plain copies (`vp_dataset_wide_sql` removed); `nvl` copies its primary operand (COALESCE-filled).

`main` is DuckDB-only, so there is no pandas `evaluate`/`apply_row_preserving` path (this is the difference from the 1.9.X PR #907).

Fixes #906.

## Checklist

- [x] Tests pass (`pytest`) — full suite: 5056 passed, 11 skipped
- [x] Documentation updated (error `1-3-3-6` description; `error_messages.rst` regenerates from `messages.py`)

## Impact / Risk

- **Behavior change.** Row-preserving operators no longer execute the rule: an aggregate rule no longer collapses a viral column dataset-wide (e.g. `abs(DS_1)`), and an enumerated rule no longer remaps values — both now copy. A viral attribute never combined no longer requires a rule.
- **Supersedes #897.** #897 made `check_datapoint` aggregate the viral attribute over the whole validation result; as a row-preserving validation it now **copies** the viral value per datapoint. The `13-1` expected output is updated accordingly. Please confirm this is the intended direction.
- No public API changes. No SDMX compatibility concerns.

## Notes

- Mirror of #907 (1.9.X), adapted to main's DuckDB-only architecture. `nvl` is single-operand for propagation.
- The unrelated `main.py` working-tree change is excluded.

### DIFF
--- a/src/vtlengine/Exceptions/messages.py
+++ b/src/vtlengine/Exceptions/messages.py
@@ -1063,8 +1063,10 @@ centralised_messages = {
     "1-3-3-6": {
         "message": "Viral attribute {name} has no viral propagation rule; declare a "
         "'define viral propagation' rule for it.",
-        "description": "Raised when a viral attribute appears in a result without a "
-        "define viral propagation rule. Every viral attribute must declare one.",
+        "description": "Raised when a viral attribute is combined -- in an operation over "
+        "two or more datasets, an aggregation/analytic group-by, or a hierarchy roll-up -- "
+        "without a 'define viral propagation' rule. Viral attributes that are only copied "
+        "through row-preserving operators do not require a rule.",
     },
     # ---------- Interpreter ----------
     "1-3-5": {
--- a/src/vtlengine/Interpreter/__init__.py
+++ b/src/vtlengine/Interpreter/__init__.py
@@ -167,13 +167,6 @@ class InterpreterAnalyzer(ASTTemplate):
             if result is None:
                 continue
 
-            if isinstance(result, Dataset):
-                # Every viral attribute must declare a viral propagation rule (issue #877).
-                vp_registry = get_current_registry()
-                for viral_comp in result.get_viral_attributes():
-                    if vp_registry.rule_for(viral_comp) is None:
-                        raise SemanticError("1-3-3-6", name=viral_comp.name)
-
             vtlengine.Exceptions.dataset_output = None
             self.datasets[result.name] = copy(result)
             results[result.name] = result
--- a/src/vtlengine/Operators/Aggregation.py
+++ b/src/vtlengine/Operators/Aggregation.py
@@ -22,6 +22,7 @@ from vtlengine.DataTypes import (
 )
 from vtlengine.Exceptions import SemanticError
 from vtlengine.Model import Component, Dataset, Role
+from vtlengine.ViralPropagation import require_rules
 
 
 def extract_grouping_identifiers(
@@ -108,6 +109,9 @@ class Aggregation(Operator.Unary):
             )
             result_components["int_var"] = new_comp
 
+        # Aggregation combines the data points of each group, so the surviving viral
+        # attributes are combined and require a propagation rule (issue #906).
+        require_rules(operand.get_viral_attributes())
         # VDS is handled in visit_Aggregation
         return Dataset(name="result", components=result_components, data=None)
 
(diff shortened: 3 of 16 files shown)


javihern98:
Will be included in general merge from 1.9.X to main

PR #915, opened 2026-07-21
Fix #910: viral sum/avg type check at combination points, avg promotes to Number

### LINKED ISSUE #910: The avg and sum aggregation rules fail for non-real types
### Initial Checks

### Summary

The aggregation rule for this simple example is not working due to says that viral attribute is string type. The viral attribute is not string and works fine with `min` and `max`.

Only fails with `avg` y `sum`.

The error:

("Viral propagation rule for va uses aggregate 'sum', which requires a numeric attribute, but va has type String.", '1-3-3-5')

### Reproducible Example

```python
define viral propagation we(variable va) is
        aggregate avg
end viral propagation;



*       Module md

md.A :=
        DS_1_num + DS_2_num;
```

### Expected vs Actual

Expected: works fine

### vtlengine version

1.9.2rc1

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

javihern98:
## Summary

Issue #910: viral propagation rules with `aggregate avg`/`sum` raised `1-3-3-5` ("va has type String") even though every dataset actually combined declares `va` as Number, while `min`/`max` worked. Two defects fixed:

1. **Type check ran at definition time over all input datasets.** The sum/avg numeric-type check in `visit_ViralPropagationDef` scanned every input dataset — including datasets never combined, or never referenced by the script at all. The variable-signature check now lives in `require_rules`, which runs exactly at the VTL 2.2 combination points (binary ops over two datasets, joins, aggregation group-by, analytic partitions, hierarchy roll-ups, if/case over datasets), so the rule is validated only where it actually executes. Value-domain-signature rules keep the definition-time check, since a value domain's type is intrinsic to the rule target.

2. **`avg` on an Integer viral attribute crashed.** The result kept the Integer type, and the pandas backend raised `pyarrow.lib.ArrowInvalid: Float value 1.5 truncated converting to int64` on fractional means. The new `apply_viral_return_types` helper promotes a viral attribute combined under an `avg` rule to `Number` in the result structure, mirroring the `Avg` operator's `return_type` (`sum` keeps the input type, mirroring `Sum`). Promoted entries are replaced with copies so operators that share component objects with their inputs (analytic, case) never mutate the input dataset structure.

Both backends are covered: the semantic pass is shared, and the DuckDB data path already produced DOUBLE for avg — the declared schema now matches it.

## Checklist

- [x] Tests pass (`pytest` — 5405 passed, 27 skipped)
- [x] Documentation updated (if applicable) — no change needed, the 1-3-3-5 message text is unchanged

## Impact / Risk

- Behavior change: a sum/avg rule over a non-numeric viral attribute now raises `1-3-3-5` only when that attribute is actually combined; scripts that previously failed at definition time despite never executing the rule now run (the #910 report).
- Output structure change: an Integer viral attribute combined under an `avg` rule is now typed `Number` in the result (previously Integer, which crashed at runtime on fractional means).
- No public API or SDMX compatibility changes.

## Notes

- Forward-port to `main` will follow in a separate PR.
- When value domains land on components, `require_rules` will also cover value-domain rules at combination points via `rule_for`; their definition-time check stays meanwhile.

Fixes #910

### DIFF
--- a/src/vtlengine/ViralPropagation/__init__.py
+++ b/src/vtlengine/ViralPropagation/__init__.py
@@ -7,11 +7,15 @@ Registry for viral attribute propagation rules as defined by the VTL 2.2
 in :mod:`vtlengine.ViralPropagation.sql`.
 """
 
+from copy import copy
 from dataclasses import dataclass, field
 from functools import reduce
 from typing import Any, Dict, Iterable, List, Optional
 
 import pandas as pd
+
+from vtlengine.DataTypes import Integer, Number
+from vtlengine.Exceptions import SemanticError
 
 
 @dataclass
@@ -188,9 +192,6 @@ def require_rules(components: Iterable[Any]) -> None:
     non-numeric viral attribute: the check runs only where the rule actually executes,
     so a non-numeric attribute that is never combined raises nothing (issue #910).
     """
-    from vtlengine.DataTypes import Integer, Number  # local imports avoid an import cycle
-    from vtlengine.Exceptions import SemanticError
-
     registry = get_current_registry()
     for comp in components:
         rule = registry.rule_for(comp)
@@ -216,10 +217,6 @@ def apply_viral_return_types(components: Iterable[Any], result_components: Dict[
     Call this right after :func:`require_rules` at every combination point, with the
     same combined components and the result-components mapping under construction.
     """
-    from copy import copy
-
-    from vtlengine.DataTypes import Number  # local import avoids an import cycle
-
     registry = get_current_registry()
     for comp in components:
         rule = registry.rule_for(comp)
--- a/src/vtlengine/ViralPropagation/__init__.py
+++ b/src/vtlengine/ViralPropagation/__init__.py
@@ -223,7 +223,7 @@ def apply_viral_return_types(components: Iterable[Any], result_components: Dict[
         if rule is None or rule.aggregate_function != "avg":
             continue
         result_comp = result_components.get(comp.name)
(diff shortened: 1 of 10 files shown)


src/vtlengine/ViralPropagation/__init__.py
```diff
@@ -183,13 +187,47 @@ def require_rules(components: Iterable[Any]) -> None:
...
+    datasets and the input structure must never be mutated.
+
+    Call this right after :func:`require_rules` at every combination point, with the
+    same combined components and the result-components mapping under construction.
+    """
+    registry = get_current_registry()
+    for comp in components:
+        rule = registry.rule_for(comp)
+        if rule is None or rule.aggregate_function != "avg":
+            continue
```
mla2001, line 224: This condition is redundant, since the return type of `avg` is `number`, the second condition would catch it anyway.

mla2001 (changes requested):
Looks almost fine, only a small change is needed.

albertohernandez1995 (dismissed):

src/vtlengine/ViralPropagation/__init__.py
javihern98: This specific guard needs to be kept, if not, many tests pointing at guarding the execution of sum/avg viral propagation rules may misbehave given that they only apply to number. In any case, I simplified the next guard to just look if the component is None, instead of checking as well if it is a Number

mla2001 (approved):
Looks fine! 😊

PR #928, opened 2026-07-27
Fix #920: keep operand measure names when a clause wraps an operator

### LINKED ISSUE #920: Error in nested rename (Duckdb)
### Initial Checks

### Summary

Seems there are some problem if we put a rename after this udo call, bring this error.

```
Error: Exception: Execution error: Binder Error: Referenced column "OBS_VALUE" not found in FROM clause!
Candidate bindings: "TIME_PERIOD"

LINE 1: ... "freq", "geo", "nrg_bal", "siec", "unit" FROM (SELECT vtl_div("OBS_VALUE", 2) AS "Energy_Consumption", "TIME_P
```

If we split, remove replace of this line  and do it rename against NGR_BAL_S1 works fine.

dataset:  [NRG_BAL_S.json](https://github.com/user-attachments/files/30274609/NRG_BAL_S.json)

datapoints: [NRG_BAL_S_datapoints.csv](https://github.com/user-attachments/files/30274611/NRG_BAL_S_datapoints.csv)

### Reproducible Example

```python
define operator subtractYear(ds dataset)
	returns dataset is
		(
		timeshift(
		ds, -1
		) + timeshift(
		timeshift(
		ds, -1
		), -1
		)
		) / 2
end operator;



*       Module md1

md1.NRG_BAL_S1 :=
	subtractYear(NRG_BAL_S)
		[rename OBS_VALUE to Energy_Consumption];
```

### Expected vs Actual

Expected: works fine nested 

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

A clause applied directly to an operator result failed to bind, because the operator
underneath named its measure after the *assignment target* rather than keeping its own
name — while the clause in between still referred to the original name:

```
DS_r := ((DS_1 + DS_1) / 2)[rename OBS_VALUE to Energy];
-- Binder Error: Referenced column "OBS_VALUE" not found in FROM clause!
```

`visit_Validation` already stashed `current_assignment` for this reason; clause visitors
did not. Reported against a UDO with `timeshift`, but neither is required to trigger it —
`drop` after `round` fails the same way.

- Stash `current_assignment` while resolving a clause's source dataset, so nested operands
  name their measures naturally.
- Resolve mono-measure comparisons to `bool_var` in the structure visitor, mirroring
  `IN` / `NOT_IN` / `ISNULL`, and make the emitted SQL agree.
- Keep the declared role when `calc` introduces a new component, instead of defaulting to
  Measure and relying on the output structure to recover it.

Fixes #920

## Checklist

- [x] Tests pass (`pytest`) — 5406 passed (duckdb), 4834 passed (pandas)


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #939, opened 2026-07-29
Fix #924: give the recursive AST passes enough stack for long expression chains

### LINKED ISSUE #924: Recursion error (DuckDb)
### Initial Checks

### Summary

I received this error with this code. The execution fails even without data.  

`maximum recursion depth exceeded`

project: [demoteam_RP_SI_SF_2026-07-23-17-44_export.zip](https://github.com/user-attachments/files/30314212/demoteam_RP_SI_SF_2026-07-23-17-44_export.zip)

### Reproducible Example

```python
Attached as file.
```

### Expected vs Actual

Expected: Works as with panda.

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

A script that adds around 157 datasets together in single statements failed with
`maximum recursion depth exceeded` on DuckDB, before touching any data. An expression
chaining N operands parses into a left-nested `BinOp`, and each level costs several stack
frames — the SQL pass alone recurses `visit_BinOp → _build_ds_ds_binary → _get_dataset_sql
→ visit`, with structure resolution recursing on top of it. That reaches CPython's default
limit of 1000 at roughly 160 to 240 operands. The pandas engine uses fewer frames per
operand and stayed just under it, which is why only DuckDB failed.

- Add `recursion_headroom()`, which raises the limit to 5000 for the enclosed pass and
  restores it afterwards, and apply it via a decorator to `run`, `semantic_analysis` and
  `_run_with_duckdb`.

The reported script now completes at the default limit, in 262.8s against 711.1s on pandas.
The supported chain length goes from about 160 operands to about 800.

## Checklist

- [x] Tests pass (`pytest`) — 5409 passed (duckdb), 4837 passed (pandas)

## Impact / Risk

- No behaviour change for scripts that already worked; the limit is only raised, and only
  for the duration of a call. If a caller has already set a higher limit it is left alone,
  and the restore is skipped if something else raised it further meanwhile.
- The ceiling is deliberately 5000 rather than something larger. Python 3.10 is still
  supported and consumes C stack per Python call, so an excessive limit would trade a clean
  `RecursionError` for a possible interpreter crash. 5000 is roughly three times what the
  reported script needed.
- Wrapping the entry points rather than the transpiler alone is deliberate: fixing only the
  transpiler moved the failure into AST construction and then into the AST `deepcopy`.
- The pandas path gets the same headroom, since the AST layer is shared.

## Notes

At this limit the recursion depth is no longer the first thing to fail: an 800-operand chain
takes about 68 seconds because dataset structures are re-resolved at every level of the
chain, which is quadratic in its length. That cost is untouched here and deserves its own
issue if it ever bites.

Tests cover the helper's restore behaviour, `create_ast` on a 400-operand chain, and an
end-to-end `run` on 250. Calling `ASTVisitor` directly on the 400-operand chain still raises
`RecursionError`, which is what the wrapper prevents.

---
Closes #924


javihern98 (approved):
Looks good, thanks! 😊

PR #942, opened 2026-07-29
Fix #925: count over an empty Data Set yields one Data Point with zero

### LINKED ISSUE #925: Count do not work without data (DuckDb)
### Initial Checks

### Summary

Doing some test I notice that if I try to count a dataset which do not have columns, and count instead of return int_var = 0, return an empty data points.

### Reproducible Example

```python
count(DS1)
```

### Expected vs Actual

Expected: Return data point with 1 row int_var=0
Current: Do not return data points

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`count(DS_1)` over a Data Set with no Data Points returned nothing on DuckDB instead of a
single Data Point holding zero. The reference manual (p193) defines count as returning "the
number of the input Data Points", with result type `dataset {measure<integer> int_var}`, so
an empty input means one row with 0.

Probing the surrounding cases turned up a second, related divergence in the same area:

| ungrouped, empty input | pandas | duckdb before |
| --- | --- | --- |
| `count(DS_1)` | `int_var: 0` | no rows |
| `DS_1[aggr X := count(Me_1)]` | `X: 0` | `X: None` |

DuckDB was also inconsistent with itself: on all-null data the first form gave 0 and the
second gave NULL.

- Skip the `HAVING COUNT(*) > 0` guard for `count`. It correctly suppresses the row for
  `sum`, `avg` and `min` over an empty Data Set, which both engines agree return no Data
  Point, but count must still report zero.
- Apply `NULLIF(COUNT(...), 0)` at Component level only when the aggregation is grouped,
  mirroring what the Data Set level path already did.

## Checklist

- [x] Tests pass (`pytest`) — 5406 passed (duckdb), 4834 passed (pandas)

--
Closes #925


javihern98 (approved):
Looks good, thanks! 😊

Issue #940, opened 2026-07-29
Pandas engine raises TypeError on component-level string concatenation with a Boolean component


javihern98:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

In the pandas engine, string concatenation (`||`) with a Boolean component inside a clause raises `TypeError: 'bool' object can't be concatenated` (`src/vtlengine/Operators/__init__.py:255`, the series `map` path).

The same implicit Boolean→String coercion works at scalar level (`true || "_x"` → `"True_x"`) and at dataset level (`DS_1 || "_x"` → `"True_x"`/`"False_x"`), so the component level should behave the same.

> [!NOTE]
> Found while working on #923 (boolean→string alignment between the pandas and DuckDB engines). The DuckDB engine executes this expression; a parity test covering it runs DuckDB-only until this issue is fixed.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r := DS_1[calc Me_2 := Me_1 || "_x"];
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Boolean", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

datapoints = {"DS_1": pd.DataFrame({"Id_1": [1, 2], "Me_1": [True, False]})}

run_result = run(
    script=script,
    data_structures=data_structures,
    datapoints=datapoints,
    return_only_persistent=False,
)

print(run_result)
```

### Expected vs Actual

Expected: `Me_2` contains `"True_x"` / `"False_x"` (Boolean implicitly cast to String, as at dataset and scalar level).

Actual:

```
  File ".../src/vtlengine/Operators/__init__.py", line 255, in <lambda>
    return series.map(lambda x: cls.py_op(x, scalar), na_action="ignore").astype(
TypeError: 'bool' object can't be concatenated
```

### vtlengine version

1.9.2rc2

### Python version

3.12.3

### OS

Linux

javihern98:
Fixed in #952

PR #946, opened 2026-07-30
Fix #923: Python-style True/False for DuckDB boolean-to-string outputs

### LINKED ISSUE #923: Boolean output format change (Duckdb)
### Initial Checks

### Summary

In DuckDb engine Bollean outputs changes from all Lowercase to fist Uppercase.

```
true -> True
false -> False
```

### Reproducible Example

```python
Any boolean output.
```

### Expected vs Actual

Expected: Return same format

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

javihern98:
## Summary

Fixes #923

The DuckDB engine serialized booleans SQL-style (`true`/`false`) while the pandas engine uses Python-style (`True`/`False`), so switching engines changed the boolean output format. This PR standardizes every boolean→string surface of the DuckDB engine on the pandas format:

- **Dataset CSVs** (`output_folder`): `COPY TO` now writes `True`/`False` for BOOLEAN columns. The formatting is applied in `_build_dataset_fetch_select` behind a new `bool_as_python_str` flag, enabled only for CSV output — parquet output keeps native BOOLEAN and in-memory `fetchdf()` keeps bool dtype (regression tests included). `_scalars.csv` already matched (`True` in both engines).
- **Explicit cast**: `cast(x, string)` from Boolean now emits the existing `_bool_to_str` `CASE` expression (`'True'`/`'False'`) instead of `CAST(x AS VARCHAR)` (`'true'`), at scalar, component, and dataset level. A new `_is_boolean_expr` predicate infers Boolean for computed sources (`cast(Me_1 = true, string)`, logical/comparison operators, `isnull`, `between`, `match_characters`) that `_get_source_vtl_type` cannot resolve.
- **Implicit coercion at component/scalar level**: `||` concatenation with a Boolean operand now produces `'True_x'` instead of `'true_x'`, and the string operators (`upper`, `lcase`, `trim`, `ltrim`, `rtrim`, `length`, `substr`, `replace`, `instr`) on a Boolean no longer crash with `BinderException` — they coerce through `_bool_to_str` exactly like the already-working dataset level.

Also re-enables the deactivated `cast(DS_1, string)` case in `tests/duckdb_transpiler/test_run.py` (it asserted the wrong column name: mono-measure String casts rename the measure to `str_var` — verified identical in both engines).

## Checklist

- [x] Tests pass (`pytest`) — full suite under both `VTL_ENGINE_BACKEND=pandas` and `=duckdb`
- [x] Documentation updated (if applicable) — `data_types.rst` (Boolean CSV serialization + explicit-cast conversion detail), `duckdb_engine.rst` (behavioural note)

## Impact / Risk

- **Breaking changes?** DuckDB-engine users' dataset CSVs change from `true`/`false` to `True`/`False` for Boolean columns, and DuckDB boolean→string cast/coercion results change from `'true'` to `'True'`. This is the intended alignment with the pandas engine (the long-standing default format). Pandas-engine output is unchanged. Round-trips are unaffected: both engines' loaders accept either casing.
- **Data/SDMX compatibility concerns?** None — parquet keeps native BOOLEAN, in-memory results keep bool dtype in both engines.
- **Notes for release/changelog?** DuckDB engine boolean outputs now match the pandas engine (`True`/`False`).

## Notes

- The pandas engine crashes on component-level `Me_1 || "_x"` with a Boolean component (pre-existing, unrelated code path) — filed as #940; the new component-concat parity test runs DuckDB-only and references it.
- New tests: `cast(true/false/1=1, string)` interpreter params (both engines), transpiler SQL assertion, `test_GH_923` file-based regression test (cast + string ops on a Boolean measure, references generated from the pandas engine), written-CSV text/in-memory dtype/scalar-concat parity tests, and a parquet native-BOOLEAN guard.


albertohernandez1995 (approved):
Looks awesome! Thanks!

PR #947, opened 2026-07-30
Fix #933: symdiff no longer drops Data Points with null measures

### LINKED ISSUE #933: Error in symdiff
### Initial Checks

### Summary

Seems that symdiff is not working in pandas engine, seem in DuckDb is working fine. We have 2 examples:

1. Second param is subset of first so, should have all rows which  Id_1 <> 1.
2. We have a non coincident inputs due to they have different Id_1.

In pandas engine we have empty results in this two operations.

### Reproducible Example

```python
set_1 := DS_1[filter Id_1 = 1];
set_2 := DS_1[filter Id_1 = 2];

symdiff_ds := symdiff(DS_1, set_1);
symdiff_ds_2 := symdiff(set_1,set_2);
```

### Expected vs Actual

Expected: Return datapoints
Current: Do not return datapoints

### vtlengine version

v1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`symdiff` returned an empty result on the pandas engine whenever the operands had a null
measure value. `Symdiff.evaluate` used `dropna()` to discard the Data Points present in both
operands: it set every non-identifier column to `None` for those rows and then dropped any row
containing a null. That also removed the rows it was supposed to keep, and with a measure that
is null on every Data Point — as in the reported data — nothing survived at all.

The same method also decided which operand a Data Point came from by testing whether its
measures were null, which misreads a Data Point whose measure is genuinely null.

- Use `merge(..., indicator=True)` and filter on `_merge != "both"`, the approach `Setdiff`
  in the same module already takes, then take each remaining row's values from the side it
  came from.

DuckDB was already correct; only the pandas engine changes.

Closes #933

## Checklist

- [x] Tests pass (`pytest`) — 5407 passed (duckdb), 4835 passed (pandas)


javihern98 (approved):
Looks good, thanks! 😊

PR #948, opened 2026-07-30
Fix #934: infer the Date series period from calendar gaps, not day cpunts

### LINKED ISSUE #934: Different behavior in fill_time_series with dates (DuckDb)
### Initial Checks

### Summary

With this example we have different behavior in pandas and duckdb. First difference is the length, pandas we have much more rows, and the origin of this difference I think is this change of behavior, in pandas we fill a date will all the dates, and now only fill with months. 

### Reproducible Example

```python
fill_time_series_date_single_ds_dt := fill_time_series(DS_3_with_date, single);
fill_time_series_period_all_ds_dt := fill_time_series(DS_3_with_date, all);
```

### Expected vs Actual

Expected: Have same behavior 

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`fill_time_series` over a Date identifier filled every calendar day on the pandas engine
where DuckDB filled monthly, so the same script returned 244 rows against 10.
`find_min_frequency` chose the period with

```python
min_days = min((d for d in days_deltas if d > 0 and d % 365 != 0 and d % 366 != 0), default=None)
return "D" if min_days else "M" if min_months else "Y"
```

`min_days` is set whenever any gap is not an exact multiple of 365 or 366 days, which holds for
any monthly series, so the answer was almost always `D`. The reference manual (p147) steps the
reference time Identifier "one period at a time" using "the period of each time series", so the
DuckDB result was the correct one.

- Rewrite `find_min_frequency` to mirror the DuckDB ladder (`vtl_date_timeshift_period_ind`): a
  gap that is not a whole number of calendar months leaves only weeks or days, otherwise the
  coarsest period dividing every gap wins. It now takes the dates rather than the timedeltas,
  since calendar month arithmetic needs the endpoints.
- Build the fill grid by stepping from the lower limit (`_period_range`) instead of a pandas
  frequency alias. `_to_pandas_freq` mapped `S` and `Q` to month-start aliases but left `M` and
  `Y` as month and year *end*, which would have generated 01-31, 02-29, ... for a monthly series
  and an empty range for an annual one. Those branches were unreachable while the inference
  always said `D`. The helper is now unused and removed.

The same inference feeds `timeshift` on Dates, and DuckDB shares its inference between the two
operators, so both engines stay consistent across them.

Closes #934

## Checklist

- [x] Tests pass (`pytest`) — 5409 passed (duckdb), 4837 passed (pandas)


javihern98 (approved):
Looks good, thanks! 😊

PR #950, opened 2026-07-30
Fix #949: fill_time_series keeps operand Data Points that miss the grid

### LINKED ISSUE #949: fill_time_series drops an existing Data Point when its date is off the inferred grid (DuckDb)
### Initial Checks

### Summary

On the DuckDb engine, `fill_time_series` over a `Date` identifier can **remove** an input Data
Point. The operator builds the expected grid with `generate_series(min, max, INTERVAL ...)` and
joins the source onto it, so any Data Point whose date does not land on that grid is discarded
along with its measure values.

The reference manual (p147) only ever adds: "The operator **fills the possibly missing** Data
Points ... If this expected Data Point is missing, it is added to the Data Set." Nothing licenses
removing one, and the output is documented to contain "the same time series as the operand".

This bites month-end series in particular. Stepping one month at a time from `2020-01-31` gives
`2020-02-29`, then `2020-03-29`, `2020-04-29` — the step clamps in February and never returns to
month ends — so a genuine `2020-04-30` Data Point falls off the grid and is lost.

> [!WARNING]
> This is silent data loss: no error is raised, and the row count can still look plausible.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r <- fill_time_series(DS_1, single);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

# A month-end series: the last Data Point is the 30th, not the 29th.
data_df = pd.DataFrame(
    {
        "Id_1": ["A", "A", "A"],
        "Id_2": ["2020-01-31", "2020-02-29", "2020-04-30"],
        "Me_1": [1.0, 2.0, 3.0],
    }
)

run_result = run(
    script=script,
    datapoints=datapoints,
    use_duckdb=True,
)

```

### Expected vs Actual

Expected: every input Data Point is still present, with added Data Points for the missing
periods. `2020-04-30` and its value `3.0` must survive.

Actual: `2020-04-30` is absent from the result. Its measure value `3.0` is lost.

| Id_1 | Id_2 | Me_1 |
| --- | --- | --- |
| A | 2020-01-31 | 1.0 |
| A | 2020-02-29 | 2.0 |
| A | 2020-03-29 | |
| A | 2020-04-29 | |

The pandas engine keeps it, returning `2020-04-30` with `3.0` alongside the generated dates.

### vtlengine version

1.9.2rc2

### Python version

3.12.3

### OS

Ubuntu 24.04.4 LTS

### Additional Context

Affected code: `_fill_time_series_date` in `src/vtlengine/duckdb_transpiler/Transpiler/__init__.py`.
The `full_grid` CTE is generated from `bounds` alone, and the source is then joined onto it, so
off-grid source rows are never carried through. A union of the generated grid with the source
Data Points, rather than a join onto the grid alone, would preserve them.

> [!NOTE]
> Independent of #934, which concerns which frequency is inferred. This reproduces unchanged on
> 1.9.2rc2 and with #934 applied, because the grid join is on the DuckDb side only.

A related but separate divergence, worth its own issue if it matters: `timeshift` on a month-end
`Date` series snaps to the target month's last day on DuckDb (`2020-02-29` shifted by 1 gives
`2020-03-31`) while pandas adds a calendar month (`2020-03-29`). The manual does not state which
is intended, so that one is a semantics question rather than a defect.

albertohernandez1995:
## Summary

On the DuckDb engine, `fill_time_series` over a `Date` identifier could **remove** an operand
Data Point. Both fill paths ended with `FROM full_grid g LEFT JOIN source s`, so a Data Point
whose time value did not fall on the generated grid had nowhere to land and was dropped along
with its measure values. No error was raised.

Month-end series trigger it: stepping one month from `2020-01-31` clamps to `2020-02-29` and then
drifts to `2020-03-29`, `2020-04-29`, so a genuine `2020-04-30` Data Point is off the grid.

The reference manual (p147) has the operator only add: "The operator fills the possibly missing
Data Points ... If this expected Data Point is missing, it is added to the Data Set", and the
output is documented to contain "the same time series as the operand".

- Join onto the grid unioned with the operand's own keys, through a shared
  `_grid_with_source_keys` helper. `UNION` deduplicates, so keys already on the grid are not
  repeated.

Applied to the TimePeriod path as well as the Date one: the flaw is in the shape they share, even
though only the Date path can currently generate a grid that misses an operand key.

Fixes #949

## Checklist

- [x] Tests pass (`pytest`) — 5430 passed (duckdb), 4853 passed (pandas)

## Impact / Risk

- DuckDb only; the pandas engine already preserved these Data Points.
- Purely additive: the result gains the Data Points that were being dropped. Series whose dates
  all sit on the grid are unchanged, which is the common case and why this went unnoticed.
- Checked across 12 series shapes covering Date monthly, quarterly, annual, daily, irregular,
  month-end, single-point and multi-group cases plus annual, quarterly and mixed TimePeriod, in
  both `single` and `all` modes: no Data Point lost and none duplicated.

## Notes

The regression test lives in `tests/duckdb_transpiler/test_run.py`, the behavioural end-to-end
suite, rather than a data-file suite. The defect only appears when the inferred period is coarser
than daily, and until #934 lands the pandas engine still fills such a series daily, so a single
reference file cannot satisfy both CI legs. `tests/duckdb_transpiler` is skipped under
`VTL_ENGINE_BACKEND=pandas`, so the test is scoped to the affected engine while still asserting on
resulting rows rather than generated SQL. Once #934 is merged this can move to a data-file test if
preferred.


tests/duckdb_transpiler/test_run.py
javihern98, line 1: Need to change these tests to use the same pattern as others, maybe even as Bug Tests

javihern98 (changes requested):

tests/duckdb_transpiler/test_run.py
albertohernandez1995: Done it.

javihern98 (approved):
Looks good now, thanks! 😊

PR #951, opened 2026-07-30
Fix #944: Keep viral attributes in membership on the DuckDB backend

### LINKED ISSUE #944: Viral attributes are not maintained in membership (DuckDb)
### Initial Checks

### Summary

With this example, it return correct data structure but not the viral attribute.

[DS_1_datapoints.csv](https://github.com/user-attachments/files/30510509/DS_1_datapoints.csv)
[DS_1.json](https://github.com/user-attachments/files/30510510/DS_1.json)

### Reproducible Example

```python
ds_from_membership := DS_1#Me_int;
```

### Expected vs Actual

Expected: Maintain the viral attribute

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

javihern98:
## Summary

Fixes #944

With the DuckDB backend, membership (`DS_r := DS_1#Me_int;`) returned the correct output structure (including viral attributes) but the viral attribute data column was missing. The fetch layer silently skips components declared by semantic analysis but absent from the DuckDB table, so structure-side gaps surface as missing data.

**Membership (first commit).** Two sites diverged from the pandas reference (`Membership.validate`):

- `_visit_binop_membership` built its SELECT list from identifiers + target only, so viral columns never reached the DuckDB table. It now copies viral attributes through unchanged (membership is row-preserving, not a combination point — no propagation rule runs).
- `_build_membership_structure` dropped viral attributes from intermediate structures and wrongly promoted a viral target to a `COMP_NAME_MAPPING` measure, which made `DS_1#VAt_1 [calc Me_x := 1]` fail with a live BinderException (phantom `str_var`). It now mirrors `Membership.validate`.

**Intermediate structures (second commit).** The same structure gap existed in three more builders — `_build_ds_ds_binop_structure`, `_build_aggregation_structure`, `_build_boolean_result_structure` — losing viral data whenever a ds-ds binary, an aggregation, or a boolean-producing operator (`isnull`, `in`/`not_in`) sits **nested inside** a larger expression, e.g. `sum((DS_1 + DS_2) group by Id_1)`, `abs(DS_1 + DS_2)`, `sum(DS_1 group by Id_1)[calc Me_2 := Me_1]`, `isnull(DS_1)[keep bool_var]`. In `sum(DS_1 group by Id_1) + DS_2` the DuckDB backend also silently copied DS_2's viral value instead of pair-combining it with the aggregation's through the propagation rule. All three builders now carry viral attributes, mirroring the pandas results (the SQL emission already keyed off these structures, so no SQL changes were needed).

**Reserved component names for viral attributes (third commit).** The names the engine uses for automatically generated measures (`str_var`, `num_var`, `int_var`, `time_var`, `time_period_var`, `date_var`, `duration_var`, `bool_var`, `null_var` — the `COMP_NAME_MAPPING` values, now exported as `RESERVED_COMPONENT_NAMES`) are reserved: a **viral attribute** may not use them, since membership promotion would produce a measure colliding with the viral column. The check lives in `Component.__post_init__` (new error `1-3-3-7`), the single choke point through which every construction path flows — data-structure loading (both JSON layouts and SDMX), `calc viral attribute`, and `rename`. Other roles may still use these names: the engine itself outputs `bool_var`/`int_var` measures and those results must round-trip as inputs. The membership data path (second commit) still sources the promoted measure explicitly from the target component as defense-in-depth.

Tests follow the repo's file-based convention (`ViralHelper.BaseTest` + data files, reference outputs generated from pandas runs), added to `tests/ViralAttributes/test_viral_attributes.py` as three new groups covered by both CI backend lanes:

- **Group 11 — membership (9 cases)**: measure/identifier/attribute/viral targets, aggregate-rule per-row copy, unary/calc over membership, binary and aggregation over memberships.
- **Group 12 — nested intermediates (8 cases)**: ds-ds binary under aggregation/unary/keep, aggregation under binary/unary/calc, isnull/in under keep.
- **Group 13 — reserved names (4 cases)**: viral `int_var` declared in the input structure (load-time `1-3-3-7`), `rename` to `str_var`, `calc viral attribute bool_var`, and an allowed-roles case (`bool_var` measure / `int_var` attribute still load).

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- DuckDB now matches the pandas reference and the structure reported by semantic analysis in all fixed cases.
- Generated SQL for datasets without viral attributes is character-identical (existing transpiler SQL expectation tests unchanged).

> [!WARNING]
> New validation (breaking for affected inputs): declaring, calculating or renaming a **viral attribute** with one of the reserved component names (`str_var`, `num_var`, `int_var`, `time_var`, `time_period_var`, `date_var`, `duration_var`, `bool_var`, `null_var`) now raises `SemanticError 1-3-3-7`. Such structures were previously accepted and produced ambiguous, backend-dependent results when a membership promotion collided with the viral column. Identifiers, measures and plain attributes may keep using these names.

## Notes

> [!NOTE]
> The intermediate-structure gap originally flagged here as a follow-up is now fixed in this PR (second commit). The remaining structure builders (join, validation, hierarchy, unpivot, clauses) already carried viral attributes from the issue #877/#906 work.

- The `1-3-3-7` entry in `Exceptions/messages.py` flows into the auto-generated error documentation at docs build time; no manual docs change is needed.


albertohernandez1995 (approved):
Great job!

PR #953, opened 2026-07-30
Fix #945: expose derived scalar tables with a value column in DuckDB

### LINKED ISSUE #945: Error in equality (DuckDb)
### Initial Checks

### Summary

First line works fine but if i put second give me this error.
```
Execution error: Binder Error: Column "value" referenced that exists in the SELECT clause - but this column cannot be referenced before it is defined
```

And if I replace first with `ds_from_membership_3 := 9` also works fine.

### Reproducible Example

```python
ds_from_membership_3 := max(DS_1#Me_int)#Me_int;
in_member_one_row <- ds_from_membership_3 = 9;
```

### Expected vs Actual

Expected: works

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

javihern98:
## Summary

Fixes #945

With the DuckDB backend, referencing a derived scalar produced by a membership on an ungrouped aggregation (e.g. `x := max(DS_1#Me_int)#Me_int;`) from a later statement failed with:

```
Binder Error: Column "value" referenced that exists in the SELECT clause - but this column cannot be referenced before it is defined
```

The transpiler stores derived scalars as single-row tables read back as `(SELECT value FROM "<name>")`. In `visit_Start`, the scalar branch only added the `AS value` alias when the right-hand side transpiled to a bare expression; a full `SELECT` (what membership emits) was materialized as-is, so the table kept the measure column name (`Me_int`) and every downstream reference to the scalar failed to bind — in equalities, arithmetic, and inside dataset expressions alike. The pandas engine handled all of these correctly.

The fix folds full-`SELECT` scalar assignments into a scalar subquery aliased to `value`:

```sql
-- before
SELECT "Me_int" FROM (SELECT MAX("Me_int") AS "Me_int" FROM (SELECT "Id_1", "Me_int" FROM "DS_1") HAVING COUNT(*) > 0)
-- after
SELECT (SELECT "Me_int" FROM (SELECT MAX("Me_int") AS "Me_int" FROM (SELECT "Id_1", "Me_int" FROM "DS_1") HAVING COUNT(*) > 0)) AS value
```

This also matches the pandas engine on empty inputs: a zero-row aggregation now yields a one-row `NULL` scalar table instead of a zero-row table (which the result fetcher would have degraded to a `Dataset`).

### Interaction with #944 (#951)

After merging `1.9.X`, membership emits viral attribute columns (#951). On a dataset **without identifiers** membership is a scalar extraction (`Membership.validate` returns a `Scalar`), so a kept viral attribute made the scalar subquery fail with `Binder Error: Subquery returns 2 columns`. The membership visitor now drops viral attributes in the no-identifier case, mirroring the pandas semantics (a scalar has no attributes).

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- Breaking changes? No. Only the DuckDB scalar-assignment SQL shape changes; results now match the pandas engine.
- Notes for release/changelog? Bug fix for the DuckDB backend (`use_duckdb=True`).

## Notes

- Regression tests run under both backends via `VTL_ENGINE_BACKEND`:
  - `GH_945_1` — scalar consumed by an equality (the issue reproduction)
  - `GH_945_2` — scalar consumed inside a `calc` clause
  - `GH_945_3` — scalar membership over an ungrouped aggregation that kept a viral attribute (#951 interaction)
- Full suite verified locally under both backends after the `1.9.X` merge (DuckDB: 5460 passed, pandas: 4886 passed).


albertohernandez1995 (dismissed):
Great job!

albertohernandez1995 (approved):

PR #954, opened 2026-07-30
Fix #917: cast input scalar values to their declared type on the DuckDB path

### LINKED ISSUE #917: Error with parenthesis and scalars (DuckDB)
### Initial Checks

### Summary

We receive this error when we try to execute this code with `useDuckDb` parameter.
The scalar is a integer with value 1.

```
Error: Exception: Execution error: Binder Error: Could not choose a best candidate function for the function call "+(INTEGER_LITERAL, STRING_LITERAL)". In order to select one, please add explicit type casts.
	Candidate functions:
	+(INTEGER, DATE) -> DATE
	+(INT
```

### Reproducible Example

```python
parenthesis_sc := 100 - (99 + sc1);
```

### Expected vs Actual

Expected: work
Current: brings error

### vtlengine version

1.9.2rc1

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

javihern98:
## Summary

Fixes #917

With `use_duckdb=True`, an input scalar whose value was provided as a Python string (e.g. `scalar_values={"sc1": "1"}` for an `Integer` scalar) was inlined into the generated SQL as a string literal, so `parenthesis_sc := 100 - (99 + sc1);` failed with:

```
Binder Error: Could not choose a best candidate function for the function call "+(INTEGER_LITERAL, STRING_LITERAL)"
```

> [!NOTE]
> The issue says "the scalar is a integer with value 1": `sc1` is *declared* `Integer`, but the binder error proves the value reached the engine as a Python **string** — `_to_sql_literal` renders a true `int` unquoted (`99 + 1`, which works on 1.9.2rc1's identical scalar path), while only a `str` value takes the quoting branch that produced `99 + '1'` (`STRING_LITERAL`). The pandas backend accepts the same input because it casts `"1"` to `1` on load, so from the caller's side both backends received "an integer scalar with value 1".

The pandas path loads scalar values through `_handle_scalars_values`, which validates the name (`0-1-2-6`), checks castability (`0-1-2-7`) and **casts** the value to the declared data type. The DuckDB path (`_run_with_duckdb`) assigned the raw value instead — the `Scalar.value` setter only checks castability but stores the value as-is, so the transpiler rendered the `str` as a quoted SQL literal.

The fix replaces the ad-hoc assignment loop in `_run_with_duckdb` with the same `_handle_scalars_values` call the pandas path uses, so both backends load input scalar values identically.

## Checklist

- [x] Tests pass (`pytest`, both `VTL_ENGINE_BACKEND=duckdb` and `pandas`)

## Impact / Risk

- Breaking changes? Two intentional alignments of the DuckDB path with the documented pandas behavior:
  - An unknown scalar name in `scalar_values` now raises `InputValidationException` `0-1-2-6` (it was silently ignored before).
  - Values are cast to the declared type, so e.g. `1.0` for an `Integer` scalar now yields an `int` result instead of leaking a `float`.
- Notes for release/changelog? DuckDB backend now accepts string-typed `scalar_values` for numeric/boolean scalars, same as the pandas backend.

## Notes

- New file-based test `Bugs.GeneralBugs.test_GH_917` reproduces the issue script (`parenthesis_sc := 100 - (99 + sc1);` with `scalars={"sc1": "1"}`) and pins the scalar value through `DS_r <- DS_1 + parenthesis_sc;`, since scalar references are compared by name/type only.
- Verified the cast keeps SQL rendering intact for every scalar type: String/Date/Time_Period/Time/Duration casts return `str` (Date still renders as `DATE '...'`, TimePeriod still normalizes), while Integer/Number/Boolean become native Python types.


albertohernandez1995 (approved):
Great job!

Issue #931, opened 2026-07-28
Stock to flow weird error (DuckDb)


Pablolgn:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

With this code examples of code we receive this error.

```
Error: Exception: Execution error: Binder Error: No function matches the given name and argument types '-(BOOLEAN, BOOLEAN)'. You might need to add explicit type casts.
	Candidate functions:
	-(TINYINT) -> TINYINT
	-(TINYINT, TINYINT) -> TINYINT
	-(SMALLINT) ->
```

[DS_3_with_date.json](https://github.com/user-attachments/files/30454542/DS_3_with_date.json)
[DS_2_with_tp.json](https://github.com/user-attachments/files/30454541/DS_2_with_tp.json)

### Reproducible Example

```python
stock_to_flow_period_ds_tp := stock_to_flow(DS_2_with_tp);
stock_to_flow_period_ds_dt := stock_to_flow(DS_3_with_date);
```

### Expected vs Actual

Expected: Work like pandas

### vtlengine version

1.9.2rc2

javihern98:
Merged in #943

PR #955, opened 2026-07-31
Fix #935: flow_to_stock keeps an Integer measure integral on DuckDb

### LINKED ISSUE #935: Error flow_to_stock (DuckDb)
### Initial Checks

### Summary

We see a different behavior in flow_to_stock, with duckdb we receive wrong type error.

[DS_3_with_date.json](https://github.com/user-attachments/files/30460988/DS_3_with_date.json)
[DS_2_with_tp.json](https://github.com/user-attachments/files/30460990/DS_2_with_tp.json)
[DS_3_with_date_datapoints.csv](https://github.com/user-attachments/files/30460991/DS_3_with_date_datapoints.csv)
[DS_2_with_tp_datapoints.csv](https://github.com/user-attachments/files/30460989/DS_2_with_tp_datapoints.csv)

```
 [DIFF] Mod1.flow_to_stock_date_ds_dt.csv  (real=8, noise=1, equal=61)
      [REAL] fila 0 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 2 col 'Me_bool': 'True' -> '2.0'
      [REAL] fila 3 col 'Me_bool': 'False' -> '0.0'
      [REAL] fila 4 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 5 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 6 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 8 col 'Me_bool': 'True' -> '2.0'
      [REAL] fila 9 col 'Me_bool': 'False' -> '0.0'
  [DIFF] Mod1.flow_to_stock_date_ds_tp.csv  (real=8, noise=1, equal=61)
      [REAL] fila 0 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 2 col 'Me_bool': 'True' -> '2.0'
      [REAL] fila 3 col 'Me_bool': 'False' -> '0.0'
      [REAL] fila 4 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 5 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 6 col 'Me_bool': 'True' -> '1.0'
      [REAL] fila 8 col 'Me_bool': 'True' -> '2.0'
      [REAL] fila 9 col 'Me_bool': 'False' -> '0.0'

```

### Reproducible Example

```python
flow_to_stock_date_ds_tp := flow_to_stock(DS_2_with_tp);
flow_to_stock_date_ds_dt := flow_to_stock(DS_3_with_date);
```

### Expected vs Actual

Expected: have same bahaviour

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Two DuckDb time-operator fixes.

**#949** — `fill_time_series` on a `Date` identifier built its expected grid with
`generate_series` and joined the operand onto it, so any Data Point whose date did
not land on the generated grid was silently discarded together with its measure
values. The reference manual only ever adds Data Points, and documents the result as
containing "the same time series as the operand". Both fill paths now build their
keys as the generated grid `UNION` the operand's own keys.

**#935** — `flow_to_stock` accumulates with `SUM() OVER ()`, and DuckDb widens
`SUM()` over an integer to `HUGEINT`. `HUGEINT` has no pandas integer counterpart, so
an `Integer` measure came back as a float (`1.0`, `3.0`, `6.0`) while the component
was still declared `Integer`. The window sum is now cast back to the declared type.
`stock_to_flow` was unaffected because it subtracts rather than accumulating. The
`Number` branch is deliberately untouched: its DECIMAL accumulation is more accurate
than the pandas float64 one, so casting it to `DOUBLE` would change rendering for no
benefit.

The Boolean symptom originally reported in #935 was already resolved by #931; only
the Integer widening remained.

Review feedback applied: the #949 regression test moved from
`tests/duckdb_transpiler/test_run.py` to `TimeBugs` in `tests/Bugs/test_bugs.py`
using the standard per-code fixtures. That directory is skipped under the pandas
backend, so the original placement exercised only one engine.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- No API change.
- `fill_time_series` now returns Data Points it previously dropped, so results can
  gain rows relative to 1.9.2rc2. That is the documented behaviour, and the dropped
  rows were data loss.
- `flow_to_stock` now returns an integer-typed column where it previously returned a
  float. Values are unchanged, and CSV output was already rendering them as integers,
  so the visible effect is limited to in-memory dtypes.
- Verified against the VTL 2.2 reference manual examples for `fill_time_series`:
  examples 3-8 (`date`, `time_period`, and the mixed annual/quarterly case) match on
  both engines, including which Data Points are null-filled.


tests/Bugs/test_bugs.py
```diff
@@ -1460,6 +1462,53 @@ def test_duckdb_derived_scalar_chain(self):
...
+                return_only_persistent=False,
+                use_duckdb=engine,
+            )
+            for name in ("DS_r1", "DS_r2"):
+                measure = result[name].data["Me_2"]
+                assert result[name].components["Me_2"].data_type == Integer
+                assert is_integer_dtype(measure), (
+                    f"engine={engine}, {name}: Me_2 is declared Integer but came "
+                    f"back as {measure.dtype}"
+                )
```
javihern98: This test is not needed, in any case this is already tested via the dataload tests and the Base Test. Please remove it

albertohernandez1995: Removed.

javihern98 (approved):
Looks good, thanks! 😊

PR #956, opened 2026-07-31
Fix #936: Add tests to pin the Date period inferred by timeshift

### LINKED ISSUE #936: Time_shift behavior change (DuckDb)
### Initial Checks

### Summary

We have a behaviour change in time_shift with dates. 

Instead of return 1 day now return 1 month.

Maybe related with https://github.com/Meaningful-Data/vtlengine/issues/934

[DS_3_with_date.json](https://github.com/user-attachments/files/30462979/DS_3_with_date.json)
[DS_3_with_date_datapoints.csv](https://github.com/user-attachments/files/30462980/DS_3_with_date_datapoints.csv)

 [DIFF] Mod1.timeshift_date_ds_dt.csv  (real=10, equal=60)
      [REAL] fila 0 col 'Id_date': '2000-12-31' -> '2000-12-01'
      [REAL] fila 1 col 'Id_date': '2002-01-31' -> '2002-01-01'
      [REAL] fila 2 col 'Id_date': '2003-02-28' -> '2003-02-01'
      [REAL] fila 3 col 'Id_date': '2004-03-31' -> '2004-03-01'
      [REAL] fila 4 col 'Id_date': '2004-12-31' -> '2004-12-01'
      [REAL] fila 5 col 'Id_date': '2006-01-31' -> '2006-01-01'
      [REAL] fila 6 col 'Id_date': '2007-02-28' -> '2007-02-01'
      [REAL] fila 7 col 'Id_date': '2008-03-31' -> '2008-03-01'
      [REAL] fila 8 col 'Id_date': '2008-12-31' -> '2008-12-01'
      [REAL] fila 9 col 'Id_date': '2010-01-31' -> '2010-01-01'

### Reproducible Example

```python
timeshift_date_ds_dt := timeshift(DS_3_with_date, -1);
```

### Expected vs Actual

Expected: Have same behaviour

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

#936 reports that `timeshift` on a `Date` series shifts by a month where it used to
shift by a day. That is the intended consequence of #934 and not a defect, so this PR
changes no behaviour — it pins the current inference so it cannot regress.

The #934 fix touched only `src/vtlengine/Operators/Time.py`. DuckDb shifted by a month
before and after it; what changed is that the pandas engine now agrees, which is what
#934 asked for. The "1 day" reference in #936 is the old pandas result.

The month is also the correct inference for that Data Set: every date falls on the
first of a month and every gap is a whole number of months (13, 13, 9, …), so the
largest period fitting all gaps is one month. The old result came from the day-count
heuristic #934 removed, which read any gap that was not an exact multiple of 365 or
366 days as a day-long period.

`test_GH_936_1` covers the reported Data Set plus an unambiguously daily series, so a
future change cannot trade one inference for the other unnoticed. The issue's
single-point group is deliberately excluded: it has no gaps, so its period is not
determined by the data.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- No source change, so no behaviour or API change.
- Verified the test fails against the pre-#934 heuristic and passes on both engines
  with it, so it is a real pin rather than a restatement of current output.

Closes #936


javihern98 (commented):
Need to decide how to handle the follow-up specified in the notes

albertohernandez1995:
> Need to decide how to handle the follow-up specified in the notes

Removed the note. That behaviour is a defect on its own and now has its own task, so it does
not belong in this PR. Nothing changed here.

javihern98 (approved):
Looks good, thanks! 😊

Issue #918, opened 2026-07-22
Error flow_to_stock() with time id (Duckdb)


Pablolgn:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

Running this code with this datasets that contain time id with duckdb raise this error.

`cannot unpack non-iterable NoneType object`

dataset: [DS_4_with_time.json](https://github.com/user-attachments/files/30259355/DS_4_with_time.json)

datapoints:[DS_4_with_time_datapoints.csv](https://github.com/user-attachments/files/30259358/DS_4_with_time_datapoints.csv)

### Reproducible Example

```python
low_to_stock_date_ds_tm := flow_to_stock(DS_4_with_time);
```

### Expected vs Actual

Expected: works like pandas engine

### vtlengine version

1.9.2rc1

Pablolgn:
We have same error with `stock_to_flow_period_ds_tm := stock_to_flow(DS_4_with_time);`

Pablolgn:
Same error also in `fill_time_series(DS_4_with_time, single);`

Pablolgn:
also with `timeshift(DS_4_with_time, 1);` seems the problem is with all time operators.

javihern98:
Implemented in #958

PR #968, opened 2026-08-04
Fix #961: measure the expected Date grid from the lower limit

### LINKED ISSUE #961: Fill_time_series builds a drifting Date grid instead of anchoring to period ends
### Initial Checks

### Summary

`fill_time_series` over a `Date` identifier builds its expected grid by adding one period at a
time to the lower limit. When the lower limit is a period end, the step clamps in the shorter
month and then carries the clamped day forward for the rest of the series, so the grid drifts off
the period ends.

The reference manual sets the convention for date-typed series in its examples: for `DS_2`,
"`Id_2` is the reference time Identifier of date type and conventionally each period is
identified by its last day". Under that convention a monthly series starting at `2020-01-31` has
expected Data Points `2020-02-29`, `2020-03-31`, `2020-04-30`, not `2020-03-29` and `2020-04-29`.

Both engines behave the same way, so this is a semantics choice rather than a backend
divergence.

> [!NOTE]
> DuckDb already applies the last-day convention in `timeshift`, snapping to `LAST_DAY` when the
> input is a month end and the period is `M`, `Q`, `S` or `A`. `fill_time_series` is inconsistent
> with it on the same engine.

Not the same as #949, which was about an existing off-grid Data Point being dropped, or #934,
which was about which frequency is inferred. Both of those are fixed and this still reproduces;
it is the reason a Data Point was off the grid in #949

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r <- fill_time_series(DS_1, single);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

# Quarter ends: 31 Mar, 30 Jun, 31 Dec.
data_df = pd.DataFrame(
    {
        "Id_1": ["A", "A", "A"],
        "Id_2": ["2020-03-31", "2020-06-30", "2020-12-31"],
        "Me_1": [1.0, 2.0, 3.0],
    }
)

run_result = run(
    script=script,
    datapoints=datapoints,
    use_duckdb=True,
)

```

### Expected vs Actual

Expected: the generated Data Point for Q3 is `2020-09-30` and the series holds four quarter ends.

Actual: a spurious `2020-12-30` appears next to the real `2020-12-31`, because the grid stepped
`06-30`, `09-30`, `12-30`.

| Id_1 | Id_2 | Me_1 |
| --- | --- | --- |
| A | 2020-03-31 | 1.0 |
| A | 2020-06-30 | 2.0 |
| A | 2020-09-30 | |
| A | 2020-12-30 | |
| A | 2020-12-31 | 3.0 |

Monthly series are worse, since the drift starts one step in:

| input | actual | expected |
| --- | --- | --- |
| `2020-01-31, 2020-02-29, 2020-04-30` | `01-31, 02-29, 03-29, 04-29, 04-30` | `01-31, 02-29, 03-31, 04-30` |

Mid-month series such as `2020-01-15, 2020-02-15, 2020-04-15` come out correct, so only
period-end anchoring is affected.

### vtlengine version

1.9.2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`fill_time_series` over a `Date` identifier built its expected grid one period at a time from
the previous date. A month-based step out of a period end clamps in the shorter month, and
carrying that clamped day forward drifted the grid off the period ends: from `2020-01-31` it
reached `2020-02-29` and then `2020-03-29`, producing dates the series never uses and, for the
quarterly case, a spurious `2020-12-30` beside the real `2020-12-31`.

The manual identifies a date-typed period by its last day: "`Id_2` is the reference time
Identifier of date type and conventionally each period is identified by its last day".

The fix needs no period-end detection. Measuring each date from the lower limit rather than from
its predecessor produces the expected grid directly, because that is how both `pd.DateOffset` and
DuckDb's `INTERVAL * n` clamp a short month:

| input | before | after |
| --- | --- | --- |
| `2020-01-31, 2020-02-29, 2020-04-30` | `01-31, 02-29, 03-29, 04-29, 04-30` | `01-31, 02-29, 03-31, 04-30` |
| `2020-03-31, 2020-06-30, 2020-12-31` | `03-31, 06-30, 09-30, 12-30, 12-31` | `03-31, 06-30, 09-30, 12-31` |
| `2020-01-15, 2020-02-15, 2020-04-15` | unchanged | unchanged |

This also covers the TimeInterval grids added by #918: the pandas path shares `_period_range`,
while the DuckDb path had its own `generate_series` and would otherwise have kept drifting,
leaving the two engines disagreeing on TimeInterval fills.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- Results change for `Date` and `Time` series whose dates are period ends: the generated Data
  Points move onto the period ends, and grids that previously emitted a date one day short of a
  real Data Point no longer do. Worth a changelog entry.
- Mid-month and day-based series are unaffected, which the test covers explicitly with `DS_3`.
- Verified against the manual's `fill_time_series` examples 3-8 (`date`, `time_period` and the
  mixed annual/quarterly case): all still conform on both engines, including which Data Points
  are null-filled.

## Notes

`test_GH_949_1` changed its input dates. With the grid no longer drifting, its `2020-04-30` lands
on the grid, so the original data no longer leaves any Data Point off it and the case stopped
exercising #949. The dates are now `2020-01-30, 2020-02-29, 2020-03-31`, where the grid holds
`01-30, 02-29, 03-30` and the operand's `2020-03-31` is genuinely off-grid and still reaches the
result. Its docstring records why.

---

Closes #961


mla2001 (approved):
Looks fine! 😊

PR #970, opened 2026-08-04
Fix #962: datediff accepts a Date and a Time_Period together

### LINKED ISSUE #962: datediff operator can not mix types
### Initial Checks

### Summary

Following the [documentation](https://sdmx-twg.github.io/vtl/v2.2/reference_manual/operators/Time%20operators/Days%20between%20two%20dates/index.html#examples-of-valid-syntaxes) we can use 2 different types in the operator.

With 2 dates works fine and with 2 time_periods also.\

Same behavior in both engines.

### Reproducible Example

```python
dd_lit_3 := datediff (cast("2021Q2", time_period), cast("2021-11-04", date));
```

### Expected vs Actual

Expected: Documentation example works fine.

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`datediff` raised `1-1-1-2` whenever its operands were a `Date` and a `Time_Period`:
('Invalid implicit cast from Time_Period and Date to Time. ...', '1-1-1-2').

Two dates worked, and two time periods worked, but not one of each, on both engines.

The [reference manual](https://sdmx-twg.github.io/vtl/v2.2/reference_manual/operators/Time%20operators/Days%20between%20two%20dates/index.html)
lists `datediff (2021Q2, 2021-11-04)` among its valid syntaxes, types both operands as
`component<time> | time`, states "Additional Constraints: None", and specifies that "the last day
of the time period is assumed as the starting/ending date". Its own Example 1 mixes a `Date`
Identifier with a `Time_Period` Measure.

`SimpleBinaryTime.validate_type_compatibility` blocked the pair explicitly, while
`Date_Diff.py_op` already resolved either operand correctly, taking the last day of the period
for a `Time_Period`. So only validation was in the way. `Date_Diff` is the only subclass of
`SimpleBinaryTime`, so removing the override affects nothing else and lets the ordinary implicit
promotion apply. A class docstring records why there is no stricter rule.

Both engines now reproduce the manual's published example exactly:

| Id_2 (Date) | Me_1 (Time_Period) | result | manual |
| --- | --- | --- | --- |
| 2019-01-01 | 2020Q2 | 546 | 546 |
| 2019-07-01 | 2021Q1 | 639 | 639 |
| 2020-12-31 | 2021Q1 | 90 | 90 |

The issue's own case gives 127 in either operand order, and same-type results are unchanged.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- `datediff` accepts a mixed `Date` / `Time_Period` pair instead of raising. Nothing that worked
  before changes value.
- Genuinely invalid operands still raise: `25` (two Integer Measures) keeps `1-1-1-2`, and the two
  cases casting `"2022Q1"` to `date` keep `RunTimeError 2-1-19-8`.
- Only `datediff` is affected, since `Date_Diff` is the sole subclass of `SimpleBinaryTime`.

--

Closes #962


javihern98 (approved):
Looks good, thanks! 😊

PR #972, opened 2026-08-05
Fix #963: reject an unknown periodInd in dateadd

### LINKED ISSUE #963: Incorrect type period do not return error (DuckDb)
### Initial Checks

### Summary

In the existent documentation we have this example, that should be incorrect due to in other examples we use `A` for year time period. With duck db if we introduce a non correct time period we receive an empty dataset instead of an execution error as is pandas. 

### Reproducible Example

```python
da_4 := dateadd (cast("2020-02-10", date), 1, "Y");
```

### Expected vs Actual

Expected: Raise an error
Current: Return empty datapoints.

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`dateadd` accepted a `periodInd` that names no period. `Date_Add.validate` checked that the
params were a Scalar Integer and a Scalar String, but never that the string was one of `A`, `S`,
`Q`, `M`, `W`, `D`, so the value reached evaluation and each engine failed differently:

| engine | before |
| --- | --- |
| DuckDb | returned a null, no error |
| pandas | bare `KeyError: 'Y'` |

The issue asks for DuckDb to raise "an error as is pandas", but a `KeyError` is not a VTL error
either, so both engines needed fixing rather than one being aligned to the other.

The value is a literal at validation time, so it is now rejected there with `2-1-19-2`, the code
that already exists for this ("Invalid period indicator {period}."). Validation is shared between
the engines, so the single check covers both.

dateadd(date, 1, "Y")  invalid   pandas=SemanticError(2-1-19-2)   duckdb=SemanticError(2-1-19-2)
dateadd(date, 1, "A")  valid     pandas=OK '2021-02-10'           duckdb=OK '2021-02-10'

Checked across the scalar, component and dataset forms, and across all six valid indicators; the
valid results are unchanged on both engines.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- A script passing an unknown `periodInd` to `dateadd` now fails semantic analysis instead of
  producing a null (DuckDb) or a `KeyError` (pandas). Anything that used a valid indicator is
  unaffected.
- The check is guarded on the value being a string, so a derived scalar with no value at
  validation time is not rejected.
- `time_agg`, the other operator taking a period indicator, already validated its value and needed
  no change.

---

Closes  #963


javihern98 (approved):
Looks good, thanks! 😊

PR #973, opened 2026-08-05
Fix #964: dateadd shifts a Data Set's reference time Identifier

### LINKED ISSUE #964: dateadd do not work in datasets
### Initial Checks

### Summary

Seems the operator do not work for datasets, we receive same data points with any change. This only happens in pandas engine.

[DS_3_with_date_datapoints.csv](https://github.com/user-attachments/files/30666498/DS_3_with_date_datapoints.csv)
[DS_3_with_date.json](https://github.com/user-attachments/files/30666499/DS_3_with_date.json)

### Reproducible Example

```python
da_ds := dateadd(DS_3_with_date,  2, "W" );
```

### Expected vs Actual

Expected: Works with datasets as with scalars or components

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`dateadd` over a Data Set ignored the operand's reference time Identifier, and neither engine
limited itself to time Components:

| engine | before |
| --- | --- |
| pandas | shifted Date and Time_Period **Measures** only, so a Data Set whose time Component is an Identifier came back unchanged |
| DuckDb | applied the shift to **every** Measure, aborting with `Invalid date format, must be YYYY-MM-DD: hello` on a String Measure |

The [reference manual](https://sdmx-twg.github.io/vtl/v2.2/reference_manual/operators/Time%20operators/Add%20time%20unit%20to%20date/index.html)
types the operand as `dataset { identifier <time> _ , identifier _* }`, the same signature as
`timeshift`, so the reference time Identifier is what moves.

Both engines now share one rule: shift the time Identifier when the Data Set has one, otherwise
its Date or Time_Period Measures, and pass every other Component through.

On the Data Set from the issue, both engines now give:

| Id_date in | Id_date out | Me_str |
| --- | --- | --- |
| 2001-01-01 | 2001-01-15 | hello, unchanged |

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- A Data Set with a time Identifier now has that Identifier shifted rather than being returned
  unchanged (pandas) or failing (DuckDb). A uniform shift is injective, so the Identifier stays
  unique.
- Non-time Measures are no longer touched on DuckDb, which is what made the query abort.
- Data Sets whose only time Components are Measures keep their current behaviour, so the 12
  existing Data Set cases are unaffected.

## Notes

The Measure fallback is deliberate. The 12 existing Data Set cases pass an operand with an Integer
Identifier and a Date Measure, which does not match the documented signature; rejecting those would
be a breaking change, so they keep working rather than having a removal bundled into this fix. That
stricter alignment is worth its own task if it is wanted.

Code `28` in `ds_param` covers a Date Identifier alongside Number and String Measures, pinning both
halves of the rule. It fails on both engines against the previous code.

---

Closes #964

### DIFF
--- a/src/vtlengine/Operators/Time.py
+++ b/src/vtlengine/Operators/Time.py
@@ -1070,6 +1070,22 @@ class Date_Add(Parametrized):
     op = DATE_ADD
 
     @classmethod
+    def _dateadd_targets(cls, operand: Dataset) -> List[Component]:
+        """The Components a Data Set operand shifts.
+
+        The reference manual types the operand as ``dataset {identifier<time> _,
+        identifier _*}``, so the reference time Identifier is what moves. A Data Set
+        whose only time Components are Measures is accepted as well, and those Measures
+        move instead, which is what the operator did before it reached an Identifier at
+        all. Components of any other type are left alone either way, since
+        there is no date to add to.
+        """
+        time_ids = [c for c in operand.get_identifiers() if c.data_type in (Date, TimePeriod)]
+        if time_ids:
+            return time_ids
+        return [m for m in operand.get_measures() if m.data_type in (Date, TimePeriod)]
+
+    @classmethod
     def validate(
         cls, operand: Union[Scalar, DataComponent, Dataset], param_list: List[Scalar]
     ) -> Union[Scalar, DataComponent, Dataset]:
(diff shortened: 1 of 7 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #975, opened 2026-08-05
Fix #969: hierarchy follows the reference-manual roll-up semantics for nulls and missing data points

### LINKED ISSUE #969: hierarchy errors
### Initial Checks

### Summary

I found 2 errors with hierarchy and the hierarchical rule sets.
1. Not same result in last [example](https://sdmx-twg.github.io/vtl/v2.2/reference_manual/operators/Hierarchical%20aggregation/Hierarchical%20roll-up/index.html#example-3), we ahve one lees row the H one.
2. Null results in viral attributes are returnded as None.

The doc d not pass the exact viral propagation rule and i need to do create a library the match with the results.

[DS_1_hierarchy.csv](https://github.com/user-attachments/files/30697926/DS_1_hierarchy.csv)
[DS_1_hierarchy.json](https://github.com/user-attachments/files/30697925/DS_1_hierarchy.json)

### Reproducible Example

```python
define hierarchical ruleset HR_1(valuedomain rule VD_1) is
	A = J + K + L;

	B = M + N + O;

	C = P + Q;

	D = R + S;

	E = T + U + V;

	G = B + C;

	H = D + E;

	I = D + G;

	F = Y + W + Z
end hierarchical ruleset;

define viral propagation At_1(variable At_1) is
	when "None" then "None";
	when "Ay" then "None";
	when "Bq" then "Bq";
	when "Dx" then "Dx";
	when "Ko" then "Ko";
	when "Nj" then "Nj";
	when "Pz" then "Pz"
end viral propagation;



*       Module md

md.hr_nn_2 :=
	hierarchy(
		DS_1_hierarchy,
		HR_1
		rule Id_2);
md.hr_nz_2 :=
	hierarchy(
		DS_1_hierarchy,
		HR_1
		rule Id_2 non_zero);
md.hr_pn_2 :=
	hierarchy(
		DS_1_hierarchy,
		HR_1
		rule Id_2 partial_null);
```

### Expected vs Actual

Expected: Have same result that in docs and return null correctlly

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

javihern98:
## Summary

`hierarchy` deviated from VTL 2.2 reference-manual roll-up Examples 2–3 in two ways (#969), and crashed on a related edge:

- `partial_null`/`partial_zero` required at least one **non-null** involved data point, dropping computed nodes whose children all exist with null measures (the missing `H = D + E` row in Example 3). The spec requires only **existence**; a null measure still exists. Fixed in the interpreter partial mask and the DuckDB mode filter.
- A **missing** child data point contributed nothing to the viral-attribute combination, so the propagation rule saw only the surviving siblings (`D = R + S` yielded `Ay` instead of null). Per the manual, the attribute value of a missing data point is null. Each child is now padded to the node's group universe (pandas merge / SQL LEFT JOIN onto a per-node group CTE) before the propagation rule reduces the values.
- A dataset whose **only identifier is the rule component** crashed on both backends (pandas `IndexError` from a merge on an empty key list; DuckDB `LEFT JOIN … USING (1=1)` parse error, plus a presence guard that mistook a NULL computed value for a missing rule row). The pandas path now uses a one-row global group universe, and the SQL rule CTEs carry a constant `_present` marker joined with `ON TRUE`.

`when null then null` clauses already parsed and resolved correctly; no grammar changes. Both backends fixed symmetrically; `RM134`'s pinned output gains the docs' `H` row, `GH_969_1` pins all three docs examples end-to-end with a viral attribute, and `GH_969_2` pins the same examples over a single-identifier dataset.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- Breaking changes? Behavioral: partial modes now also produce rows whose involved points all exist with null measures (roll-up), and evaluate such rules (check_hierarchy); missing children now propagate null viral values; single-identifier hierarchies execute instead of raising. All are spec alignments.
- Notes for release/changelog? Fixes hierarchical roll-up conformance with VTL 2.2 RM Examples 2–3.

## Notes

Pinned reference outputs updated, each audited against the spec's existence rule ("the rule is computed/evaluated when at least one involved data point exists; missing data points are null"):

- `tests/ReferenceManual` 134 (roll-up partial_null): the manual's Example 3 `H` row (H = D + E, both computed nulls).
- `tests/Hierarchical` 2-1-1-3/4 (roll-up partial): `B` row where C exists with null and D is missing; 2-1-1-19/20: `N` row where A and L exist with null; 2-1-1-29/30 (`output all`): the computed null `N` now replaces the operand's value, per the `all` semantics.
- `tests/Hierarchical` 1-1-1-21/22/27/31 and `tests/Additional` 11-13/14/19/20/23/24/27/28 (check_hierarchy partial, `all`/`all_measures`): rules whose involved points all exist with null measures are now evaluated, adding rows with `bool_var` null.

Fixes #969


albertohernandez1995 (approved):
Good effort!

Issue #921, opened 2026-07-23
Different number of results in DPR (Duckdb)


Pablolgn:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

Testing with a project I notice that in some case the results are different, Pandas=14505 , DuckDb=15802.

Example: [error_anaval.txt](https://github.com/user-attachments/files/30305175/error_anaval.txt)

DataStructure:[ANCRDT_FNNCL_C.json](https://github.com/user-attachments/files/30305191/ANCRDT_FNNCL_C.json)

DataPoints: [ANCRDT_FNNCL_C_datapoints.csv](https://github.com/user-attachments/files/30305200/ANCRDT_FNNCL_C_datapoints.csv)

### Reproducible Example

```python
Pass as file
```

### Expected vs Actual

Expected: Brings same results

### vtlengine version

1.9.2rc2

javihern98:
Merged in #932

Pablolgn:
I tested the problematic project and still fail with 1.9.2rc3

Pablolgn:
I test it only with the code and data that i attached in the issue and still have same divergence in the results.

Pablolgn:
I  am tested pandas with 1.9.2rc2 Idk if this is problematic due to was an error in pandas engine.

Pablolgn:
Yeah seems that this is a bug fixed in duckDb that after the merge was solved also in pandas. i tested with last version in both engines and gives same output.

PR #977, opened 2026-08-06
Fix #976: DuckDB outputs materialize the declared component dtypes

### LINKED ISSUE #976: DuckDB backend returns column dtypes that contradict the declared output structure (errorlevel int32 vs Number)
### Initial Checks

### Summary

With `use_duckdb=True`, output Data Sets are materialized directly from `fetchdf()` with whatever physical type DuckDB inferred for each column, and the result is never reconciled against the declared output structure. The pandas backend does enforce the declared structure on the data (`Dataset.enforce_dtypes()`, called by the Interpreter after each statement), so the two backends can return the same declared structure with different data types.

The most visible case is `errorlevel` in `check_datapoint` / `check_hierarchy`: the semantic layer declares it as `Number`, but the transpiler emits the error level as a bare SQL literal (`2 AS "errorlevel"`), which DuckDB types as `INTEGER`. The fetched DataFrame therefore holds `int32` (`2`), while the pandas backend holds `double[pyarrow]` (`2.0`) — even though `semantic_analysis` reports `Number` for both.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = """
    define datapoint ruleset DR_1 (variable Me_1) is
        R_1: Me_1 >= 0 errorcode "R_1" errorlevel 2
    end datapoint ruleset;

    DS_r <- check_datapoint(DS_1, DR_1);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

data_df = pd.DataFrame({"Id_1": [1, 2, 3], "Me_1": [10, -5, -30]})

    run_result = run(
        script=script,
        datapoints=datapoints,
        use_duckdb=use_duckdb,
    )
    ds = run_result["DS_r"]
    print(f"use_duckdb={use_duckdb}")
    print(f"  declared errorlevel type: {ds.components['errorlevel'].data_type.__name__}")
    print(f"  errorlevel data dtype:    {ds.data['errorlevel'].dtype}")
```

### Expected vs Actual

Expected: both backends materialize `errorlevel` with the declared component dtype (`Number` → `double[pyarrow]`), so comparing backend outputs shows no type difference.

Actual:

| backend | declared `errorlevel` type | data dtype | value |
| ------- | -------------------------- | ---------- | ----- |
| pandas | Number | `double[pyarrow]` | `2.0` |
| DuckDB | Number | `int32` | `2` |

### vtlengine version

1.9.2rc3

### Python version

3.12.3

### OS

Linux

### Additional Context

> [!NOTE]
> Found while verifying #921: the row-count divergence there is fixed in 1.9.2rc3, but diffing the two backends' in-memory outputs still shows this dtype mismatch.

Root cause:

- The pandas backend casts every output column to its declared component dtype (`Dataset.enforce_dtypes()` in `src/vtlengine/Model/__init__.py`, called from `Interpreter/__init__.py`, using `DTYPE_MAPPING` in `src/vtlengine/DataTypes/__init__.py`).
- The DuckDB backend has no equivalent step: `_build_dp_rule_sql` (`src/vtlengine/duckdb_transpiler/Transpiler/__init__.py`) emits the errorlevel literal untyped, and the result is fetched via `fetchdf()` in `src/vtlengine/duckdb_transpiler/io/_execution.py` as-is — and `fetchdf()` always returns numpy-backed dtypes, so the physical SQL type leaks into the output.

Proposed fix, in two layers:

1. **In-memory fetch**: after `fetchdf()` in `src/vtlengine/duckdb_transpiler/io/_execution.py`, cast the fetched DataFrame to the declared component dtypes (reusing `Dataset.enforce_dtypes()`, the same enforcement the pandas Interpreter applies). This makes `run()` outputs dtype-identical across backends (`DTYPE_MAPPING`, pyarrow-backed).
2. **Fetch/COPY SQL**: in `_build_dataset_fetch_select`, `CAST` each column to the SQL type of its declared component data type (`Number` → `DOUBLE`, `Integer` → `BIGINT`, `Boolean` → `BOOLEAN`, `String` → `VARCHAR`) when the physical type differs. This covers the file-output path (`COPY` writes directly from DuckDB) and normalizes renderings such as `DECIMAL` values written as `19.0000000000` where the pandas engine writes `19`.

> [!NOTE]
> Layer 1 adds a dtype-conversion pass on in-memory outputs, which the DuckDB path avoided until now; an Arrow-native fetch (`types_mapper=pd.ArrowDtype`) could mitigate the copy cost if it matters.

`errorlevel` is just the most visible instance; the enforcement applies generically to every operator's output columns.

javihern98:
## Summary

DuckDB in-memory outputs now pass through `Dataset.enforce_dtypes()` — the same enforcement the pandas `Interpreter` already applies — so both backends materialize identical dtypes for a given declared structure. The fetch/COPY SELECT casts physical/declared type mismatches, and CSV output renders Number columns using the pandas `float_format` equivalent (`printf('%.Ng', CAST(col AS DOUBLE))`, driven by `OUTPUT_NUMBER_SIGNIFICANT_DIGITS`).

Both backends declared the same output structure but materialized different dtypes: `errorlevel` came back as `int32` under DuckDB vs `double[pyarrow]` under pandas (issue #976, found while verifying #921). Separately, DuckDB-written CSV files rendered Number values as `19.0000000000` where pandas writes `19`.

Fixes #976

## Checklist

- [x] Tests pass (`pytest`)
- [x] Documentation updated (if applicable)

## Impact / Risk

- Breaking changes? DuckDB-backend outputs only — the pandas engine is untouched (verified by a full `VTL_ENGINE_BACKEND=pandas` run). In-memory DuckDB results now come back pyarrow-backed (matching pandas) instead of native DuckDB/NumPy dtypes.
- Data/SDMX compatibility concerns? Values are unchanged everywhere. DuckDB-written CSV Number rendering now matches pandas parity (e.g. `19` instead of `19.0000000000`). Parquet columns with a physical/declared type mismatch are now cast to the declared type (e.g. `errorlevel` is `DOUBLE` instead of `int32`); passthrough Number columns remain `DECIMAL(28,10)`, the engine's canonical internal representation.
- Notes for release/changelog? Bug fix — DuckDB outputs now materialize the declared component dtypes, aligning with the pandas backend.

## Notes

- Full suite verified green in both backend modes: DuckDB backend 5530 passed / 27 skipped; pandas backend 4918 passed / 639 skipped.
- Out of scope (left for separate follow-up issues if wanted):
  - The `check` vs `check_datapoint` errorlevel *declaration* asymmetry (`Integer` vs `Number` for all-int errorlevels in `src/vtlengine/Operators/Validation.py`) — public-structure decision.
  - The internal `DECIMAL(28,10)` representation of Number inside the DuckDB pipeline (`get_column_sql_type`) is deliberate precision design and stays untouched.
  - Arrow-native fetch (`types_mapper=pd.ArrowDtype`) as an `astype`-copy optimization — mentioned in #976 as a possible follow-up, not needed for correctness.


albertohernandez1995 (approved):
Good job🤗

PR #979, opened 2026-08-06
Fix #960: timeshift reads the period from each time series

### LINKED ISSUE #960: Timeshift uses one period for the whole Data Set instead of one per time series
### Initial Checks

### Summary

`timeshift` over a `Date` identifier infers one period from the whole time Identifier column and
applies it to every time series in the Data Set.

The reference manual defines the period per series: the operator shifts, "for each time series
of the Data Set", the reference time Identifier by a number of periods "of the time series".

A Data Set holding a daily series and a monthly series therefore shifts both by whatever period
the combined column suggests. Both engines behave the same way.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r <- timeshift(DS_1, -1);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

# Series A is daily, series B is monthly.
data_df = pd.DataFrame(
    {
        "Id_1": ["A", "A", "A", "B", "B", "B"],
        "Id_2": [
            "2020-01-01", "2020-01-02", "2020-01-03",
            "2020-01-01", "2020-02-01", "2020-03-01",
        ],
        "Me_1": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0],
    }
)

run_result = run(
    script=script,
    datapoints=datapoints,
    use_duckdb=True,
)

print(run_result["DS_r"].data.sort_values(["Id_1", "Me_1"]))
```

### Expected vs Actual

Expected: series A shifts back one day, series B one month.

Actual: series B shifts back one day, because the combined column contains one-day gaps.

| Id_1 | Id_2 in | Id_2 out | expected |
| --- | --- | --- | --- |
| A | 2020-01-01 | 2019-12-31 | 2019-12-31 |
| A | 2020-01-02 | 2020-01-01 | 2020-01-01 |
| A | 2020-01-03 | 2020-01-02 | 2020-01-02 |
| B | 2020-01-01 | 2019-12-31 | **2019-12-01** |
| B | 2020-02-01 | 2020-01-31 | **2020-01-01** |
| B | 2020-03-01 | 2020-02-29 | **2020-02-01** |

### vtlengine version

1.9.2

### Python version

3.12.3

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`timeshift` inferred a single period from the whole time Identifier column and applied it to
every time series. The manual requires the reference time Identifier to be
[periodical](https://sdmx-twg.github.io/vtl/v2.2/reference_manual/typical_behaviour.html#behaviour-of-time-operators)
— "its possible values are regularly spaced and therefore have constant duration (frequency)" —
so a Data Set holding a daily series beside a monthly one has no period to shift by, and the
operator now rejects it with `1-1-19-9`, the code `fill_time_series` already raises for
intervals of more than one frequency.

The period is read from each time series rather than from the whole column, because two series
of the same period interleave into shorter gaps: two monthly series a fortnight apart used to
move by a day. A series of a single Data Point shows no gap, so it contradicts no period and
takes the one the rest of the Data Set agrees on.

| Operand | Before | Now |
| --- | --- | --- |
| daily series + monthly series | both moved by a day | `1-1-19-9` |
| two monthly series, interleaved | both moved by a day | both move by a month |
| monthly series + single Data Point | single point moved by a year | moves by a month |
| `TimeInterval` of mixed durations | shifted by row 0's duration | `1-1-19-9` |

The DuckDb engine built its frequency as a scalar subquery cross-joined onto every row; it now
partitions the `LAG` by the series keys, collapses the per-series table through a
`COUNT(DISTINCT period_ind) > 1` guard that calls `error()`, and cross-joins the single period.
`_map_query_error` maps that message back to `1-1-19-9` for `timeshift`.

## Checklist

## Impact / Risk

- **Breaking**: a Data Set whose series do not all share one period is now rejected instead of
  shifted. A Data Set holding a single series, or several of the same period, is unaffected
  except where the series interleave, which now moves by the right period. Worth a changelog
  entry.
- `TimePeriod` identifiers are untouched: each value carries its own period indicator, so
  nothing is inferred and no series can disagree.
- A uniform shift within a series is injective, so the reference time Identifier stays unique.

## Notes

`test_GH_960_1` covers the interleaved series and the single-Data-Point default;
`test_GH_960_2` pins the error on the Data Set from the issue. The error is data-dependent, so
it is asserted inline with `run()` and `pytest.raises`, the way
`test_fill_time_series_interval_mixed_frequency` does — `NewSemanticExceptionTest` runs a `1-*`
code without data on the pandas engine.

---

Closes #960


javihern98 (approved):
Looks good, thanks! 😊

PR #982, opened 2026-08-06
Fix #978: a clause reads its operand as the operand comes out

### LINKED ISSUE #978: Problem with nested clause operators (DuckDb)
### Initial Checks

### Summary

This seems similar to #920. with this code we obtain this error. and we split and use an intermediate dateset works fine.

```
Execution error: Binder Error: Referenced column "CONF_STATUS" not found in FROM clause!
Candidate bindings: "CONCEPT", "COUNTRY", "OBS_VALUE", "EXTTRADE_PRODUCT_SITC", "MOVEMENT"

LINE 1: CREATE TABLE "result" AS SELECT "ADJUSTMENT", "CONCEPT"
```

[RESULT.union_all.json](https://github.com/user-attachments/files/30783318/RESULT.union_all.json)

### Reproducible Example

```python
result <- round(union_all,5)[calc UNIT_MEASURE := "EUR"];
```

### Expected vs Actual

Expected: Do not raise error

### vtlengine version

1.9.2rc3

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

A clause applied straight to an operator failed on DuckDb:

```
Binder Error: Referenced column "CONF_STATUS" not found in FROM clause!
Candidate bindings: "CONCEPT", "COUNTRY", "OBS_VALUE", ...
```

The transpiler held two disagreeing views of the same node. For `round(union_all, 5)` its structure
reported the Attributes, while the query it emitted selected the Identifiers and Measures only:

```
STRUCT round(DS,5) -> ['Id_1', 'Me_1', 'At_1']
SQL    round(DS,5) -> SELECT "Id_1", ROUND(...) AS "Me_1" FROM "DS_1"
```

`calc` enumerated the structure's Components, so it asked for a column the subquery never selected.
Splitting the statement in two hid it, because the intermediate Data Set was materialised first.

Three changes:

- `calc` carries the operand through as `t.*`, excluding only the Measures it replaces, so it never
  depends on the structure agreeing with the query.
- A lone Measure whose data type changes takes the generic `int_var` name, mirroring the `bool_var`
  handling `isnull` already had. The rule is that the rename happens only where the type changes:
  `ceil` over a `Number` gives `int_var`, `ceil` over an `Integer` keeps the name it had, `round(ds, 5)`
  keeps it and `round(ds)` does not.
- The base of a calc chain is resolved with the assignment stashed, as the other clauses already do.

Every operator family now agrees with the pandas engine: `round` and `trunc` with and without a
number of digits, `ceil` over `Number` and over `Integer`, `floor`, `abs`, `power`, `sqrt`, `upper`,
`length`, `isnull`, a plain Data Set, and a calc overwriting an existing Measure.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- Fixes a failure; no script that worked before changes its result. The Attributes were already
  absent from these operators' output on both engines.
- The projection of a calc changes shape, so `test_calc_clause_new_column` was updated: it asserted
  the SQL listed the Components literally, which is exactly the detail that had to change.
- `calc` output puts a replaced Measure last rather than in its original position. Components are
  matched by name, and the full suite is green on both engines.

---
Closes #978


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #986, opened 2026-08-06
Fix #981: ceil, floor, round and trunc give back an Integer on DuckDb

### LINKED ISSUE #981: Ceil and floor return a float on DuckDb where the Measure is declared Integer
### Initial Checks

### Summary

`ceil` and `floor` over a `Number` Measure produce an `Integer` Measure, named `int_var`. On DuckDb
the values come back as floats even though the Component is declared `Integer`, while the pandas
engine returns integers.

This is the same class as #935, where `flow_to_stock` widened an `Integer` Measure to a float: the
declared Component type and the values disagree.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r <- ceil(DS_1);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

data_df = pd.DataFrame({"Id_1": ["a", "b"], "Me_1": [1.2, 2.7]})

run_result = run(
    script=script,
    datapoints=datapoints,
    use_duckdb=True,
)

print(run_result["DS_r"].data.dtypes)
```

### Expected vs Actual

Expected: `int_var` holds 2 and 3, as an integer, matching the declared `Integer` Component.

Actual: DuckDb returns 2.0 and 3.0 with dtype `float64`; the pandas engine returns 2 and 3 with
dtype `int64[pyarrow]`.

| engine | int_var | dtype |
| --- | --- | --- |
| pandas | 2, 3 | `int64[pyarrow]` |
| DuckDb | 2.0, 3.0 | `float64` |

### vtlengine version

1.9.X

### Python version

3.12.3

### OS

_No response_

### Additional Context

`CEIL` and `FLOOR` in DuckDb return the type they are given, so a `DOUBLE` Measure stays a `DOUBLE`.
The result should be cast to the declared `Integer` type, as #935 did for the accumulated Measure of
`flow_to_stock`.

Reproduces on its own, without any clause after it, so it is independent of #978.

albertohernandez1995:
## Summary

`ceil` and `floor` over a `Number` Measure give an `Integer` Measure, but DuckDb returned floats
while the Component said `Integer`:

| engine | int_var | dtype |
| --- | --- | --- |
| pandas | 2, 3 | `int64[pyarrow]` |
| DuckDb | 2.0, 3.0 | `float64` |

`CEIL` and `FLOOR` in DuckDb return the type they are given, so a `DOUBLE` Measure stays a
`DOUBLE`. Both are now cast back to the declared type, as #935 did for `flow_to_stock`.

The same defect covered two operators the issue did not name: **`round` and `trunc` without a
number of digits** also give an `Integer` and were also returning floats. With digits they keep
giving a `Number` and are untouched.

| | declared | before | after |
| --- | --- | --- | --- |
| `ceil(ds)` | Integer | 2.0, 3.0 | 2, 3 |
| `floor(ds)` | Integer | 1.0, 2.0 | 1, 2 |
| `round(ds)` | Integer | 1.0, 3.0 | 1, 3 |
| `trunc(ds)` | Integer | 1.0, 2.0 | 1, 2 |
| `round(ds, 2)` | Number | 1.2, 2.7 | unchanged |
| `length(ds)` | Integer | already an integer | unchanged |

`length` was already correct because `LENGTH` returns a `BIGINT`.

Making the round and trunc half work needed one more change: `visit_ParamOp` synthesised
`params_sql = ["0"]` when the call carried no digits, so by the time the generator ran it could no
longer tell the two forms apart. The generator already defaults the precision to `0`, so that
injection was redundant and has been removed.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- Values from these four operators come back as integers on DuckDb instead of floats, matching the
  declared Component and the pandas engine. Nothing that was already an integer changes.
- Checked on the dataset and the scalar forms, and on negative values, where `ceil(-3.4)` is `-3`
  and `floor(-3.4)` is `-4` on both engines.
- Four SQL-shape tests expected the previous templates and were updated. Two of them were
  parametrised on a bare function name; they now carry the whole expression, so `abs`, `exp`, `ln`
  and `sqrt` keep asserting their own shape rather than inheriting a cast.

## Notes

The test Data Set deliberately has no null. A null makes the pandas engine hold an `Integer` Measure
as `float64`, which depends on how the Data Points were loaded rather than on this operator, so a
negative value is used instead to pin the integer result without encoding that artifact.

---

Closes #981


javihern98 (approved):
Looks good, thanks! 😊

PR #989, opened 2026-08-07
Fix #983: a Time_Period scalar keeps the requested representation

### LINKED ISSUE #983: `time_period_output_format` is ignored for scalars when `output_folder` is set (pandas engine)
### Initial Checks

### Summary

With `output_folder=None` both engines honour `time_period_output_format="natural"` for time_period scalars
(`2020-01`), but when `output_folder` is set the pandas engine writes `2020-M01` to `_scalars.csv`, ignoring the
requested format; the DuckDB engine writes `2020-01` as expected.

### Reproducible Example

```python
cast_sc_str_time_period := cast("2020-M01", time_period);
```

### Expected vs Actual

Expected: have same behavior in all the cases 

### vtlengine version

1.9.2rc3

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

A `Time_Period` scalar written to `output_folder` ignored `time_period_output_format` on the pandas
engine. It applied to every format, not only `natural`, and to the returned value as well as the
file:

| format | file, before | returned, before | both, after |
| --- | --- | --- | --- |
| `natural` | `2020-M01` | `2020-M01` | `2020-01` |
| `vtl` | `2020-M01` | `2020-M01` | `2020M1` |
| `sdmx_gregorian` | `2020-M01` | `2020-M01` | `2020-01` |
| `sdmx_reporting` | `2020-M01` | `2020-M01` | `2020-M01` |

`sdmx_reporting` matched only by coincidence, being the representation that was leaking through.
The DuckDb engine was already correct in all four.

There were two causes:

- `_save_scalars_efficient` wrote `str(scalar.value)` without applying the representation, while
  `_save_datapoints_efficient` passes it to `save_datapoints`. That accounts for the file.
- The API formats results only when `output_folder is None`. Data Sets are formatted as they are
  written, but the scalar handed to the file is a `copy()`, so the object returned to the caller was
  never formatted. That accounts for the returned value.

Both now apply the representation. They act on different objects, so nothing is converted twice.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- A `Time_Period` scalar written to a folder, and the one handed back alongside it, now carry the
  requested representation on the pandas engine. Anything asking for `sdmx_reporting` is unchanged,
  since that is what was being produced.
- Scalars of other types and the Data Sets written beside them are untouched: verified with a script
  mixing a `Time_Period` scalar, an Integer, a String and a Data Set, where both engines now write
  byte-identical output.
- The DuckDb engine needed no change.

## Notes

`test_scalar_time_period_output_format_with_output_folder` covers all four formats and asserts the
file and the returned value together, on both backends. It fails three of the four cases on the
pandas engine against the previous code.

---

Closes #983


javihern98 (approved):
Looks good, thanks! 😊

PR #992, opened 2026-08-07
Fix #987: null survives a viral propagation round trip

### LINKED ISSUE #987: Problems with `null` in viral propagation rules
### Initial Checks

### Summary

With this propagation rules I see a few errors with `create_ast_with_comments()` and `ASTString()`. 
1. Else condition is lost in create_ast.
2. `null` is converted to None in create_ast, and in ASYStrings this is converted to "None".

### Reproducible Example

```python
define viral propagation At_1 (variable At_1) is
    when null then null;
    when "Ay" then null;
    when "Bq" then "Bq";
    when "Dx" then "Dx";
    when "Ko" then "Ko";
    when "Nj" then "Nj";
    when "Pz" then "Pz";
    else null
end viral propagation;
```

### Expected vs Actual

Expected: have same code after create ast and convert to string

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

### LINKED ISSUE #1001: Wrong results for analytics without `partition by` (pandas )
### Initial Checks

### Summary

When an analytic invocation has an `order by` but no `partition by`, the pandas backend
partitions by every identifier not named in the `order by`, instead of treating the whole
dataset as a single partition. When one of those identifiers is unique per row the windows
collapse to a single row, so every analytic operator silently returns the current row's
value. The DuckDB backend (`use_duckdb=True`) is not affected.

### Reproducible Example

```python
import pandas as pd
from vtlengine import run

STRUCTURES = {
    "datasets": [
        {
            "name": "DS_DEMO",
            "DataStructure": [
                {"name": "Id_ord", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Id_other", "type": "String", "role": "Identifier", "nullable": False},
                {"name": "Me_num", "type": "Number", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

DATA = pd.DataFrame(
    {
        "Id_ord": [1, 2, 3, 4, 5, 6],
        "Id_other": ["a", "b", "c", "d", "e", "f"],
        "Me_num": [5.0, 2.0, 8.0, 1.0, 9.0, 3.0],
    }
)

SCRIPT = """
DS_min   := DS_DEMO [calc m := min(Me_num over (order by Id_ord))];
DS_count := DS_DEMO [calc m := count(Me_num over (order by Id_ord))];
"""

EXPECTED = {
    "DS_min": [5.0, 2.0, 2.0, 1.0, 1.0, 1.0],   # running minimum
    "DS_count": [1, 2, 3, 4, 5, 6],             # running count
}


def evaluate(use_duckdb):
    out = run(
        script=SCRIPT,
        data_structures=STRUCTURES,
        datapoints={"DS_DEMO": DATA.copy()},
        return_only_persistent=False,
        use_duckdb=use_duckdb,
    )
    return {n: list(ds.data.sort_values("Id_ord")["m"]) for n, ds in out.items()}


pandas_result = evaluate(use_duckdb=False)
duckdb_result = evaluate(use_duckdb=True)

for name in ("DS_min", "DS_count"):
    print(name)
    print("  expected:", EXPECTED[name])
    print("  duckdb  :", duckdb_result[name])
    print("  pandas  :", pandas_result[name],
          "<-- WRONG" if pandas_result[name] != EXPECTED[name] else "")
```

### Expected vs Actual

Expected: work equal in both engines

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

# Output:

###  MIN("Me_num") OVER (ORDER BY "Id_ord" asc)

albertohernandez1995:
## Summary

A viral propagation rule using `null` did not survive `create_ast` followed by `ASTString`:

```
in:   when null then null;  ...  else null
out:  when "None" then "None";  ...        <- the else clause is gone
```

Two separate causes:

- `ASTString` wrapped every constant in quotes, so `null`, which reaches the rule as `None`, was
  rendered as the string `"None"`. Constants are now written the way they were read, with `null` as
  a keyword. Numbers and booleans are no longer quoted either, since the grammar accepts any
  `constant` in these clauses.
- The else clause was emitted only when `default_value` was set, and `else null` leaves it `None`,
  which is exactly what a rule with no else clause looks like. `ViralPropagationDef` now records
  whether one was written.

> [!NOTE]
> `resolve_single` and `resolve_group` both return `rule.default_value`, so `else null` and no else
> clause already resolve identically. The new flag exists for the round trip and changes no
> behaviour; the field carries a default, so the single place that builds the rule is unaffected.

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- `prettify` and `ASTString` output changes only for rules that use `null`, a number or a boolean,
  which previously came out quoted or were dropped. Rules built from strings are unchanged.
- No change to how a rule resolves at runtime.
- Checked over seven shapes, each idempotent and each keeping the else clause present or absent as
  written: no else, `else "ZZ"`, `else null`, `aggregate max`, a binary `when ... and ...`, a named
  rule, and a `valuedomain` target.

## Notes

`GH_987.vtl` is registered in both `params_prettier`, which compares the prettified text against a
reference, and `params`, which renders and reparses and compares the two ASTs. The second catches
the dropped else through the new flag. Both fail against the previous code.

---

Closes #987

### DIFF
--- a/src/vtlengine/AST/ASTConstructor.py
+++ b/src/vtlengine/AST/ASTConstructor.py
@@ -173,7 +173,9 @@ class ASTVisitor:
 
         propagation_name = Terminals().visitVarID(ctx_list[3]).value
         signature_type, target = self.visitVpSignature(ctx_list[5])
-        enumerated_clauses, aggregate_clause, default_value = self.visitVpBody(ctx_list[8])
+        enumerated_clauses, aggregate_clause, default_value, has_default = self.visitVpBody(
+            ctx_list[8]
+        )
 
         token_info = extract_token_info(ctx)
 
@@ -184,6 +186,7 @@ class ASTVisitor:
             enumerated_clauses=enumerated_clauses,
             aggregate_clause=aggregate_clause,
             default_value=default_value,
+            has_default=has_default,
             **token_info,
         )
 
@@ -200,6 +203,7 @@ class ASTVisitor:
         enumerated_clauses: List[EnumeratedVpClause] = []
         aggregate_clause: Optional[AggregateVpClause] = None
         default_value: Optional[str] = None
+        has_default = False
 
         for child in ctx_list:
             if child.is_terminal:
@@ -210,8 +214,9 @@ class ASTVisitor:
                 aggregate_clause = self.visitAggregationVpClause(child)
             elif child.ctx_id == RC.DEFAULT_VP_CLAUSE:
                 default_value = self.visitDefaultVpClause(child)
+                has_default = True
 
-        return enumerated_clauses, aggregate_clause, default_value
+        return enumerated_clauses, aggregate_clause, default_value, has_default
 
     def visitEnumeratedVpClause(self, ctx: Any) -> Any:
         """enumeratedVpClause: (IDENTIFIER COLON)? WHEN vpCondition THEN constant ;"""
--- a/src/vtlengine/AST/ASTString.py
+++ b/src/vtlengine/AST/ASTString.py
@@ -240,19 +240,35 @@ class ASTString(ASTTemplate):
             )
 
     # ---------------------- Viral Propagation ----------------------
+    @staticmethod
+    def _vp_constant(value: Any) -> str:
+        """A viral propagation constant written the way it was read.
+
+        ``null`` reaches the rule as None and is a keyword rather than a value, so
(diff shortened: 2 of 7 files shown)


mla2001 (approved):
Looks fine! 😊

PR #993, opened 2026-08-07
Fix #988: a nested operator names its Measures for itself

### LINKED ISSUE #988: Error to find component (DuckDb)
### Initial Checks

### Summary

We see this error when we try this code.

```
 Execution error: Binder Error: Referenced column "Me_str" not found in FROM clause!
Candidate bindings: "Va_str", "Id_date", "int_var", "Id_period"

LINE 1: ... SELECT "Id_1", "Id_2", "Id_3", "Id_date", "Id_period", LENGTH("Me_str") AS "int_var
```

[DS_1_datapoints.csv](https://github.com/user-attachments/files/30821100/DS_1_datapoints.csv)
[DS_1.json](https://github.com/user-attachments/files/30821099/DS_1.json)

### Reproducible Example

```python
ds_str := DS_1#Me_str;
trim_ds := length(trim(ds_str));
```

### Expected vs Actual

Expected: Do not fail

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

A nested operator over a Component failed on DuckDb:

```
Binder Error: Referenced column "Me_str" not found in FROM clause!
Candidate bindings: "Va_str", "Id_date", "int_var", "Id_period"
LINE 1: ... LENGTH("Me_str") AS "int_var" ...
```

`_apply_measures` names its output Measure after the assignment target. In
`length(trim(ds))` the target's Measure is `int_var`, so both operators claimed that name:
`trim` renamed `Me_str` to `int_var`, and the `length` wrapped around it still referred to
`Me_str`, which the subquery no longer produced. The candidate bindings in the error show the
inner query had already emitted `int_var`.

The operand's query is now built with `current_assignment` stashed, which is what
`_resolve_clause_dataset` already does for clause operands, so only the operator producing the
assignment's result takes the declared name and the ones underneath keep their own.

Checked across nesting depths, all matching the pandas engine:

| expression | before | after |
| --- | --- | --- |
| `ds_str := DS_1#Me_str; length(trim(ds_str))` | BinderException | `int_var` |
| `length(trim(DS_1#Me_str))` | BinderException | `int_var` |
| `length(trim(upper(DS_1#Me_str)))` | BinderException | `int_var` |
| `upper(trim(DS_1#Me_str))` | BinderException | `Me_str` |
| `length(DS_1#Me_str)` | already worked | unchanged |

## Checklist

- [x] Tests pass (`pytest`)

## Impact / Risk

- Fixes a failure; a single operator, where the target's name does apply, is unchanged. The full
  suite needed no reference updates on either engine.
- `_apply_measures` backs most operators, so the change is broad in reach but narrow in effect: it
  only stops an operand from adopting a name that belongs to the statement's result.

## Notes

This is the third place the assignment target leaked into an operand, after #920 and #978. It is
not a duplicate of either: #988's script still fails on `cr-978`, since that fix covers a clause
reading its operand rather than an operator nested in another.

`test_GH_988_1` covers the nested forms and a lone `trim`, and includes a viral attribute since
those pass through the same helper. It fails against the previous code.

---

Closes #988


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #997, opened 2026-08-07
Fix #990: having and analytic count stop corrupting viral attributes and non-numeric measures

### LINKED ISSUE #990: Having error (Pandas)
### Initial Checks

### Summary

With this code we receive this error, Seems is related with having clause.

Execution error: "['Va_str'] not in index"

### Reproducible Example

```python
c_ds_3 := count ( DS_1#Me_num group by Id_1 having count() > 2 );
```

### Expected vs Actual

Expected: Do not raise error 

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes two Pandas-engine failures reported in #990. The DuckDB engine was already correct in both cases.

**`having` over an operand with a viral attribute** — `count ( DS_1#Me_1 group by Id_1 having count() > 2 )` raised `Execution error: "['Va_str'] not in index"`. The having analysis rebuilds a placeholder Data Set and was projecting its data down to identifiers + measures while deliberately keeping viral attributes among the components, so the aggregation then selected a column that was not in the data. Behind that sat a second defect: the propagated viral values were written into the result positionally, but the having clause had already dropped groups from it, so the two no longer lined up (`Length of values (4) does not match length of index (3)`). They are now joined on the grouping keys.

**Analytic `count` over a non-numeric measure** — `DS_1 [calc n := count(Me_str over (partition by Id_2))]` raised `Invalid value '-1' for dtype string`. Before running its query the operator replaced the nulls of every measure with the numeric sentinel `-1`, regardless of the measure's data type. Removing it fixes the failure and also removes an engine divergence: the sentinel made the Pandas engine count the null values, while the DuckDB engine counts the non-null ones.

## Checklist

- [x] Tests pass (`pytest`) — DuckDB 5540 passed / 27 skipped, Pandas 4931 passed / 636 skipped

## Impact / Risk

- Behaviour change: analytic `count` on the Pandas engine no longer counts null values, so a partition whose measure is entirely null now yields `0` instead of the row count. This aligns it with the DuckDB engine, which already behaved this way. No existing test changed.

## Notes

- Regression tests `GH_990_1` and `GH_990_2` cover the two cases; both fail on the unpatched tree with the errors from the issue.
- Out of scope and not reported in the issue: dataset-level analytic count, `count(DS_1 over (partition by Id_2))`, is broken on both engines — Pandas raises a merge `ValueError` on the Date identifier and DuckDB drops every measure from the result. It needs its own issue.

---

Closes #990


mla2001 (approved):
Looks fine! 😊

PR #999, opened 2026-08-07
Fix #994: aggregate operators in an aggr clause must accept any Component role

### LINKED ISSUE #994: Uncontrolled error min/max against identifier
### Initial Checks

### Summary

This examples raise `IndexError: list index out of range`. Both engines

### Reproducible Example

```python
DS_bug_1_id_int    := DS_1 [aggr m := min(Id_1) group by Id_2];
DS_bug_2_id_str    := DS_1 [aggr m := min(Id_2) group by Id_1];
DS_bug_3_id_date   := DS_1 [aggr m := min(Id_date) group by Id_2];
DS_bug_4_id_period := DS_1 [aggr m := min(Id_period) group by Id_2];
```

### Expected vs Actual

Expected: works

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes #994. `min` and `max` inside an `aggr` clause raised `IndexError: list index out of range` on both engines whenever the aggregated Component was not a Measure:

```
DS_1 [aggr m := min(Id_1) group by Id_2]
```

The Data Set synthesized as the operand of the aggregate operator copies the Component being aggregated, and it was copying its role along with it. For an Identifier that operand ended up with no Measures at all, so taking its first Measure blew up. The DuckDB engine failed at the same place because it runs the same semantic analysis before transpiling.

The reference manual applies the operator "to the firstOperand Component of the input Data Set", types the operand as `component :: component` with *Additional constraints: None*, and says the role of the calculated Component comes from `aggrRole`, defaulting to measure. The aggregated Component is therefore the Measure of the Data Set it aggregates, which is what it is now given. An aggregated Identifier is held under an internal name so that it also stays an Identifier of the operand and can still serve as a grouping key; that name never reaches the result, since the regular aggregation reads the aggregation result's first Measure and gives it the name declared in the clause.

The problem was wider than the four cases in the report — plain Attributes and Viral Attributes failed the same way. All of these now work and agree across both engines:

| `DS_1 [aggr m := ... group by Id_2]` | Before | After |
| --- | --- | --- |
| `min(Id_1)` (Integer Identifier) | `IndexError` | `1, 1, 1, 4` |
| `min(Id_2)` (String Identifier) | `IndexError` | `A, A, A, D` |
| `min(Id_date)` (Date Identifier) | `IndexError` | `2001-01-01, ...` |
| `min(Id_period)` (Time Period Identifier) | `IndexError` | `2000M1, ...` |
| `max(Id_1)`, `sum(Id_1)`, `avg(Id_1)` | `IndexError` / misleading `SemanticError` | `3, 3, 3, 4` / `6, 6, 6, 4` / `2, 2, 2, 4` |
| `min(At_int)` (Attribute) | `IndexError` | `10, 40, 70, 100` |
| `min(Va_str)` (Viral Attribute) | `IndexError` | `C, N, C, C` |
| `min(Id_2) group by Id_2` (the grouping key itself) | `IndexError` | `A, B, C, D` |

## Checklist

- [x] Tests pass (`pytest`) — DuckDB 5539 passed / 27 skipped, Pandas 4930 passed / 636 skipped

## Impact / Risk

- No breaking change: every expression covered here raised `IndexError` before, so nothing that used to return a result changes. The whole suite passes unchanged on both engines.

## Notes

- Regression test `GH_994_1` covers the four Identifier types from the report plus an Attribute operand, `max`, and the grouping key aggregated as `m7`, in a single script; it fails with `IndexError` on the unpatched tree.
- Also checked and unaffected: `count`, `group except`, an aggregation with no grouping clause, two Identifiers aggregated in one clause, the same Identifier aggregated twice, and an explicit `attribute` aggrRole.

---

Closes #994


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

Issue #1002, opened 2026-08-07
Analytic invocation ignores the default windowClause and computes a running aggregate over the partition


albertohernandez1995:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

When an analytic invocation has an `order by` but no window clause, both engines compute a *running* aggregate (each Data Point sees only the Data Points up to itself) instead of aggregating the whole partition.

The reference manual is explicit about the default in the Analytic invocation section:

> If the whole windowClause is omitted then the default is `data points between unbounded preceding and unbounded following`.

That means the window should span the entire partition, so every Data Point of a partition gets the same value. Both engines instead inherit the SQL default, `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`.

Both engines agree with each other, so this is not an engine divergence — it is the same wrong default in both.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = """
    DS_r <- DS_1 [calc m := sum(Me_1 over (partition by Id_1 order by Id_2))];
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Id_2", "type": "String", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Number", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

data_df = pd.DataFrame(
    {
        "Id_1": [1, 1, 1, 2, 2, 2],
        "Id_2": ["A", "B", "C", "A", "B", "C"],
        "Me_1": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0],
    }
)

datapoints = {"DS_1": data_df}

run_result = run(script=script, data_structures=data_structures, datapoints=datapoints)

print(run_result["DS_r"].data)
```

### Expected vs Actual

Actual, identical on both engines:

| Id_1 | Id_2 | Me_1 | m |
| ---- | ---- | ---- | ---- |
| 1 | A | 1.0 | 1.0 |
| 1 | B | 2.0 | 3.0 |
| 1 | C | 3.0 | 6.0 |
| 2 | A | 4.0 | 4.0 |
[...]

PR #1003, opened 2026-08-07
Fix #995: an omitted partition clause partitions by the unordered Identifiers

### LINKED ISSUE #995: Bug analytic without `partition by` min/max (Pandas)
### Initial Checks

### Summary

With this examples we have `1.1, 4.4, 7.7, 2.2, …, 10.0 `in pandas an `1.1` for all in duckDb.

### Reproducible Example

```python
DS_bug_an_max_5_sin_part := DS_1 [calc m := max(Me_num over (order by Id_2 desc, Id_1 desc))];
DS_bug_5_an_sin_part := DS_1 [calc m := min(Me_num over (order by Id_2, Id_1))];
```

### Expected vs Actual

expected: have same results

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes #995. An analytic invocation with an `order by` but no `partition by` returned different values on each engine:

```
DS_1 [calc m := min(Me_num over (order by Id_2, Id_1))]
```

The reference manual defines the default in the Analytic invocation section:

> partitionClause — clause that specifies how to partition Data Points in groups to be analysed separately. The input Data Set is partitioned according to the values of one or more Identifier Components. **If the clause is omitted, then the Data Set is partitioned by the Identifier Components that are not specified in the orderClause.**

The Pandas engine already implemented this. The DuckDB transpiler's `_resolve_partition_cols` handled the `except` and `except all` forms but resolved an omitted clause to an empty list, emitting no `PARTITION BY` and analysing the whole Data Set as a single partition. **The Pandas engine was the correct one here**, which is the opposite of what the issue title suggests.

All analytic operators share that code path, which matches the report that `median` was affected too:

| `DS_1 [calc m := ... over (order by Id_2, Id_1)]` | Pandas | DuckDB before | DuckDB after |
| --- | --- | --- | --- |
| `min` | `1.1, 2.2, ... 10.0` | `1.1` for every row | agrees with Pandas |
| `max` (desc) | `1.1, 2.2, ... 10.0` | `10.0` for every row | agrees with Pandas |
| `median` | `1.1, 2.2, ... 10.0` | `1.1, 1.65, 2.2, ...` | agrees with Pandas |
| `sum` | `1.1, 2.2, ... 10.0` | `1.1, 3.3, 6.6, ...` | agrees with Pandas |
| `avg` | `1.1, 2.2, ... 10.0` | `1.1, 1.65, 2.2, ...` | agrees with Pandas |
| `count` | `1` for every row | `1, 2, 3, ... 10` | agrees with Pandas |

## Checklist

- [x] Tests pass (`pytest`) — DuckDB 5541 passed / 27 skipped, Pandas 4932 passed / 636 skipped

## Impact / Risk

- Behaviour change on the DuckDB engine only, and only for analytic invocations that have an order clause and no partition clause. It brings that engine in line with the Pandas engine and with the reference manual.
- No existing test changed, so nothing in the suite depended on the previous single-partition behaviour.

## Notes

- Regression test `GH_995_1` covers `min`, `max` with `desc`, `median` and `sum`. The data is built so the implicit partitions hold several Data Points each, which the reporter's dataset did not — there every remaining Identifier combination was unique, so each Data Point formed its own partition. The test fails on the DuckDB engine without this change.
- Also checked and unchanged: an explicit `partition by`, `partition by` with no order clause, an invocation with neither clause, an explicit window clause, and `lag`.

> [!NOTE]
> Found while investigating and reported separately: the reference manual also states that *"if the whole windowClause is omitted then the default is data points between unbounded preceding and unbounded following"*, i.e. the whole partition. Both engines instead inherit the SQL default of `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW` and compute a running aggregate. Both engines agree with each other, so it is not a divergence and is out of scope here. Check issue #1002 

---

Closes #995


javihern98 (approved):
Looks good, thanks! 😊

javihern98 (commented):
Looks good, thanks! 😊

PR #1006, opened 2026-08-07
Fix #1002: an omitted window clause covers the whole partition

### LINKED ISSUE #1002: Analytic invocation ignores the default windowClause and computes a running aggregate over the partition

albertohernandez1995:
## Summary

Fixes #1002. An analytic invocation with an `order by` and no window clause returned a *running* aggregate on both engines instead of aggregating the whole partition.

The reference manual defines the default in the Analytic invocation section:

> If the whole windowClause is omitted then the default is `data points between unbounded preceding and unbounded following`.

That spans the entire partition, so every Data Point of a partition gets the same value. Neither engine emitted a frame when the clause was absent, so both inherited the SQL default of `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`. Both engines agreed with each other, which is why this was never caught as a divergence — it was the same wrong default in both.

Both now emit the default from one shared constant, `DEFAULT_ANALYTIC_WINDOW`, so the two engines cannot drift apart on it again:

- `Analytic.analyticfunc` in `src/vtlengine/Operators/Analytic.py`
- `SQLTranspiler._build_over_clause` in `src/vtlengine/duckdb_transpiler/Transpiler/__init__.py`

For `DS_1 [calc m := sum(Me_1 over (partition by Id_1 order by Id_2))]`:

| Id_1 | Id_2 | Me_1 | m before | m after |
| ---- | ---- | ---- | -------- | ------- |
| 1 | A | 1.0 | 1.0 | 6.0 |
| 1 | B | 2.0 | 3.0 | 6.0 |
| 1 | C | 3.0 | 6.0 | 6.0 |
| 2 | A | 4.0 | 4.0 | 15.0 |
| 2 | B | 5.0 | 9.0 | 15.0 |
| 2 | C | 6.0 | 15.0 | 15.0 |

> [!NOTE]
> `last_value` was wrong for the same reason and is fixed as a side effect. With the SQL default frame it returned the *current* Data Point; it now returns the last one of the partition, which is what makes the operator meaningful.

## Checklist

- [x] Tests pass (`pytest`) — DuckDB 5541 passed / 27 skipped, Pandas 4932 passed / 636 skipped

## Impact / Risk

> [!WARNING]
> This is a behaviour change on both engines. Any script that relied on the previous running aggregate from an analytic invocation without a window clause will now get the aggregate of the whole partition. The running behaviour is still available by writing the window clause explicitly: `data points between unbounded preceding and current data point`.

- Exactly two existing tests encoded the previous behaviour, `GH_750_8` and `GH_750_10`, and both failed identically on the two engines. Their references now hold the partition totals and their descriptions were reworded, since they described the result as a "running sum".

## Notes

- Regression test `GH_1002_1` covers `sum` and `count` with no window clause, `last_value` (which the old default silently broke), and the same `sum` with an explicit `data points between unbounded preceding and current data point` window, which must keep returning the running total.
- Also checked and unchanged: `first_value`, `rank`, `ratio_to_report` and an explicit bounded window.

---

Closes #1002


mla2001 (approved):
Looks fine! 😊

PR #1012, opened 2026-08-10
Fix #998: the analytic sum rejects non numeric Measures

### LINKED ISSUE #998: Uncontrolled error sum
### Initial Checks

### Summary

With this code brings this error .

```
Error: Exception: Execution error: Binder Error: No function matches the given name and argument types 'sum(VARCHAR)'. You might need to add explicit type casts.
	Candidate functions:
	sum(DECIMAL) -> DECIMAL
	sum(BOOLEAN) -> HUGEINT
	sum(SMALLINT) -> HUGEINT
	s
```

### Reproducible Example

```python
DS_bug_an_sum_4_string := DS_1 [calc m := sum(Me_str over (partition by Id_2))];
```

### Expected vs Actual

Expected: do not return uncontrolled error 

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

An analytic `sum` over a String Measure escaped validation and failed inside the query engine:

```
DS_1 [calc m := sum(Me_str over (partition by Id_2))]
```

```
Execution error: Binder Error: No function matches the given name and argument types 'sum(VARCHAR)'.
```

The reference manual types the operand of `sum` as `measure<number>`, so this should be rejected during semantic analysis. The analytic `Sum` operator simply declared no `type_to_check`, so the check never ran. Every sibling gets this right, which is why only `sum` was affected:

| Operator | Aggregate | Analytic before | Analytic after |
| --- | --- | --- | --- |
| `sum` | `Number` | *none* | `Number` |
| `avg`, `median`, `stddev_pop`, `stddev_samp`, `var_pop`, `var_samp` | `Number` | `Number` | unchanged |
| `max`, `min` | *none* | *none* | unchanged |

`max` and `min` correctly declare nothing, since they accept any Measure. That is also why `avg`, `median` and `stddev_pop` over a String already raised a clean `SemanticError`, and why the `aggr` and Data Set level forms of `sum` were fine — only the analytic form let it through.

Both engines now raise `SemanticError 1-1-1-1: Invalid implicit cast from String to Number`.

## Checklist

## Impact / Risk

- An uncontrolled `BinderException` becomes a `SemanticError`, so the failure is now reported during semantic analysis instead of execution.
- `sum(Me_bool over (...))` is also rejected now, where it previously ran. This aligns the analytic form with the aggregate one: `DS_1 [aggr m := sum(Me_bool) group by Id_2]` and the Data Set level `sum` already rejected Boolean with the same error. No existing test covered it.

## Notes

- Regression test `GH_998_1` asserts the `SemanticError` on both engines; it fails with the `BinderException` on the unpatched tree.
- Also verified unchanged: analytic `sum` over Integer and Number Measures, and the `avg` / `median` / `stddev_pop` error paths.

---

Closes #998


javihern98 (approved):
Looks good, thanks! 😊

Issue #1004, opened 2026-08-07
lags errors (Pandas)


Pablolgn:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

bugs 2 and 4 are shared with another operators rest are only for this one.

Output | pandas | duckdb
-- | -- | --
_1_sin_offset | ParserException: syntax error at or near ")" 💥 | NULL, NULL, NULL, 1.1, 4.4, … ✅
_2_sin_part | NULL ×10 ❌ | NULL, 3.3, 6.6, 1.1, …, 9.9 ✅
_3_id_sin_default | SemanticError: Found structure not nullable and null values 💥 | NULL, NULL, NULL, 1, 1, 1, 2, 2, 2, NULL ✅
_4_ds_viral | ValueError: merge datetime64/string 💥 | NULL, NULL, NULL, hello, ESS23, … ✅

<br class="Apple-interchange-newline">

### Reproducible Example

```python
DS_bug_an_lag_1_sin_offset := DS_1 [calc m := lag(Me_num over (partition by Id_2 order by Id_1))];

DS_bug_an_lag_2_sin_part := DS_1 [calc m := lag(Me_num, 1 over (order by Id_2, Id_1))];

DS_bug_an_lag_3_id_sin_default := DS_1 [calc m := lag(Id_1, 1 over (partition by Id_2 order by Id_1))];

define viral propagation vp_va_str (variable Va_str) is
    aggregate min
end viral propagation;

DS_bug_an_lag_4_ds_viral := lag ( DS_1[drop Me_interval], 1 over (partition by Id_2 order by Id_1) );
```

### Expected vs Actual

Expected: works in both engines

### vtlengine version

1.9.2rc4

Pablolgn:
Same error in lead

Issue #996, opened 2026-08-07
Several count errors


Pablolgn:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

I found some bugs some are already fixed in duckdb but fails in pandas.

## DuckDB fixes 5 of the 8 bugs

| Case | Pandas | DuckDB |
|---|---|---|
| `[aggr n := count() group by Id_2]` | `NA, NA, NA, NA` ❌ | `3, 3, 3, 1` ✅ |
| analytic `count(Me_int over (partition by Id_2))`, group D | `1` ❌ | `0` ✅ |
| analytic `order by` without `partition by` | all `1` ❌ | `1…10` ✅ |
| `having` + viral attribute | `KeyError: ['Va_str']` 💥 | `3, 3, 3` ✅ |
| analytic over String with nulls | `TypeError: Invalid value '-1'` 💥 | `3, 2, 2, 1` ✅ |

## Present in both engines — 3 bugs

| Case | Expected | Both |
|---|---|---|
| `count(DS_1)` | `10` | `0` |
| `count(DS_1 group by Id_1)` | `3, 3, 3, 1` | `NA, NA, NA, NA` |
| `[aggr n := count(Me_int) group by Id_2]`, all-null group | `0` | `NA` |

These three are the ones worth reporting: they are not backend-specific, they
are in the core. The first two are the same problem — `count` fails with a
dataset operand; with `#component` it works correctly on both engines.

### Reproducible Example

```python
DS_bug_1_ds_simple := count ( DS_1 );
DS_bug_2_ds_group  := count ( DS_1 group by Id_1 );
DS_bug_3_grupo_nulo := DS_1 [aggr n := count(Me_int) group by Id_2];
DS_bug_4_noarg     := DS_1 [aggr n := count() group by Id_2];
DS_bug_5_an_nulos  := DS_1 [calc n := count(Me_int over (partition by Id_2))];
DS_bug_6_an_sin_part := DS_1 [calc n := count(Me_num over (order by Id_2, Id_1))];
DS_bug_7_having    := DS_1 [aggr n := count(Me_num) group by Id_2 having count(Me_num) > 2];
DS_bug_8_an_string := DS_1 [calc n := count(Me_str over (partition by Id_2))];
```

### Expected vs Actual

[...]

Issue #959, opened 2026-08-03
Count skips Data Points that have a null Measure


albertohernandez1995:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

`count` over a Data Set does not count a Data Point when one of its Measures is null, so it
returns fewer Data Points than the operand holds. On grouped counts it can return null instead
of a number.

The reference manual says the operator "returns the number of the input Data Points". It does
not exclude nulls, and it types the operand as a plain `dataset` with "Additional Constraints:
None".

> [!IMPORTANT]
> Fixing this changes results: `count` will return larger numbers wherever the operand holds
> Data Points with null Measures, and a number where it previously returned null. It needs a
> changelog entry.

Three places implement the skipping:

| engine | location | expression |
| --- | --- | --- |
| pandas | `Operators/Aggregation.py`, `evaluate` | `result_df.dropna(subset=measure_names, how="any")` |
| pandas | `Operators/Aggregation.py`, `_agg_func` | `COUNT(<first Measure>)` |
| DuckDb | `duckdb_transpiler/Transpiler/__init__.py` | `COUNT(CASE WHEN <all Measures> IS NOT NULL THEN 1 END)`, and a separate `COUNT(CASE WHEN <any Measure> IS NOT NULL ...)` for the no-operand form |

### Reproducible Example

```python
import pandas as pd

from vtlengine import run

script = """
    DS_r <- count(DS_1 group by Id_1);
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Id_2", "type": "String", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Number", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

data_df = pd.DataFrame(
    {
        "Id_1": [1, 1, 2, 2],
        "Id_2": ["a", "b", "a", "b"],
        "Me_1": [1.0, None, 3.0, 4.0],
    }
)

datapoints = {"DS_1": data_df}

run_result = run(
    script=script,
    data_structures=data_structures,
    datapoints=datapoints,
    use_duckdb=True,
)

print(run_result["DS_r"].data)
```

[...]

PR #1013, opened 2026-08-10
Fix #1000: analytic viral attributes merge on a single Date representation

### LINKED ISSUE #1000: Error on first_value with date identifier and viral attributes (Pandas)
### Initial Checks

### Summary

This is the error with this example. Seems the problem comes from the combination of date identifier and viral attributes.

You are trying to merge on datetime64[us] and string columns for key 'Id_date'. If you wish to proceed you should use pd.concat`

### Reproducible Example

```python
DS_bug_an_firstvalue_3_ds_viral := first_value ( DS_1[drop Me_interval] over (partition by Id_2 order by Id_1) );
```

### Expected vs Actual

Expected: works equal in both engines.

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes #1000. An analytic invocation over a Data Set with a Date Identifier and viral attributes failed on the Pandas engine:

```
first_value ( DS_1[drop Me_interval] over (partition by Id_2 order by Id_1) )
```

```
You are trying to merge on datetime64[us] and string columns for key 'Id_date'.
If you wish to proceed you should use pd.concat
```

`Analytic.evaluate` re-attaches viral attributes by merging the result on the Identifiers, because the query only returns Identifiers and Measures. It took those Identifiers from `operand.data`, which still holds the values as they were loaded, while `result.data` carries them as the query returned them. For a Date Identifier there were three representations in play:

| Source | dtype of `Id_date` |
| --- | --- |
| `operand.data`, as loaded | `string` |
| the normalized frame the query actually ran on | `date64[ms][pyarrow]` |
| `result.data`, from the query | `datetime64[us]` |

The merge keys now both come from the normalized representation. The cast at the end of `evaluate` still applies the declared component dtype, so the returned structure is unchanged — `Id_date` comes back as `string` on both engines, as before.

The trigger is the combination the issue identified: dropping the viral attribute made the same expression work, and the failure was not specific to `first_value` — `last_value`, `rank`, `count` and `sum` over a Data Set failed identically.

## Checklist

## Impact / Risk

- Pandas engine only; the DuckDB engine already handled all of these. No existing test changed.
- No API changes, no structural change to results: the declared component dtypes are applied after the merge exactly as before.
- No SDMX compatibility concerns.

## Notes

- Regression test `GH_1000_1` covers a Data Set level `first_value` with a Date Identifier and a viral attribute; it fails with the reported `ValueError` on the unpatched tree.
- The viral attribute merge is shared by every analytic operator, so any operator invoked over a Data Set with a Date Identifier was affected. One of the cases reported in #1004 reaches this same line and is therefore resolved as well; that issue is otherwise untouched here and is being handled on its own branch.

> [!IMPORTANT]
> A pre-existing Data Set level defect was **masked by this crash** and now surfaces as wrong output instead of an exception. It is not caused by this change and is not addressed here: `count ( DS_1 over (partition by Id_2) )` drops every Measure on the DuckDB engine, while the Pandas engine returns a separate count per Measure. Related to the `count` work in #959 and #996.

---

Closes #1000


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #1014, opened 2026-08-10
Fix #1004: lag and lead accept an omitted offset and return a nullable result.

### LINKED ISSUE #1004: lags errors (Pandas)

albertohernandez1995:
## Summary

Fixes two of the cases reported in #1004, for both `lag` and `lead` (the issue notes "Same error in lead"):

**The offset is optional but could not be omitted.** `DS_1 [calc m := lag(Me_1 over (partition by Id_2 order by Id_1))]` raised:

```
ParserException: Parser Error: syntax error at or near ")"
```

Despite appearances this is the DuckDB SQL parser, not the VTL one: the query was assembled by joining the parameter list unconditionally, so an absent offset produced `LAG("Me_1", )`. The arguments are now appended only when they are present.

**The result was not nullable.** `DS_1 [calc m := lag(Id_1, 1 over (partition by Id_2 order by Id_1))]` raised:

```
SemanticError: Found structure not nullable and null values.
```

`lag` and `lead` have no value to shift in at the first (or last) Data Point of each partition, so their result always holds nulls. Validation was passing the operand's Component through unchanged, which keeps `nullable = False` for an Identifier, and the role setter then rejected the nulls it had just produced. The result Component is now marked nullable for these two operators, on both the Component and the Data Set paths. The declared nullability matches on both engines, so the output structures agree.

## Checklist

## Impact / Risk

- Pandas engine only for the parse failure; the DuckDB engine already handled both expressions. No existing test changed.
- The nullability change affects the declared output structure of `lag` and `lead` when the operand is a non nullable Component. It reports what the data already contained, so it removes a false rejection rather than relaxing a real constraint.

## Notes

- Regression test `GH_1004_1` covers all four combinations in one script: `lag` and `lead` without an offset, and `lag` and `lead` over a non nullable Identifier. It fails with the `ParserException` on the unpatched tree.

> [!NOTE]
> The two remaining cases of #1004 are covered by other work and are deliberately not addressed here, so the issue should stay open until they land:
> - the Data Set level case with viral attributes, reported as `ValueError: merge datetime64/string`, is the same defect as #1000 and is fixed on that branch
> - the case without a `partition by` clause depends on  #995 and #1001

---

Closes #1004


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #1015, opened 2026-08-10
Fix #1005: an analytic operator in a having clause raises a semantic error

### LINKED ISSUE #1005: Semantic error with rank
### Initial Checks

### Summary

This error is raised with this code: `'DataComponent' object has no attribute 'get_measures'`

### Reproducible Example

```python
DS_bug_an_rank_3_having := DS_1[drop Va_str][aggr m := count(Me_num) group by Id_2
                                             having rank(over (order by Id_2)) > 1];
```

### Expected vs Actual

Expected: do not fail

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes #1005. An analytic operator inside a `having` clause crashed both engines:

```
DS_1[drop Va_str][aggr m := count(Me_num) group by Id_2
                  having rank(over (order by Id_2)) > 1]
```

```
AttributeError: 'DataComponent' object has no attribute 'get_measures'
```

`visit_Analytic` had no way of knowing it was inside a `having` clause, so it evaluated the analytic and returned a Component. The having handler then treated that Component as a Data Set and asked it for its Measures.

The reference manual rules this out twice. On the condition itself:

> havingCondition — a condition (boolean expression) at component level ... The havingCondition refers to the groups specified through the groupingClause, therefore **it must invoke aggregate operators** (e.g. avg, count, max …). A correct example of havingCondition is `max(obs_value) < 1000`, while the condition `obs_value < 1000` is not a right havingCondition, because it refers to the values of single Data Points and not to the groups.

And on analytic invocations:

> **The analytic invocation cannot be nested in other Aggregate or Analytic invocations.**

It is now rejected during semantic analysis with a new error code, `1-1-3-1`, the free slot in the Analytic family:

```
At op rank: An analytic operator cannot be used in a having clause,
which must invoke aggregate operators.
```

This is not specific to `rank` — `sum(Me_num over (order by Id_2))` in a `having` clause crashed identically and is now caught by the same check. Valid aggregate conditions are unaffected.

## Checklist

## Impact / Risk

- An uncontrolled `AttributeError` becomes a `SemanticError`, reported during semantic analysis instead of at execution. Both engines were affected identically, since both run the same semantic analysis.
- Expressions that used to work are unaffected: no existing test changed, and an aggregate `having` condition behaves exactly as before.
- `docs/error_messages.rst` lists the new code automatically, as it is generated from the centralised messages at docs build time and is not tracked in the repository.

## Notes

- Regression test `GH_1005_1` asserts `1-1-3-1` on both engines; it fails with the reported `AttributeError` on the unpatched tree.

---

Closes #1005


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):
Looks good, thanks! 😊

PR #1016, opened 2026-08-10
Fix #1007: implicit promotion of two types no longer favours the left one

### LINKED ISSUE #1007: Asymmetric promotion Integer/numeric
### Initial Checks

### Summary

The nvl case is the worst: two engines, two different results, neither correct, no exception thrown. Nvl.validate also ignores promotion on its own (data_type=left.data_type).
Two verified workarounds, both correct on both engines: swapping the branches (then 2.5 else 1) or cast(1, number).
I narrowed the scope: the arithmetic operators (+, *, /) are not affected as they have their own return type logic. The bug shows up in the conditionals, which call binary_implicit_promotion directly.

  | pandas | duckdb
-- | -- | --
1a/1b | ArrowInvalid: Float value 2.500000 was truncated converting to int64 💥 | 1,3,1,3,… — rounds 2.5 → 3 ❌
1c | ArrowInvalid 💥 | 1,4,7,2,5,9,3,7,9,10 — 8.8→9, 6.6→7 ❌
1d nvl | 2 (truncs) ❌ | 3 (rounds) ❌ — correct: 2.5

### Reproducible Example

```python
DS_bug_cond_1a_if_int_num   := DS_1[drop Va_str][calc m := if Me_bool then 1 else 2.5];
DS_bug_cond_1b_case_int_num := DS_1[drop Va_str][calc m := case when Me_bool then 1 else 2.5];
DS_bug_cond_1c_if_comps     := DS_1[drop Va_str][calc m := if Me_bool then Me_int else Me_num];
DS_bug_cond_1d_nvl_int_num  := DS_1[drop Va_str][calc m := nvl(Me_int, 2.5)];
```

### Expected vs Actual

Expected: works fine

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes #1007. The implicit promotion of two types was resolved in favour of the left operand:

| Call | Before | After |
| --- | --- | --- |
| `binary_implicit_promotion(Integer, Number)` | `Integer` | `Number` |
| `binary_implicit_promotion(Number, Integer)` | `Number` | `Number` |

`Integer` and `Number` each belong to the other's implicit promotion set, so the first branch always matched and returned the left type unchanged. The same function already handled this correctly on its other path, where a subtype returns the type it promotes to — the branch is even commented "For Integer and Number". It now applies that rule when no type to check is given, which is the path the conditionals use.

That explains both workarounds reported in the issue: swapping the branches puts `Number` on the left, and `cast(1, number)` removes the asymmetry.

The conditionals take their result type straight from that promotion, so an Integer branch alongside a Number one produced an Integer result and the Number values were then forced into it, differently on each engine:

| Expression | Pandas before | DuckDB before | Both after |
| --- | --- | --- | --- |
| `if Me_bool then 1 else 2.5` | `ArrowInvalid` | `1, 3` (rounds 2.5) | `1.0, 2.5` |
| `case when Me_bool then 1 else 2.5` | `ArrowInvalid` | `1, 3` | `1.0, 2.5` |
| `if Me_bool then Me_int else Me_num` | `ArrowInvalid` | rounds 8.8 to 9 | exact values |
| `nvl(Me_int, 2.5)` | `2` (truncates) | `3` (rounds) | `2.5` |

`nvl` needed two more fixes of its own, both noted in the report:

1. `Nvl.validate` called the type validation but discarded what it returned, declaring the type of its first operand instead — in all three branches (Scalar, Component and Data Set).
2. `Nvl.evaluate` then filled from the first operand's data, keeping its `Int64` dtype, so `fillna(2.5)` truncated to `2` even once the declared type was right. The fill now happens in the promoted type.

The report is also correct that the arithmetic operators are unaffected, as they have their own return type logic: `Me_int + 2.5` already gave `Number`.

## Checklist

- [x] Tests pass (`pytest`) — DuckDB 5543 passed / 27 skipped, Pandas 4934 passed / 636 skipped

## Impact / Risk

> [!IMPORTANT]
> This changes a core function of the type system, so it reaches every operator that promotes two types without declaring its own return type. Mixing an Integer and a Number now yields a Number whichever side each is on. Operators that already declared a return type, the arithmetic ones among them, are unaffected.

- Exactly one existing test encoded the previous behaviour: `GH_921_1` declared `R3` as Integer for `case when C then 0 else Me_1 + 1`, where `Me_1` is a Number. That was the truncation this issue reports, so the reference now declares `Number`. Its values are unchanged, and `R1 := if C then 0 else 1` still yields Integer in the same test.
- Promotion stays symmetric elsewhere and nothing is over promoted: two Integers still give Integer, `Date` with `TimeInterval` still resolves to `TimeInterval` from either side, and incompatible pairs still raise on both orders.

## Notes

- Regression test `GH_1007_1` covers all four reported expressions plus `nvl(Me_int, 7)`, which must stay Integer so the fix cannot be satisfied by promoting everything. It fails with `ArrowInvalid` on the unpatched tree.
- This branch was brought up to date with `1.9.X`, which merged #995 in the meantime; the merge was clean.

---

Closes #1007


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):
Looks good, thanks! 😊

PR #1018, opened 2026-08-10
Fix #1009: an aggregation in a filter clause raises a semantic error

### LINKED ISSUE #1009: Filter error
### Initial Checks

### Summary

With this example I received this error in both engines.

` 'Aggregation' object has no attribute 'value'`

### Reproducible Example

```python
DS_bug_cl_5_filter_aggr := DS_1[filter sum(Me_num) > 1];
```

### Expected vs Actual

Expected: works

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes #1009. An aggregation used as a filter condition crashed both engines:

```
DS_1[filter sum(Me_num) > 1]
```

```
AttributeError: 'Aggregation' object has no attribute 'value'
```

The reference manual types the filter condition as a Component expression evaluated per Data Point:

> filterCondition — a condition (boolean expression) at component level, having only Components of the input Data Sets as operands, **which is evaluated for each joined Data Point** and filters them

> filterCondition :: `component<boolean>`

An aggregation returns a value per group, not per Data Point, so it cannot be a filter condition. This is the mirror image of the having clause, which the manual says *must* invoke aggregate operators.

The interpreter did not reject it. The aggregation was resolved as if it were an aggregate clause and came back as a Data Set, and the filter branch then tried to read a Component name off the condition through `child.left.value`, which an `Aggregation` node does not have.

The `calc` clause already guarded exactly this case one line earlier, raising `1-2-14` when an operand returns as a Data Set — `DS_1[calc m := sum(Me_num)]` has always produced a clean `SemanticError`. `filter` was simply missing from that condition, so it now shares the same guard:

```
DS_1[filter sum(Me_num) > 1]
  -->  SemanticError 1-2-14: At op filter: Cannot perform an aggregation inside this clause.
```

Not specific to `sum`: `count` and `max` failed identically and are covered by the same check.

## Checklist

- [x] Tests pass (`pytest`) — DuckDB 5543 passed / 27 skipped, Pandas 4934 passed / 636 skipped

## Impact / Risk

- An uncontrolled `AttributeError` becomes a `SemanticError`, reported during semantic analysis instead of at execution. Both engines were affected identically, as both run the same semantic analysis.
- The message of `1-2-14` was reworded from "Cannot perform aggregation inside a Calc" to "Cannot perform an aggregation inside this clause", since it already carries `At op {op}` and now serves both clauses. Nothing depends on the wording: the only test referencing the code asserts the code itself, and `docs/error_messages.rst` is generated at docs build time.
- Analytic operators in a filter are unaffected and keep working, which is correct since they produce a value per Data Point. Verified alongside a plain condition.

## Notes

- Regression test `GH_1009_1` asserts `1-2-14` on both engines; it fails with the reported `AttributeError` on the unpatched tree.

---

Closes #1009


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #1019, opened 2026-08-10
Fix #1010: a calculated Identifier is checked on its data, not on the declared nullability of its expression

### LINKED ISSUE #1010: Error with calc identifier
### Initial Checks

### Summary

The message assume that we have null identifiers when the calc is never null. Is a semantic error.

 `('Found structure not nullable and null values. Please check transformation with output Dataset tst.DS_bug_cl_6_calc_id_nullable', '1-1-1-16')`

### Reproducible Example

```python
DS_bug_cl_6_calc_id_nullable := DS_1[calc identifier i := Me_num * 2];
```

### Expected vs Actual

Expected: Didn't brings error.

### vtlengine version

1.9.2rc4

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

Fixes #1010. A calculated Identifier was rejected on the **declared** nullability of the expression it is calculated from:

```
DS_1[calc identifier i := Me_num * 2]
  -->  SemanticError 1-1-1-16: Found structure not nullable and null values
```

`Me_num` is declared nullable, so `Me_num * 2` is too, and a calculated Identifier could therefore never be derived from a Measure regardless of the data. Deriving one from an Identifier or from a constant worked, since those are declared not nullable.

An Identifier constrains its **values**, not the declaration of the expression that produces them, and validation has no data to decide that with. The result is now always declared not nullable, and the null check is made on the data. Both engines agree in both directions:

| | no nulls in the data | nulls in the data | semantic analysis only |
| --- | --- | --- | --- |
| Pandas | works | `SemanticError 1-1-1-16` | accepted |
| DuckDB | works | `SemanticError 1-1-1-16` | accepted |

Semantic analysis accepting it is deliberate: without data it cannot know, so the decision belongs at execution.

> [!IMPORTANT]
> Moving the check to the data made the DuckDB engine silently produce a NULL Identifier for expressions that do contain nulls, since it never had a data level check of its own. The transpiler now emits the same check through the existing `error('VTL ...')` convention and it is mapped back to `1-1-1-16`. Without that, the two engines would have disagreed on invalid input.

## Checklist

- [ ] Tests pass (`pytest`) — **five existing tests fail pending the decision below**

## Impact / Risk

> [!WARNING]
> This changes established behaviour. Five tests currently assert the opposite, and they split in two groups once each is run against its own data:
>
> | Test | Expression | With its own data |
> | --- | --- | --- |
> | `GL_196_4` | `if ... then true else null` | still rejected, the data does produce nulls |
> | `GL_196_6` | `if not isnull(X) then Y else X` | still rejected, the data does produce nulls |
> | `GL_218_22` | `A + B` | accepted, the data has no nulls |
> | `JoinsGeneralTests::test_2` | `d1#Me_1 + d2#Me_11` | accepted, the data has no nulls |
> | `JoinsGeneralTests::test_3` | same shape | accepted, the data has no nulls |
>
> The first two are still caught, and fail only because `NewSemanticExceptionTest` runs the Pandas engine in semantic only mode for a `1-*` code, where there is no data. They need to go through `run()`.
>
> The other three assert the rejection of expressions that never produce a null, which is the false positive this issue reports. `JoinsGeneralTests::test_2` already records the doubt in its own docstring: *"Status: BUG ... so should fail. There is a discrepancy between semantic and Base. In this case the semantic is wrong"*.

- The five tests were left untouched, since inverting three of them from expecting an error to expecting a result is a change to established behaviour rather than part of a bug fix. Deciding that is what this PR needs before it can go green.

## Notes

- Regression tests: `GH_1010_1` covers the reported expression over data without nulls and fails on the unpatched tree; `GH_1010_2_null_identifier_rejected` covers an expression that does produce nulls and is written against `run()`, since semantic analysis has no data and no longer decides it.
- `_map_query_error` exceeded the complexity limit once the new mapping was added, so the macro mappings that carry their own VTL code were extracted into `_map_vtl_macro_error`.

Closes #1010


albertohernandez1995:
Heads up before this is reviewed: the fix works and the quality gate is clean, but it changes behaviour that five existing tests currently assert the opposite of, so their expectations need a decision before this can go green.

**What the fix does.** A calculated Identifier was rejected on the *declared* nullability of the expression it comes from. `Me_num` is declared nullable, so `Me_num * 2` is too, and `calc identifier i := Me_num * 2` could never work no matter what the data contained. An Identifier constrains its values, not the declaration of the expression that produces them, so the result is now always declared not nullable and the null check is made on the data. Both engines agree in both directions:

| | no nulls in the data | nulls in the data | semantic analysis only |
| --- | --- | --- | --- |
| Pandas | works | `SemanticError 1-1-1-16` | accepted |
| DuckDB | works | `SemanticError 1-1-1-16` | accepted |

Semantic analysis accepting it is deliberate: without data it cannot know, so the decision belongs at execution.

> [!IMPORTANT]
> Moving the check to the data made the DuckDB engine silently produce a NULL Identifier for expressions that do contain nulls, because it never had a data level check of its own. The transpiler now emits the same check through the `error('VTL ...')` convention, mapped back to `1-1-1-16`. Without that, the two engines would have disagreed on invalid input.

**The conflict.** Five tests fail, and they split in two groups once each is run against its own data:

| Test | Expression | With its own data |
| --- | --- | --- |
| `GL_196_4` | `if ... then true else null` | still rejected, the data does produce nulls |
| `GL_196_6` | `if not isnull(X) then Y else X` | still rejected, the data does produce nulls |
| `GL_218_22` | `A + B` | accepted, the data has no nulls |
| `JoinsGeneralTests::test_2` | `d1#Me_1 + d2#Me_11` | accepted, the data has no nulls |
| `JoinsGeneralTests::test_3` | same shape | accepted, the data has no nulls |

The first two are still caught. They fail only because `NewSemanticExceptionTest` runs the Pandas engine in semantic only mode for a `1-*` code, where there is no data to judge on, so they need to go through `run()` instead.

The other three assert the rejection of expressions that never produce a null, which is the false positive this issue reports. `GL_218_22` is a single row with no nulls in it. `JoinsGeneralTests::test_2` already records the doubt in its own docstring:

> Status: BUG ... the components Me_1, Me_11 provided from DS_1 and DS_2 has "isNull": true, so should fail. ***There is a discrepancy between semantic and Base. In this case the semantic is wrong***

**What is needed to land this.** Either:

1. Accept the new semantics: `GL_196_4` and `GL_196_6` move to `run()` based checks, and `GL_218_22` plus the two join tests flip from expecting an error to expecting a result, with references.
2. Keep the current rule and close #1010 as intended behaviour, using the note in `JoinsGeneralTests::test_2` as the place where it was already questioned.

I did not rewrite the five tests, since inverting three of them from error to success is a change to established behaviour rather than part of a bug fix.

src/vtlengine/Operators/RoleSetter.py
```diff
@@ -52,8 +52,7 @@ class Identifier(RoleSetter):
     @classmethod
     def validate(cls, operand: ALLOWED_MODEL_TYPES, data_size: int = 0) -> DataComponent:
         result = super().validate(operand)
-        if result.nullable:
-            raise SemanticError("1-1-1-16")
+        result.nullable = False
         return result
```
javihern98, line 59: The solution to this is wrong. We must keep in the validate the check for result.nullable as it must be False from the input

javihern98 (changes requested):

src/vtlengine/Operators/RoleSetter.py
albertohernandez1995: Restoring that check reverts #1010 — the expression fails again on both engines. Refining nullability from data breaks parity, since DuckDB validates without data. Hence the check lives on the data.

javihern98: Need to differentiate between validate and evaluate. The change in evaluate is needed, this one in the validate goes against the standard

albertohernandez1995: The split is already there: `validate` sets the result not nullable, `evaluate` rejects actual nulls. What I removed is the rejection on *declared* nullability — `evaluate` calls `validate`, so restoring it rejects clean data on both engines and fails four tests, including this PR's own.

javihern98: Discussed in the meeting:

- Keep the check of result.nullable and raise a SemanticError there
- See below comment to change the evaluate method

src/vtlengine/Operators/RoleSetter.py
```diff
@@ -62,6 +61,12 @@ def evaluate(  # type: ignore[override]
     ) -> DataComponent:
         if isinstance(operand, Scalar) and operand.value is None:
             raise SemanticError("1-1-1-16")
+        if (
+            isinstance(operand, DataComponent)
+            and operand.data is not None
+            and operand.data.isnull().any()
+        ):
+            raise SemanticError("1-1-1-16")
```
javihern98: This must be changed to a RuntimeError, with a code starting with 2

src/vtlengine/duckdb_transpiler/io/_execution.py
```diff
@@ -65,16 +65,11 @@ def _map_time_agg_error(msg: str, msg_lower: str) -> RunTimeError:
...
-    - Division by zero → RunTimeError 2-1-3-1
-    - Cast errors → SemanticError 1-1-5-1
-    """
-    msg = str(error)
-    msg_lower = msg.lower()
+def _map_vtl_macro_error(msg: str, msg_lower: str) -> Optional[Exception]:
+    """Map the errors raised by the VTL SQL macros, which carry their own code."""
+    # calc identifier: the calculated expression produced a null
+    if "vtl 1-1-1-16" in msg_lower:
+        return SemanticError("1-1-1-16")
```
javihern98: This must be a RunTimeError with code starting with 2

javihern98 (changes requested):

src/vtlengine/Operators/RoleSetter.py
albertohernandez1995: Solved

src/vtlengine/Operators/RoleSetter.py
albertohernandez1995: Solved

src/vtlengine/duckdb_transpiler/io/_execution.py
albertohernandez1995: Solved

javihern98 (approved):
Looks good now, thanks! 😊

PR #1039, opened 2026-08-12
Fix #1027: a Component named twice in a keep or drop clause

### LINKED ISSUE #1027: keep with a repeated produce inconsistent (pandas)
### Initial Checks

### Summary

[keep Me_num, Me_num] makes the pandas backend duplicate the column. Two consequences:

The returned Dataset violates its own model: components declares 2 entries, data carries 3 columns.
With output_folder, the duplicate reaches the CSV. vtlengine's own loader then rejects that file with InputValidationException 0-1-2-3, so the engine emits output it cannot read back. In a chained pipeline the failure surfaces far from its cause.
The DuckDB backend deduplicates and produces a clean 2-column dataset that reloads fine.

### Reproducible Example

```python
import pathlib, tempfile
import pandas as pd
from vtlengine import run

STRUCTURE = {"datasets": [{"name": "DS_1", "DataStructure": [
    {"name": "Id_1", "type": "Integer", "nullable": False, "role": "Identifier"},
    {"name": "Me_num", "type": "Number", "nullable": True, "role": "Measure"}]}]}
DATA = {"DS_1": pd.DataFrame({"Id_1": [1, 2, 3], "Me_num": [1.1, 2.2, 3.3]})}
SCRIPT = "DS_r := DS_1[keep Me_num, Me_num];"

# 1. The returned Dataset declares 2 Components but its data carries 3 columns.
ds = run(script=SCRIPT, data_structures=STRUCTURE, datapoints=DATA,
         return_only_persistent=False)["DS_r"]
print("components:", list(ds.components))    # ['Id_1', 'Me_num']
print("columns   :", list(ds.data.columns))  # ['Id_1', 'Me_num', 'Me_num']  <-- mismatch

# 2. The CSV written to output_folder cannot be read back by vtlengine itself.
out = pathlib.Path(tempfile.mkdtemp())
run(script=SCRIPT, data_structures=STRUCTURE, datapoints=DATA,
    return_only_persistent=False, output_folder=str(out))
print("header    :", (out / "DS_r.csv").read_text().splitlines()[0])  # Id_1,Me_num,Me_num

RELOAD = {"datasets": [{"name": "DS_r", "DataStructure": STRUCTURE["datasets"][0]["DataStructure"]}]}
run(script="DS_final := DS_r[calc x := Me_num * 2];", data_structures=RELOAD,
    datapoints={"DS_r": str(out / "DS_r.csv")}, return_only_persistent=False)
# InputValidationException: ("Columns 'Me_num' is/are duplicated.", '0-1-2-3')
```

### Expected vs Actual

Either behaviour would be fine, as long as the invariant len(data.columns) == len(components) holds:

deduplicate, matching the DuckDB backend — keep Me_num, Me_num ≡ keep Me_num; or
reject the repeated name in semantic_analysis with a SemanticError, in both backends.

### vtlengine version

1.9.2rc5

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`[keep Me_num, Me_num]` made the pandas engine take the column twice, so the Data Set declared
two Components while its data carried three columns, and the CSV written to `output_folder`
could not be read back by the engine's own loader (`0-1-2-3`). The issue allowed either
deduplicating or rejecting; this rejects, because a repeated name in a `rename` clause is
already a `1-1-6-9`, so `keep` and `drop` now follow the rule that was there rather than a
second one.

| Expression | Before | Now |
| --- | --- | --- |
| `DS_1[keep Me_num, Me_num]` | pandas: 2 Components, 3 columns. DuckDb deduplicated | `1-1-6-9` |
| `DS_1[drop Me_num, Me_num]` | pandas deduplicated. DuckDb: `ParserException: Duplicate entry "Me_num" in EXCLUDE list` | `1-1-6-9` |
| `left_join(… keep Me_1, Me_1)` | both deduplicated silently | `1-1-6-9` |
| `left_join(… drop Me_1, Me_1)` | pandas deduplicated. DuckDb: the same parser error | `1-1-6-9` |

`Keep.validate` and `Drop.validate` share a `_check_repeated` guard. A join clause flattened its
operands through `list(set(...))`, so a repeat never reached the clause that rejects it; it now
flattens without deduplicating, which also stops the component order depending on set hashing.

## Checklist

## Impact / Risk

- **Breaking** for a script that names a Component twice in a `keep` or `drop`. That script had
  no defined meaning: it produced a Data Set the engine could not read back on pandas, and a
  parser error on DuckDb for `drop`. The fix is to remove the repeat.
- Nothing changes for clauses whose names are distinct.
- Needs a release note.

## Notes

`GL_67_Fail.vtl` listed `IMPRMNT_STTS` twice inside one `keep` of sixty components, so the new
check fired before the `1-1-6-10` that fixture pins. The duplicate is removed from the script,
which keeps the fixture testing what it was written for.

`GH_1027_1` to `GH_1027_4` cover the plain `keep` and `drop` and their join forms.

Closes #1027


javihern98 (approved):
Looks good, thanks! 😊

PR #1040, opened 2026-08-12
Fix #1028: unpivot types its Measure as the one its Measures fit in

### LINKED ISSUE #1028: unpivot over Measures of different types silently corrupts data on the ( DuckDB )
### Initial Checks

### Summary

When unpivot stacks Measures of different types into a single Measure, semantic_analysis types the new Measure as Integer — it takes the type of one source Measure instead of a type that fits all of them. Number values then no longer fit, and the two backends diverge:

pandas aborts with a raw ArrowInvalid that names neither the operator nor the components involved.
DuckDB casts to Integer and rounds, with no warning: Me_num 1.1 → 1, 7.7 → 8.
The DuckDB path is the dangerous one — the run succeeds and the numbers are wrong.

Note the spec's Additional constraints for unpivot: "All the measures of op must be defined on the same Value Domain." So rejecting mixed-type input would also be a valid resolution. What should not happen is accepting it and rounding.

The order of the Measures makes no difference: [keep Me_int, Me_num] and [keep Me_num, Me_int] both yield Integer, so the type does not come from the source order.

### Reproducible Example

```python
import pandas as pd
from vtlengine import run, semantic_analysis

STRUCTURE = {"datasets": [{"name": "DS_1", "DataStructure": [
    {"name": "Id_1",   "type": "Integer", "nullable": False, "role": "Identifier"},
    {"name": "Me_int", "type": "Integer", "nullable": True,  "role": "Measure"},
    {"name": "Me_num", "type": "Number",  "nullable": True,  "role": "Measure"}]}]}
DATA = {"DS_1": pd.DataFrame({"Id_1": [1, 2], "Me_int": [1, 2], "Me_num": [1.1, 7.7]})}
SCRIPT = "DS_r := DS_1[unpivot Id_x, Me_v];"

comp = semantic_analysis(script=SCRIPT, data_structures=STRUCTURE)["DS_r"].components["Me_v"]
print("Me_v declared type:", comp.data_type.__name__)   # Integer  <-- should be Number

    try:
        df = run(script=SCRIPT, data_structures=STRUCTURE, datapoints=DATA,
                 return_only_persistent=False, use_duckdb=use_duckdb)["DS_r"].data
        print("duckdb" if use_duckdb else "pandas",
              df.sort_values(["Id_1", "Id_x"])[["Id_1", "Id_x", "Me_v"]].to_dict("records"))
    except Exception as e:
        print("duckdb" if use_duckdb else "pandas", f"{type(e).__name__}: {e}")
```

### Expected vs Actual

Either resolution is fine, as long as no value is silently altered and both backends agree:

type Me_v as the common supertype of the source Measures — Number here, since Integer is a subtype of Number — giving 1, 1.1, 2, 7.7; or
reject the mixed-type operand in semantic_analysis with a SemanticError, per the same Value Domain constraint.

### vtlengine version

1.9.2rc2

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`unpivot` stacks the Measures of its operand into a single one, and typed that Measure by
folding with `unary_implicit_promotion`, which answers whether an operand may be used where a
type is expected and returns the operand's own type when the two are in a subtype relation. So
`unary_implicit_promotion(Integer, Number)` gives `Integer`, and the result took the type of
whichever Measure came first:

| Structure order | Before | Now |
| --- | --- | --- |
| `Me_int` Integer, `Me_num` Number | `Me_v` typed Integer: pandas raised `ArrowInvalid`, DuckDb rounded `1.1` to `1` and `7.7` to `8` | `Me_v` typed Number, values `1.0, 1.1, 2.0, 7.7` |
| `Me_num` Number, `Me_int` Integer | `Me_v` typed Number | unchanged |

The fold now uses `binary_implicit_promotion`, the common supertype the engine already applies
to `if`, `nvl` and arithmetic. Measures that fit in no common type keep raising `1-1-1-1`, now
from the promotion itself, which retires a bare
`ValueError("All measures must have the same data type on unpivot clause")` that could not be
reached.

| Measures | Result |
| --- | --- |
| Integer + Number, in any order and any number of Measures | `Number` |
| Integer + Integer | `Integer` |
| String + Number, Boolean + Integer | `SemanticError 1-1-1-1` |

## Checklist

## Impact / Risk

- A Data Set whose Measures mix Integer and Number now unpivots to a Number Measure. On DuckDb
  that changes results, which were rounded before; on pandas the same script raised
  `ArrowInvalid`.
- Measures of one type are unaffected, and incompatible Measures are rejected as before.
- Needs a release note.

## Notes

The issue reports the type as independent of the order of the Measures. It is not: declaring the
Number Measure first already gave the right type, which is why `GH_1028_1` pins both orders.

The other resolution the issue offers, rejecting any difference in type per *"All the measures of
op must be defined on the same Value Domain"*, would refuse an ordinary Data Set holding an
Integer count beside a Number amount, while VTL makes Integer a subtype of Number and the engine
promotes across that boundary everywhere else.

Closes #1028


javihern98 (approved):
Looks good, thanks! 😊

PR #1041, opened 2026-08-12
Fix #1029: random requires an Integer index

### LINKED ISSUE #1029: Random accepts number as index (DuckDb)
### Initial Checks

### Summary

Following the docs random only can accept integer as index, in pandas give error in duckdb is executed.

`'float' object cannot be interpreted as an integer`

### Reproducible Example

```python
random_lit_4 :=  random ( 3.14159 ,  0.3 );
```

### Expected vs Actual

Expected: ahve same behavior, return error

### vtlengine version

1.9.2rc5

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

The index of `random` counts the numbers of the sequence to skip, so the documentation types it
as an Integer. `Random.validate` checked it with `binary_implicit_promotion`, which answers which
type two types have in common, so a Number index came back as `Number` and was accepted. The
pandas engine then called `range()` on it and raised a bare
`TypeError: 'float' object cannot be interpreted as an integer`, while the DuckDb engine, which
has no such call, ran the expression and returned a number.

| Expression | Before | Now |
| --- | --- | --- |
| `random(3.14159, 0.3)` | pandas `TypeError`, DuckDb returned a number | `1-1-1-1`, *Invalid implicit cast from Number to Integer* |
| `random(Me_int, Me_num)` | `AttributeError: 'DataComponent' object has no attribute 'value'` on both | `1-1-1-1` |
| `random(Me_num, Me_int)` | the same `AttributeError` | works on both |
| `random(3.14159, 2)` | worked | unchanged |

Two defects in the same check went with it, because the reproduction runs through both:

- The value was read off the index whatever it was, so a Component index raised an
  `AttributeError` before the operator was reached, a valid Integer Component included. Only a
  Scalar carries a value to read.
- A negative index reported the Scalar object on the pandas engine and the literal without its
  minus on DuckDb, so `random(3.14159, -1)` read `Value 1 could not be negative`. Both now report
  `Value -1`.

## Checklist

## Impact / Risk

- **Breaking on DuckDb**: a script passing a Number index ran and now fails semantic analysis. It
  failed on the pandas engine already, with an error naming neither the operator nor the operand.
- An Integer index is unaffected, and a Component index works where it used to crash.
- Needs a release note.

## Notes

Worth a separate issue: the two backends return different numbers for the same seed and index —
`random(3.14159, 2)` gives `0.670288` on pandas and `0.421471` on DuckDb — because the pandas
engine seeds Python's Mersenne Twister and skips `index` draws while DuckDb uses a hash
expression. Both are conformant, since the manual does not fix the sequence, but a script's
result depends on the backend and no fixture using `random` can be backend-independent.

Closes #1029


javihern98 (approved):
Looks good, thanks! 😊

PR #1044, opened 2026-08-13
Fix #1036: a time operator over a null Scalar reports no value

### LINKED ISSUE #1036: Error when pass empty scalar getYear (DuckDb)
### Initial Checks

### Summary

We receive this error if we have scalar with empty value.

### Reproducible Example

```python
gy_sc := getyear(sc_date);
```

### Expected vs Actual

Expected: Have same value that we have in pandas, return empty sclar.

### vtlengine version

1.9.2rc5

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

albertohernandez1995:
## Summary

`getyear(sc_date)` over a Scalar holding no value failed on the DuckDb engine with
`BinderException: Could not choose a best candidate function for the function call "year(NULL)"`,
because a null Scalar was written as a bare `NULL` and an untyped NULL matches no overload. The
literal now carries the type the Scalar declares, so the SQL reads
`year(CAST(NULL AS TIMESTAMP))`.

Running the same shape through the neighbouring time operators showed the same defect pointing
the other way, with the pandas engine raising where DuckDb was right:

| Expression over a null Scalar | pandas | DuckDb | Now |
| --- | --- | --- | --- |
| `getyear`, `getmonth`, `dayofmonth`, `dayofyear` | `None` | `BinderException` | `None` |
| `period_indicator` | `ValueError: invalid literal for int() with base 10: 'None'` | `None` | `None` |
| `time_agg("A", …)` | `TypeError: argument of type 'NoneType' is not iterable` | `None` | `None` |

Both pandas cases read the absent value as the string `"None"` and parsed it. They now check for
the null first, as the Component paths of the same operators already did through
`na_action="ignore"`.

`length`, `abs`, `upper`, `not`, arithmetic, comparison and `isnull` over a null Scalar were
already consistent, and stay so.

## Checklist

## Impact / Risk

- Scripts that failed now run: nothing that worked changes its result.
- Every null literal the DuckDb engine writes from a typed value is now a typed NULL, which
  changes the generated SQL wherever a Scalar or a Constant carries a type. A literal with no
  type keeps the bare `NULL`.
- Needs a release note.

## Notes

`test_time_operator_over_null_scalar` sits in the time operators' own test file, which already
carries the `scalar_values` machinery, and runs on the selected backend.

Worth a separate issue: `nvl` over a null Scalar returns null rather than the default on the
pandas engine — `nvl(sc_int, 5)` gives `None` there and `5` on DuckDb.

---

Closes #1036


javihern98 (approved):
Looks good, thanks! 😊

Issue #1052, opened 2026-08-14
SDMX-CSV files with custom or optional columns fail to load with error 0-3-1-15


javihern98:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

The SDMX-CSV field guides explicitly allow implementers to append custom columns that are not defined in the Data Structure:

- SDMX-CSV 1.0 ([field guide](https://github.com/sdmx-twg/sdmx-csv/blob/v1.0/data-message/docs/sdmx-csv-field-guide.md), for SDMX 2.1): *"Implementers have the possibility to add any other custom columns as required, e.g. serieskey, sender, prepared, etc."*
- SDMX-CSV 2.x ([field guide](https://github.com/sdmx-twg/sdmx-csv/blob/master/data-message/docs/sdmx-csv-field-guide.md), for SDMX 3.x): *"Implementers have the possibility to add any other custom columns as required, e.g. updated, prepared, etc."* Custom columns appear after the standard columns (structure identification, action, dimensions, measures, attributes).

When reading SDMX-CSV datapoint files, the engine only strips the standard SDMX columns (`DATAFLOW`, `STRUCTURE`, `STRUCTURE_ID`, `ACTION`). Any custom column survives sanitization and is then rejected by the strict extra-columns check with `DataLoadError` `0-3-1-15`, so spec-compliant SDMX-CSV files cannot be loaded. The standard optional columns of SDMX-CSV 2.x (`STRUCTURE_NAME`, `SERIES_KEYS`, `OBS_KEYS`) are rejected the same way, even though pysdmx reads the same file correctly (see Additional Context).

Instead of sanitizing SDMX-CSV by hand, files whose first header column is `DATAFLOW` (1.0) or `STRUCTURE` (2.x) should be read with `pysdmx.io.get_datasets`, exactly as SDMX-ML and SDMX-JSON datapoint files already are. After reading, any column not defined in the DataStructure must be deleted instead of raising an error: custom columns are valid SDMX-CSV content with no VTL meaning, and pysdmx intentionally keeps them in `PandasDataset.data` (it only drops the standard SDMX columns listed above).

[...]

PR #1054, opened 2026-08-17
Fix #1053: the DuckDb engine rejects columns outside the DataStructure

### LINKED ISSUE #1053: DuckDB backend silently ignores columns not defined in the DataStructure instead of raising 0-3-1-15
### Initial Checks

### Summary

#646 (PR #647) introduced `DataLoadError` `0-3-1-15`, raised when input data contains columns that are not defined in the DataStructure, and #649 (PR #650) moved the check into `_validate_pandas` so it covers all loading paths. However, the check only exists in the pandas loading path: with `use_duckdb=True` none of the DuckDB loaders enforce it, so unknown columns are silently ignored.

As a result, the same script with the same datapoints raises `0-3-1-15` on the pandas backend and succeeds on the DuckDB backend. Both backends should agree; since the strict check is the intended behavior (#646), the DuckDB backend should raise `0-3-1-15` as well.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = """
    DS_r <- DS_1 * 10;
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

# Me_2 is not defined in the DataStructure
data_df = pd.DataFrame({"Id_1": [1, 2], "Me_1": [10.0, 20.0], "Me_2": ["A", "B"]})

    try:
        result = run(
            script=script,
            datapoints={"DS_1": data_df.copy()},
            use_duckdb=use_duckdb,
        )
        print(f"use_duckdb={use_duckdb}:")
        print(result["DS_r"].data)
    except Exception as e:
```

### Expected vs Actual

**Expected:** both backends reject the input with `DataLoadError` `0-3-1-15`, since `Me_2` is not defined in the DataStructure.

**Actual:** only the pandas backend raises; the DuckDB backend silently drops `Me_2` and executes the script:

```
use_duckdb=False: DataLoadError: ('On Dataset DS_1 loading: Found columns not defined in the DataStructure: Me_2. Please check loaded file', '0-3-1-15')
use_duckdb=True:
   Id_1   Me_1
0     1  100.0
1     2  200.0
```

### vtlengine version

1.9.2

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

**Observed behavior for every input kind carrying a column not defined in the DataStructure** (vtlengine 1.9.2):

| Input | pandas backend | DuckDB backend |
| ----- | -------------- | -------------- |
| pandas DataFrame | `DataLoadError` 0-3-1-15 | Loads, column ignored |
| Plain CSV file | `DataLoadError` 0-3-1-15 | Loads, column ignored |
| Parquet file | Not supported (fails with `0-1-1-8` trying to parse it as CSV) | Loads, column ignored |

**Affected code (DuckDB backend, `src/vtlengine/duckdb_transpiler/io/_io.py`):**

- DataFrames: `extract_datapoint_paths` collects them explicitly "without pandas validation" and `register_dataframes` inserts them selecting only the DataStructure components (`build_dataframe_select_columns`), so extra columns are never checked.
- CSV files: `load_datapoints_duckdb` builds its `INSERT ... SELECT` from the DataStructure components only (`build_select_columns` in `_validation.py`); columns outside the DataStructure are read but never selected.
- Parquet files: `_load_parquet` follows the same component-driven pattern.

The check itself lives only in the pandas path: `_check_extra_columns`, called from `_validate_pandas` (`src/vtlengine/files/parser/__init__.py`).

> [!NOTE]
> Scope boundary with #1052: for inputs identified as SDMX (SDMX-CSV files, SDMX-ML/SDMX-JSON, `run_sdmx()` datasets), columns outside the DataStructure are spec-legal and must be **dropped** on both backends. This issue covers only plain inputs (DataFrames, plain CSV, Parquet), where raising `0-3-1-15` is the intended behavior on both backends.

> [!NOTE]
> The Parquet row on the pandas backend is included only to complete the picture: Parquet input is effectively DuckDB-only today (the pandas loader tries to read it as CSV and fails with `0-1-1-8`), so once the DuckDB loader enforces the check, Parquet inputs with extra columns should raise `0-3-1-15` too.

albertohernandez1995:
## Summary

`0-3-1-15` was raised only on the pandas loading path, so with `use_duckdb=True` a column the
DataStructure does not define was silently dropped and the script ran. The DuckDb loaders build
their `INSERT ... SELECT` from the components, so anything else in the input was read and never
selected.

| Input | pandas | DuckDb before | DuckDb now |
| --- | --- | --- | --- |
| DataFrame | `0-3-1-15` | loads, column ignored | `0-3-1-15` |
| Plain CSV | `0-3-1-15` | loads, column ignored | `0-3-1-15` |
| Parquet | `0-1-1-8`, the loader reads it as a CSV | loads, column ignored | `0-3-1-15` |

One helper, `check_extra_columns`, is called from the three paths that take input:
`load_datapoints_duckdb` for CSV, `_load_parquet`, and `register_dataframes`.

## Checklist

## Impact / Risk

- **Breaking on DuckDb**: input carrying a column outside the DataStructure now fails to load
  where it used to run. It already failed on the pandas engine, so this closes a divergence
  rather than opening one, and Parquet input, which is DuckDb-only today, is covered by the same
  rule.
- Clean input is unaffected on either engine.
- Needs a release note.

## Notes

The check runs after `handle_sdmx_columns`, which takes out `DATAFLOW`, `STRUCTURE`,
`STRUCTURE_ID` and `ACTION` — the same four columns `_sanitize_sdmx_columns` removes on the
pandas path before its own check. SDMX-ML and SDMX-JSON files reach both loaders through
`load_sdmx_datapoints` already sanitised, so they pass as before and the 159 SDMX tests stay
green. Dropping the remaining columns for inputs identified as SDMX is #1052, and belongs on both
engines at once.

Closes #1053

### DIFF
--- a/src/vtlengine/duckdb_transpiler/io/_io.py
+++ b/src/vtlengine/duckdb_transpiler/io/_io.py
@@ -18,6 +18,7 @@ from vtlengine.duckdb_transpiler.io._validation import (
     build_create_table_sql,
     build_csv_column_types,
     build_select_columns,
+    check_extra_columns,
     check_missing_identifiers,
     get_column_sql_type,
     handle_sdmx_columns,
@@ -260,8 +261,9 @@ def load_datapoints_duckdb(
         # 4. Handle SDMX-CSV special columns
         keep_columns = handle_sdmx_columns(csv_columns, components)
 
-        # Check required identifier columns exist
+        # Check required identifier columns exist, and that no other column is left
         check_missing_identifiers(id_columns, keep_columns, file_path)
+        check_extra_columns(keep_columns, components, dataset_name)
 
         # 5. Build column type mapping and SELECT expressions
         csv_dtypes = build_csv_column_types(components, keep_columns)
@@ -350,6 +352,7 @@ def _load_parquet(
 
         keep_columns = handle_sdmx_columns(parquet_cols, components)
         check_missing_identifiers(id_columns, keep_columns, file_path)
+        check_extra_columns(keep_columns, components, dataset_name)
 
         select_exprs = _build_dataframe_select_columns(components, df_columns=parquet_cols)
 
@@ -624,6 +627,10 @@ def register_dataframes(
             continue
 
         components = input_datasets[name].components
+
+        # A DataFrame carries its columns as they are, so the SDMX markers are taken
+        # out before what is left is checked against the DataStructure.
+        check_extra_columns(handle_sdmx_columns(list(df.columns), components), components, name)
 
         # Detect Date columns that contain time values → TIMESTAMP instead of DATE
         type_overrides = _detect_date_type_overrides(df, components)
--- a/src/vtlengine/duckdb_transpiler/io/_validation.py
+++ b/src/vtlengine/duckdb_transpiler/io/_validation.py
@@ -482,6 +482,21 @@ def build_select_columns(
     return select_cols
 
 
+def check_extra_columns(
+    columns: List[str],
+    components: Dict[str, Component],
+    dataset_name: str,
+) -> None:
+    """Reject the columns the DataStructure does not define.
+
+    The SDMX marker columns are taken out by ``handle_sdmx_columns`` before this
+    runs, so what is left is what the pandas loader rejects too.
+    """
+    extra_columns = sorted(set(columns) - set(components))
+    if extra_columns:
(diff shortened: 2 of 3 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #1078, opened 2026-08-25
Fix #1061: the DuckDb engine reports a datapoint input it cannot resolve

### LINKED ISSUE #1061: DuckDb engine loads an empty dataset instead of raising when the datapoints input cannot be resolved
### Initial Checks

### Summary

Four datapoint inputs that the pandas engine either rejects or resolves end up as an **empty dataset with no error** on the DuckDb engine, so the script runs to completion and produces empty results:

| Input | Where |
| --- | --- |
| Path that does not exist | `duckdb_transpiler/io/_io.py:225-226` — `if not file_path.exists(): return _create_empty_table(...)` |
| Directory containing `DS_1.csv` | `extract_datapoint_paths` has no `is_dir()` branch, so the directory stem matches no dataset |
| List entry whose stem matches no dataset | the list branch only keeps `if name in input_datasets`, silently dropping the rest |
| Dict value of `None` | treated as "no datapoints for this dataset" and skipped |

On the pandas engine the equivalent checks live in `API/_InternalApi.py` (existence check, `Not found dataset ...`, "all Paths or all DataFrames").

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}


def show(label, datapoints):
        try:
            result = run(
                script=script,
                datapoints=datapoints,
                use_duckdb=use_duckdb,
            )
        except Exception as e:


    folder = Path(tmp) / "input"
    folder.mkdir()
    (folder / "DS_1.csv").write_text("Id_1,Me_1\n1,10\n")
    unknown = Path(tmp) / "UNKNOWN.csv"
    unknown.write_text("Id_1,Me_1\n1,10\n")

    show("missing file", {"DS_1": Path(tmp) / "does_not_exist.csv"})
    show("directory   ", folder)
    show("unknown stem", [unknown])
    show("dict None   ", {"DS_1": None})
```

### Expected vs Actual

**Actual:**

```
missing file | use_duckdb=False: DataLoadError: ('/tmp/.../does_not_exist.csv file not found. Please verify that the file exists and the provided path is correct. Please check loaded file', '0-3-1-1')
missing file | use_duckdb=True: loads []
directory    | use_duckdb=False: loads [{'Id_1': 1, 'Me_1': 10.0}]
directory    | use_duckdb=True: loads []
unknown stem | use_duckdb=False: InputValidationException: Not found dataset UNKNOWN in datastructures.
unknown stem | use_duckdb=True: loads []
dict None    | use_duckdb=False: InputValidationException: Invalid datapoints. All values in the dictionary must be Paths, or all values must be Pandas Dataframes.
dict None    | use_duckdb=True: loads []
```

**Expected:** the DuckDb engine behaves like the pandas engine — `0-3-1-1` for a missing file, every CSV in a directory loaded, `Not found dataset UNKNOWN` for an unmatched stem, and a validation error for a `None` value.

> [!WARNING]
> This is silent data loss: a run whose input never arrived returns an empty result instead of failing, so downstream statistical output changes with no signal.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

albertohernandez1995:
## Summary

Four datapoint inputs ended as an empty dataset with no error on the DuckDb engine, so a script
whose input never arrived ran to completion over nothing. Each is now resolved the way the pandas
engine resolves it:

| Input | pandas | DuckDb before | DuckDb now |
| --- | --- | --- | --- |
| path that does not exist | `0-3-1-1` | loads `[]` | `0-3-1-1` |
| directory holding `DS_1.csv` | loads it | loads `[]` | loads it |
| list entry whose stem names no dataset | `Not found dataset UNKNOWN` | dropped silently | `Not found dataset UNKNOWN` |
| dict value `None` | `All values in the dictionary must be Paths…` | skipped | same message |

`extract_datapoint_paths` resolves every form through the same two steps: `_datapoint_files`
rejects a path that does not exist and expands a directory into the datapoint files in it, and
`add_path` reports the dataset a file names when no structure defines it.
`load_datapoints_duckdb` no longer turns a missing path into an empty table, which left
`validate_input_path`, already present and already raising `0-3-1-1`, unreachable.

A directory as a dictionary value stays an error, as it is on the pandas engine: a dictionary
names one file per dataset. Directories are read for `.csv`, `.xml` and `.parquet`, the set the
pandas engine reads plus the format only this engine can.

Checked over sixteen forms — dict, list, bare path and string, a directory as each of them, an
empty directory, and a value that is neither a Path nor a DataFrame — all agreeing across the two
engines.

## Checklist

## Impact / Risk

- **Breaking on DuckDb**, and deliberately so: a run whose input could not be resolved now fails
  instead of returning empty results. Every case already failed, or resolved, on the pandas
  engine.
- A `None` value in the datapoints dictionary is no longer accepted as "this dataset has no
  data". A dataset that carries no datapoints is left out of the dictionary, which is what the
  pandas engine has always required.
- Needs a release note.

## Notes

Two test expectations moved with the change, both having pinned the behaviour reported here.
`tests/Helper.py` mapped a semantic-only dataset to `None`, which 110 tests relied on; it now
leaves the dataset out. `test_parquet_nonexistent_path_empty_table` asserted that a missing
Parquet path yields an empty table, and now expects `0-3-1-1`, as the CSV path does.

Closes #1061


javihern98 (approved):
Looks good, thanks! 😊

PR #1079, opened 2026-08-25
Fix #1062: the DuckDb engine reads a Time_Period value as the file wrote it

### LINKED ISSUE #1062: DuckDb engine accepts malformed Time_Period values: '2024-Q' loads NULL, '2024X15' becomes '2024D15', '2024A1' becomes '2024'
### Initial Checks

### Summary

The `vtl_period_normalize` macro (`duckdb_transpiler/sql/init.sql`) normalizes whatever it is given instead of rejecting what is not a valid Time_Period, so malformed values load as something else on the DuckDb engine while the pandas engine raises `0-3-1-6`:

| Value | DuckDb result | Why |
| --- | --- | --- |
| `2024-Q` | `NULL` | `'2024-Q' \|\| TRY_CAST('' AS INT)` concatenates to NULL |
| `2024X15` | `2024D15` | every indicator that is not A/S/Q/M/W falls into the daily branch |
| `2024A1` | `2024` | the period number is dropped |
| `2024q1` | `2024Q1` | the macro uppercases; the pandas regexes are case-sensitive |

The first three are data alteration without any signal. The last one is a straight accept/reject divergence — whichever engine is right, they should agree.

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Me_1", "type": "Time_Period", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

for value in ("2024-Q", "2024X15", "2024A1", "2024q1"):
            try:
                result = run(
                    script=script,
                    use_duckdb=use_duckdb,
                )
                loaded = result["DS_r"].data["Me_1"].tolist()
                print(f"{value:>8} | use_duckdb={use_duckdb}: loads {loaded}")
            except Exception as e:
                print(f"{value:>8} | use_duckdb={use_duckdb}: {type(e).__name__}: {e}")
```

### Expected vs Actual

**Actual** (pandas errors abbreviated — all four are `DataLoadError ... 0-3-1-6`, "is not in a valid format"):

```
  2024-Q | use_duckdb=False: DataLoadError 0-3-1-6
  2024-Q | use_duckdb=True: loads [<NA>]
 2024X15 | use_duckdb=False: DataLoadError 0-3-1-6
 2024X15 | use_duckdb=True: loads ['2024D15']
  2024A1 | use_duckdb=False: DataLoadError 0-3-1-6
  2024A1 | use_duckdb=True: loads ['2024']
  2024q1 | use_duckdb=False: DataLoadError 0-3-1-6
  2024q1 | use_duckdb=True: loads ['2024Q1']
```

**Expected:** the DuckDb loader rejects a value it cannot normalize into a canonical Time_Period, with the same `0-3-1-6` the pandas loader raises, instead of producing NULL or a different period.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

Found while auditing datapoint-loading parity between the two engines on `1.9.X`. Out-of-range period *numbers* (`2024M13`, `2023D366`) are a separate hole in the same loader and are reported separately.

albertohernandez1995:
## Summary

`vtl_period_normalize` rewrites whatever it is given, and `_validate_loaded_table` ran it
**before** the regex check — so the check read the macro's output instead of the file's content
and confirmed a value that was never there. Malformed Time_Period values loaded as a different
period, while the pandas loader raises `0-3-1-6`:

| Value | Before | Cause | Now |
| --- | --- | --- | --- |
| `2024-Q` | `NULL` | checked after normalizing | `0-3-1-6` |
| `2024X15` | `2024D15` | checked after normalizing | `0-3-1-6` |
| `2024A1` | `2024` | `^\d{4}A\d?$` allowed the number | `0-3-1-6` |
| `2024q1` | `2024Q1` | the check ran `UPPER(...)` | `0-3-1-6` |

Three changes: the temporal check runs on the raw values before normalization, it no longer
upper cases them, and the annual forms take no stray period number (`^\d{4}A$`, `^\d{4}-A1$`), as
the pandas regexes read them. The duplicate check stays after normalization, so `2024-Q1` and
`2024Q1` are still read as one Data Point.

Two divergences of the same kind went with them: `2024-A`, the hyphenated annual form without its
number, and `2024-01-01t00:00:00/2024-12-31T00:00:00`, a Time interval with a lower case `t`,
which passed only through the `UPPER(...)` the three temporal types shared.

## Checklist

## Impact / Risk

- **Breaking on DuckDb**, and deliberately so: a file holding a malformed Time_Period now fails to
  load instead of loading a different period. Every one of those values already failed on the
  pandas engine.
- Every accepted input format still loads and still reaches the same canonical period, on both
  engines.
- A Time_Period or Time value written in lower case is now rejected, as the pandas loader rejects
  it.
- Needs a release note.

## Notes

`GH_1062_1` holds the eleven accepted input formats with their canonical periods; `GH_1062_2` to
`GH_1062_6` hold one malformed value each.

The error tests of this suite read their inputs through `load_error`, which loads them with the
pandas engine, so on the DuckDb backend they reported nothing about it: the five new cases passed
with this fix reverted. `test_errors` now runs the code under the engine being tested, the way
`test_case_ds` in the same file already did, and `GL_440_2` gains that coverage with them.

Out-of-range period numbers (`2024M25`, `2023D400`) are a separate hole in the same loader and are
reported separately; the digit ranges are untouched here.

Closes #1062


mla2001 (approved):
Looks fine! 😊

PR #1080, opened 2026-08-25
Fix #1063: bound the period number a Time_Period carries on DuckDb

### LINKED ISSUE #1063: Out-of-range Time_Period numbers pass the DuckDb load and are written verbatim to the output CSV
### Initial Checks

### Summary

The DuckDb loader validates the *shape* of a Time_Period but never its *range*: the regex in `duckdb_transpiler/io/_validation.py` accepts `M99`, `W99` and `D999`, and no SQL check bounds the period number. With `output_folder` set, those values are written straight to the output CSV and the run reports success.

The in-memory path does catch them later (`RunTimeError 2-1-19-7/9` at representation time), so the bug only surfaces — and silently — when results go to files.

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Me_1", "type": "Time_Period", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

for value in ("2024M13", "2024W60", "2023D366"):
            out = Path(tmp) / "out"
            out.mkdir()
            try:
                run(
                    script=script,
                    output_folder=out,
                    use_duckdb=use_duckdb,
                )
                written = (out / "DS_r.csv").read_text().strip().replace("\n", " / ")
                print(f"{value:>9} | use_duckdb={use_duckdb}: DS_r.csv = {written}")
            except Exception as e:
                print(f"{value:>9} | use_duckdb={use_duckdb}: {type(e).__name__}: {e}")
```

### Expected vs Actual

**Actual:**

```
  2024M13 | use_duckdb=False: RunTimeError: ('Period Number must be between 1 and 12 for period indicator M. ...', '2-1-19-7')
  2024M13 | use_duckdb=True: DS_r.csv = Id_1,Me_1 / 1,2024M13
  2024W60 | use_duckdb=False: DataLoadError: ('... Time period '2024W60' is not in a valid format ...', '0-3-1-6')
  2024W60 | use_duckdb=True: DS_r.csv = Id_1,Me_1 / 1,2024W60
 2023D366 | use_duckdb=False: RunTimeError: ('Invalid day 366 for year 2023. ...', '2-1-19-9')
 2023D366 | use_duckdb=True: DS_r.csv = Id_1,Me_1 / 1,2023D366
```

**Expected:** the DuckDb loader rejects an out-of-range period number at load time (M > 12, S > 2, Q > 4, W > 53, D beyond the days in the year), so file output and in-memory output agree and neither writes an invalid period.

> [!WARNING]
> With `output_folder` the run succeeds and the invalid periods land in the output file, which is then read back as valid input by whatever consumes it.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

Found while auditing datapoint-loading parity between the two engines on `1.9.X`. Malformed period *shapes* (`2024-Q`, `2024X15`, `2024A1`) are a separate defect in the same loader and are reported separately.

albertohernandez1995:
## Summary

The DuckDb loader validated the *shape* of a Time_Period and never its *range*: the regex
accepted `M99`, `W99` and `D999`, and no check bounded the period number. With `output_folder`
set, those values were written straight to the output CSV and the run reported success, so an
invalid period was read back as valid input by whatever consumed the file.

| Value | Before | Now |
| --- | --- | --- |
| `2024M13`, `2024M25` | written to the CSV | `0-3-1-6` |
| `2024W54`, `2024W60` | written | `0-3-1-6` |
| `2023D366`, `2024D400`, `1900D366` | written | `0-3-1-6` |
| `2024D366`, `2000D366`, `2024W53`, `2024M12` | loaded | loaded, unchanged |

Two parts, both in `_validation.py`:

- The month, week and day of year are bounded in `TIME_PERIOD_PATTERN`, in the compact and the
  hyphenated form, leading zeros included. Each bound is written once, as `_MONTH`, `_WEEK` and
  `_DAY_OF_YEAR`.
- `_time_period_day_out_of_range` measures the day against `DAYOFYEAR(MAKE_DATE(year, 12, 31))`,
  which a regex cannot do: `2024D366` is a day of a leap year and `2023D366` is no day at all. It
  folds into the same single query and reports the same `0-3-1-6`. The century rule holds, so
  `2000D366` loads and `1900D366` does not.

## Checklist

## Impact / Risk

- **Breaking on DuckDb**, and deliberately so: a file holding a period number out of range now
  fails to load instead of being written to the output. The pandas engine already refused every
  one of those values.
- Every valid period still loads, the leap year included.
- Needs a release note.

## Notes

`GH_1063_1` holds the period numbers at the edge of what each indicator accepts, the leap year
included. `GH_1063_2` to `GH_1063_9` hold the eight values both engines reject as the inputs are
read.

Four values report differently, though neither engine writes them: `2024M13` and `2024W54` raise
`2-1-19-7` on the pandas engine and `2023D366` and `1900D366` raise `2-1-19-9`, all of them from
the period handler while the result is written, because its regexes accept them and leave the
range to it. They state no single code, so they are covered by a test that asserts both engines
refuse and no output file is written. Bounding those two regexes the way this PR bounds the DuckDb
one would let all four join `error_param`; the tests pinning `2-1-19-7` and `2-1-19-9` construct
`TimePeriodHandler` directly, so they would not break. It changes a public error code on the
pandas engine, so it is left out of this PR.

The error tests of this suite read their inputs through `load_error`, which loads them with the
pandas engine, so on the DuckDb backend they reported nothing about it: the eight new cases passed
with this fix reverted. `test_errors` now runs the code under the engine being tested, the way
`test_case_ds` in the same file already did, and `GL_440_2` gains that coverage with them.

Closes #1063


javihern98 (approved):
Looks good, thanks! 😊

PR #1055, opened 2026-08-17
Fix #1052: an SDMX input drops the columns outside its DataStructure

albertohernandez1995:
## Summary

The SDMX-CSV field guides let implementers append custom columns, and SDMX-CSV 2.x has optional
standard ones (`STRUCTURE_NAME`, `SERIES_KEYS`, `OBS_KEYS`). The engine stripped only
`DATAFLOW`, `STRUCTURE`, `STRUCTURE_ID` and `ACTION`, so anything else met the strict
extra-columns check and a spec-compliant file failed to load with `0-3-1-15`. The same happened
through `run_sdmx()`, where pysdmx keeps the custom columns in `PandasDataset.data`.

| Input | Columns outside the DataStructure |
| --- | --- |
| plain CSV, DataFrame, Parquet | `0-3-1-15`, unchanged |
| SDMX-CSV 1.0 and 2.x, SDMX-ML, SDMX-JSON, `run_sdmx()` | dropped |

**Every SDMX format is read by pysdmx, on both engines.** `load_datapoints` reads SDMX-CSV with
`load_sdmx_datapoints`, as it already read `.xml` and `.json`, and `extract_datapoint_paths` does
the same on the DuckDb side, where a single `_sdmx_dataframe` helper replaces the three copies of
the SDMX block that were there. The native CSV loader is left for plain files, with its
extra-columns check unconditional.

**The format is pysdmx's verdict, not a rule written here.** `is_sdmx_csv_file` asks
`process_string_to_read`, the same call `read_sdmx` makes, so a file read as SDMX-CSV here is a
file pysdmx parses. It is taken on the header line, which is all the CSV branch of that detection
reads: over the 6488 CSV files in the suite the header-line verdict agrees with reading the whole
document, on the 27 SDMX-CSV ones and on every plain one. That authority fixes two cases a
hand-written rule got wrong:

| Header | Hand-written rule | pysdmx verdict |
| --- | --- | --- |
| `STRUCTURE,Id_1,Me_1`, no `STRUCTURE_ID` | SDMX, sent to a reader that refuses it: `0-3-1-8` on pandas, silent fallback on DuckDb | plain, loaded by both |
| `Id_1,DATAFLOW,Me_1`, marker not first | plain: `0-3-1-15` on both | SDMX-CSV 1.0, loaded by both |

**What is dropped, and where.** `drop_undefined_columns` runs inside `_sanitize_sdmx_columns`,
[...]

PR #1081, opened 2026-08-25
Fix #1064: an SDMX marker column is read only where an SDMX-CSV file carries one

### LINKED ISSUE #1064: An ACTION column in a plain CSV silently deletes rows and is dropped on the DuckDb engine
### Initial Checks

### Summary

`handle_sdmx_columns` (`duckdb_transpiler/io/_validation.py:329-331`) excludes an `ACTION` column with no SDMX precondition — it does not require `STRUCTURE`/`STRUCTURE_ID` to be present — and `duckdb_transpiler/io/_io.py:288-289` then appends `WHERE "ACTION" != 'D' OR "ACTION" IS NULL`. So on any plain CSV that happens to have a column called `ACTION`, the DuckDb engine **deletes the rows marked `D`** and drops the column, while the pandas engine raises `0-3-1-15` for the undeclared column.

Because the exclusion happens *before* `check_extra_columns`, the extra-column check added in #1053 never sees `ACTION`.

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

    # A plain CSV (no STRUCTURE / STRUCTURE_ID columns), so not SDMX-CSV
    csv_path.write_text("Id_1,Me_1,ACTION\n1,10,I\n2,20,D\n")
        try:
            result = run(
                script=script,
                use_duckdb=use_duckdb,
            )
        except Exception as e:
```

### Expected vs Actual

**Actual:**

```
use_duckdb=False: DataLoadError: ('On Dataset DS_1 loading: Found columns not defined in the DataStructure: ACTION. Please check loaded file DS_r', '0-3-1-15')
use_duckdb=True: loads [{'Id_1': 1, 'Me_1': 10.0}]
```

The `Id_1 = 2` row is gone on DuckDb, with no warning.

**Expected:** SDMX marker handling only applies to an SDMX-CSV input (the pandas loader requires `STRUCTURE` first). On a plain CSV, `ACTION` is a column the DataStructure does not define, so both engines should raise `0-3-1-15`.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

Found while auditing datapoint-loading parity between the two engines on `1.9.X`. Same code area as #1053 (extra columns) and #1052 (SDMX-CSV optional columns) — whichever fix lands should keep the SDMX-CSV behaviour those cover while making the plain-CSV case an error.

albertohernandez1995:
## Summary

`handle_sdmx_columns` excluded `STRUCTURE`, `STRUCTURE_ID` and `ACTION` wherever they appeared,
with no precondition, and the loader then appended `WHERE "ACTION" != 'D' OR "ACTION" IS NULL`. On
a plain CSV holding a column called `ACTION`, the DuckDb engine **deleted the rows marked `D`** and
dropped the column, while the pandas engine reports the undefined column with `0-3-1-15`. Because
the exclusion runs before `check_extra_columns`, the check added in #1053 never saw those columns.

| Input | Before | Now |
| --- | --- | --- |
| plain CSV with `ACTION` | the row marked `D` deleted, the column dropped | `0-3-1-15` |
| plain CSV with `STRUCTURE` or `STRUCTURE_ID` | the column dropped | `0-3-1-15` |
| plain CSV with `DATAFLOW` | `0-3-1-15` | unchanged |
| `ACTION` defined by the DataStructure | loaded | loaded, both rows |
| SDMX-CSV with `ACTION=D` | the row filtered out | unchanged |

A file is SDMX-CSV when its first column names the structure its rows belong to, and only then
does it carry the columns around it, which is the condition the pandas loader reads them under.
`DATAFLOW` already required it. The `ACTION` row filter follows the column that was actually
excluded, so it can only fire for an SDMX-CSV file.

## Checklist

## Impact / Risk

- **Breaking on DuckDb**, and deliberately so: a plain CSV carrying a column named after an SDMX
  marker now fails to load instead of losing the column, and its rows, without a signal. The
  pandas engine already refused it.
- SDMX-CSV files are unaffected: the markers are still dropped and the rows `ACTION` marks deleted
  still stay out.
- A DataStructure that defines a Component named `ACTION`, `STRUCTURE` or `STRUCTURE_ID` keeps
  loading it as data.
- Needs a release note.

## Notes

`STRUCTURE_ID` in a plain CSV was dropped the same way and is covered here too, though the issue
only reports `ACTION`.

This touches the same function as #1055 (issue #1052), which reads SDMX-CSV with pysdmx. The two
are compatible — once that lands an SDMX-CSV file no longer reaches this loader, and the condition
added here simply never fires — but they conflict textually in `_validation.py` and `_io.py`.

Closes #1064


mla2001 (approved):
Looks fine! 😊

PR #1084, opened 2026-08-25
Fix #1067: the DuckDb engine stores a temporal value as it validated it

### LINKED ISSUE #1067: Duration values are validated case-insensitively but stored raw on DuckDb, so 'a' and ' A' never match 'A'
### Initial Checks

### Summary

The DuckDb loader validates a `Duration` value as `UPPER(TRIM(...))` (`duckdb_transpiler/io/_validation.py:58`, and the check around lines 373-377) but stores the original string. So `a` and `' A'` are accepted and kept verbatim, and every later comparison against the canonical `A` quietly fails. The pandas loader rejects both values outright.

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = 'DS_r <- DS_1[calc Me_2 := if Me_1 = cast("A", duration) then 1 else 0];'

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Me_1", "type": "Duration", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

    csv_path.write_text("Id_1,Me_1\n1,A\n2,a\n3, A\n")
        try:
            result = run(
                script=script,
                use_duckdb=use_duckdb,
            )
        except Exception as e:
```

### Expected vs Actual

**Actual:**

```
use_duckdb=False: DataLoadError: ('On Dataset DS_1 loading: not possible to cast column Me_1 to Duration. Error found: "Impossible to cast a from type str to Duration." ...', '0-3-1-6')
use_duckdb=True: loads [{'Id_1': 1, 'Me_1': 'A', 'Me_2': 1}, {'Id_1': 2, 'Me_1': 'a', 'Me_2': 0}, {'Id_1': 3, 'Me_1': ' A', 'Me_2': 0}]
```

Rows 2 and 3 hold the same duration as row 1 as far as the load-time validation is concerned, yet they compare as different.

**Expected:** either the loader normalizes the value it validated (store `A`, so `Me_2` is `1` for all three rows), or it rejects `a` / `' A'` like the pandas loader does. What it must not do is accept a value under one spelling and store another.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

albertohernandez1995:
## Summary

The DuckDb loader validated a temporal column as `UPPER(TRIM(value))` but stored the raw string, so ` A` passed validation as `A` and was then kept as ` A` — a value that never matches a `Duration` literal again. #1062 already fixed the case half of this; what was left was the space.

Probing showed the gap reaches all three temporal types, and that each needs a different answer, because that is what the pandas loader does: `Duration` does not strip and so refuses ` A`, while `check_time_period` and `check_time` both call `value.strip()` and accept it. Three changes bring the engines back together:

- `validate_temporal_columns` checks a `Duration` exactly as it was written, and keeps trimming the other two.
- `_normalize_time_period_columns` normalises `vtl_period_normalize(TRIM(...))`.
- a new `_normalize_time_interval_columns` trims a `Time` column, next to the period one in `_validate_loaded_table`.

## Checklist

- [x] Tests pass (`pytest`) — 5648 passed on DuckDb, 5040 on pandas

## Impact / Risk

- Not a breaking change: a file whose temporal values carry no surrounding space loads exactly as before.
- A `Duration` written with a space around it now raises `0-3-1-6` on DuckDb instead of loading a value that silently compared to nothing. That is the pandas behaviour, and it turns a wrong result into an error.
- A `Time_Period` or a `Time` written with a space around it now loads trimmed on DuckDb, matching pandas.

## Notes

- The two new tests live in `tests/DataLoad/test_dataload.py` and follow that file's code-fixture pattern; both are backend-aware through `DataLoadExceptionTest` / `BaseTest`, so they run on whichever engine the suite is under. Reverting the source makes both fail.

Closes #1067


mla2001 (approved):
Looks fine!

PR #1085, opened 2026-08-25
Fix #1068: enforce the documented Boolean set on both engines

### LINKED ISSUE #1068: Boolean and Integer coercion diverges between engines, and the pandas loader turns any unrecognized Boolean string into False
### Initial Checks

### Summary

Two problems in the same area, one of them independent of parity:

1. **`_parse_boolean` accepts anything.** `files/parser/__init__.py` maps every value that is not `true`/`1` to `False`, so `abc`, `yes` and `2` all load as `False` on the pandas engine instead of failing. Garbage becomes a valid Boolean data point.
2. **The engines coerce differently.** On the DataFrame path DuckDb applies SQL cast semantics — `yes` → `True` (VARCHAR→BOOL accepts y/n), `1.0`/`0.0` → `True`/`False`, and Python `True`/`False` in an `Integer` column → `1`/`0` — while the pandas loader gives `False`, `False`/`False`, and a `0-3-1-6` respectively.

### Reproducible Example

```python
import tempfile
from pathlib import Path

import pandas as pd

from vtlengine import run


script = "DS_r <- DS_1;"


def structures(measure_type):
    return {
        "datasets": [
            {
                "name": "DS_1",
                "DataStructure": [
                    {"name": "Me_1", "type": measure_type, "role": "Measure", "nullable": True},
                ],
            }
        ]
    }


def show(label, measure_type, datapoints):
        try:
            result = run(
                script=script,
                datapoints=datapoints,
                use_duckdb=use_duckdb,
            )
            print(f"{label} | use_duckdb={use_duckdb}: loads {result['DS_r'].data['Me_1'].tolist()}")
        except Exception as e:


    csv_path.write_text("Id_1,Me_1\n1,abc\n2,yes\n3,2\n")
    show("Boolean CSV 'abc','yes','2' ", "Boolean", {"DS_1": csv_path})

show("Boolean DataFrame 'yes'     ", "Boolean", {"DS_1": pd.DataFrame({"Id_1": [1], "Me_1": ["yes"]})})
show("Boolean DataFrame 1.0 / 0.0 ", "Boolean", {"DS_1": pd.DataFrame({"Id_1": [1, 2], "Me_1": [1.0, 0.0]})})
show("Integer DataFrame True/False", "Integer", {"DS_1": pd.DataFrame({"Id_1": [1, 2], "Me_1": [True, False]})})
```

### Expected vs Actual

**Actual:**

```
Boolean CSV 'abc','yes','2'  | use_duckdb=False: loads [False, False, False]
Boolean CSV 'abc','yes','2'  | use_duckdb=True: DataLoadError 0-3-1-6 (Could not convert string 'abc' to BOOL)
Boolean DataFrame 'yes'      | use_duckdb=False: loads [False]
Boolean DataFrame 'yes'      | use_duckdb=True: loads [True]
Boolean DataFrame 1.0 / 0.0  | use_duckdb=False: loads [False, False]
Boolean DataFrame 1.0 / 0.0  | use_duckdb=True: loads [True, False]
Integer DataFrame True/False | use_duckdb=False: DataLoadError 0-3-1-6 (could not convert string to float: 'True')
Integer DataFrame True/False | use_duckdb=True: loads [1, 0]
```

**Expected:** one documented set of accepted Boolean representations, enforced by both engines and on both the CSV and DataFrame paths; anything outside it raises `0-3-1-6` instead of silently becoming `False`. The same decision is needed for Boolean values in an `Integer` column.

> [!IMPORTANT]
> Item 1 is a correctness bug on its own: an unreadable Boolean cell currently produces a valid `False` data point on the default engine.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

Found while auditing datapoint-loading parity between the two engines on `1.9.X`. Related but older: #39 (`_parse_boolean` and Boolean inputs). The `Id_1` in the DuckDb CSV message is wrong — that misattribution is reported separately.

albertohernandez1995:
## Summary

`_parse_boolean` mapped every value it could not read to `False`, so `abc`, `yes` and `2` all loaded as a perfectly valid `False` data point on the pandas engine. The two engines also disagreed on the DataFrame path, where DuckDb applied SQL cast semantics: `yes` → `True`, `1.0`/`0.0` → `True`/`False`, and a Python `True`/`False` in an `Integer` column → `1`/`0`, against `False`, `False`/`False` and a `0-3-1-6` on pandas.

The set of accepted representations did not need deciding — the documentation and the `Boolean` type both already state it: `true`/`false` whatever the case, `1`, `0`, a real boolean, or a number compared against zero. Neither loader enforced it. This change makes both of them do so, on the CSV and the DataFrame path alike:

- `_parse_boolean` reads the value through `Boolean.check`/`Boolean.cast` and raises for anything outside the set, so the loader reports `0-3-1-6`.
- The value now reaches it as it is; stringifying it first hid a real boolean and a number behind their text, which is what made `1.0` load as `False`.
- `_cast_exact_integer` reads a boolean as `1` or `0`, the reading `Integer.check`/`Integer.cast` and the DuckDb engine already gave it.
- A new `build_boolean_cast` maps a text value by hand on the DuckDb side, where the native cast reads `yes`, `y` and `t` as booleans as well. A column that already is a boolean or a number keeps DuckDB's cast.

One test fixture wrote a null Boolean as the literal `NULL`, which the pandas loader was quietly turning into `False` and which the DuckDb engine already refused; it now uses an empty field, the documented way of writing a null in a CSV.

## Checklist

- [x] Tests pass (`pytest`) — 5657 passed on DuckDb, 5049 on pandas
- [x] Documentation updated (if applicable)

## Impact / Risk

- **Behaviour change, and the point of the fix:** a Boolean value outside the documented set now raises `0-3-1-6` on the pandas engine instead of loading as `False`. A script that was reading such a file was reading wrong data, so this turns a silent wrong answer into an error at load time. The DuckDb engine already refused most of these.
- `yes`, `y` and `t` no longer load as `True` on the DuckDb engine; they were never part of the documented set and never worked on pandas.
- A boolean in an `Integer` column now loads as `1`/`0` on the pandas engine, where it used to raise; nothing that worked before stops working.
- Data/SDMX compatibility: none. A file whose Boolean values are written as documented loads exactly as before.

## Notes

- The tests live in `tests/DataLoad/test_dataload.py` and follow both of that file's patterns: code fixtures for the CSV cases, and a class parametrized over `use_duckdb` for the DataFrame path. Reverting the source makes 5 of them fail on DuckDb and 7 on pandas — the ones that stay green are the cases the issue reports DuckDb already handled.
- The misattributed column name in the DuckDb CSV error message (`Id_1` for a defect in `Me_1`) is reported separately and is untouched here.

Closes #1068


javihern98 (commented):
Need to give evidence here that the output is exactly the same as well

albertohernandez1995:
Output measured two ways: between the engines, and against `1.9.X`. Both the values held in memory and the exact bytes of the written CSV were compared, the bytes by md5.

**Between the engines, the output is identical in every case** — same values, same bytes:

| input | values | CSV bytes (both engines) |
| --- | --- | --- |
| CSV `true`,`FALSE`,`1`,`0` | `[True, False, True, False]` | `Id_1,Me_1\n1,True\n2,False\n3,True\n4,False\n` |
| CSV `"""TRUE"""`, `"""false"""` | `[True, False]` | `Id_1,Me_1\n1,True\n2,False\n` |
| CSV with an empty value | `[True, <NA>]` | `Id_1,Me_1\n1,True\n2,\n` |
| DataFrame Python bools | `[True, False]` | `Id_1,Me_1\n1,True\n2,False\n` |
| DataFrame ints `1`/`0` | `[True, False]` | `Id_1,Me_1\n1,True\n2,False\n` |
| DataFrame floats `1.0`/`0.0` | `[True, False]` | `Id_1,Me_1\n1,True\n2,False\n` |
| DataFrame strings `True`/`0` | `[True, False]` | `Id_1,Me_1\n1,True\n2,False\n` |
| Integer column, Python bools | `[1, 0]` | `Id_1,Me_1\n1,1\n2,0\n` |

**Against `1.9.X`, only three cases change, and each is one this PR is about.** Running the same comparison with the three source files reverted to `origin/1.9.X` and diffing the output:

| case | before, pandas / duckdb | after |
| --- | --- | --- |
| quoted `"""TRUE"""` | `False` / `True` — bytes differ | both `True`, bytes equal |
| floats `1.0` / `0.0` | `False, False` / `True, False` — bytes differ | both `True, False`, bytes equal |
| Integer column holding booleans | `0-3-1-6` / `1, 0` | both `1, 0`, bytes equal |

Everything else is byte for byte what it was: `true`/`FALSE`/`1`/`0` from a CSV, an empty value, Python bools, ints and strings all produce exactly the same file as before.

A value outside the documented set now gets the same error code on both engines rather than loading as `False` on one of them: `abc`, `yes`, `2`, `y` and `t` all raise `0-3-1-6` on either engine.

> [!NOTE]
> The `"""TRUE"""` row is a value change on the pandas engine that no test was watching — `tests/DataLoad/test_dataload.py::test_32` loads that file but only asserts that loading succeeds, never what the value became. It was loading as `False`, and now reads as `True`, which is what the DuckDb engine already read it as.

javihern98 (dismissed):
Looks good after reviewing comment, thanks! 😊

javihern98 (approved):

PR #1086, opened 2026-08-25
Fix #1069: enforce declared nullability at load time on both engines

### LINKED ISSUE #1069: A non-nullable Measure accepts an empty value on the pandas loader, and DuckDb reports a missing Measure as an Identifier error
### Initial Checks

### Summary

Declared nullability is not enforced consistently for Measures at load time:

- **pandas.** An empty cell in a **non-nullable** `String` Measure loads as `None` — `read_csv` is called with `na_values=[""]` and no nullability check follows, so the declared structure is violated with no error.
- **DuckDb.** The same input raises `0-3-1-3` but blames an Identifier: `An Identifier cannot have null values, found null values on unknown` — the component is a Measure, and the name is `unknown`.
- **DuckDb, DataFrame path.** A DataFrame that simply does not carry a non-nullable Measure gets the same misleading `0-3-1-3` when it has rows, and **no error at all** when it has zero rows (the DataFrame path fills NULL unconditionally, unlike the CSV path). pandas raises the correct `0-3-1-5` "Component Me_1 is missing in Datapoints" in both cases.

### Reproducible Example

```python
import tempfile
from pathlib import Path

import pandas as pd

from vtlengine import run


script = "DS_r <- DS_1;"


def structures(measure_type):
    return {
        "datasets": [
            {
                "name": "DS_1",
                "DataStructure": [
                    {"name": "Me_1", "type": measure_type, "role": "Measure", "nullable": False},
                ],
            }
        ]
    }


def show(label, measure_type, datapoints):
        try:
            result = run(
                script=script,
                datapoints=datapoints,
                use_duckdb=use_duckdb,
            )
        except Exception as e:


    csv_path.write_text("Id_1,Me_1\n1,\n")
    show("empty value in CSV       ", "String", {"DS_1": csv_path})

show("DataFrame without Me_1   ", "Number", {"DS_1": pd.DataFrame({"Id_1": [1, 2]})})
show("same, zero rows          ", "Number", {"DS_1": pd.DataFrame({"Id_1": pd.Series(dtype="int64")})})
```

### Expected vs Actual

**Actual:**

```
empty value in CSV        | use_duckdb=False: loads [{'Id_1': 1, 'Me_1': None}]
empty value in CSV        | use_duckdb=True: DataLoadError: ('On Dataset DS_1 loading:  An Identifier cannot have null values, found null values on unknown. ...', '0-3-1-3')
DataFrame without Me_1    | use_duckdb=False: DataLoadError: ('... Component Me_1 is missing in Datapoints. ...', '0-3-1-5')
DataFrame without Me_1    | use_duckdb=True: DataLoadError: ('...  An Identifier cannot have null values, found null values on unknown. ...', '0-3-1-3')
same, zero rows           | use_duckdb=False: DataLoadError: ('... Component Me_1 is missing in Datapoints. ...', '0-3-1-5')
same, zero rows           | use_duckdb=True: loads []
```

**Expected:**

- a null in a non-nullable Measure is an error on both engines, naming the Measure;
- a DataFrame missing a non-nullable Measure raises `0-3-1-5` on both engines, with or without rows.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

albertohernandez1995:
## Summary

Declared nullability was not enforced for a Measure at load time. An empty value in a non-nullable Measure loaded as a null on the pandas engine, leaving the declared structure violated with no error; the DuckDb engine refused the same input but reported it as `An Identifier cannot have null values, found null values on unknown`, naming neither the right role nor the component. On the DataFrame path a missing non-nullable Measure got that same misleading error when the frame had rows, and no error at all when it had none, since the path fills a missing column with NULL and no constraint can fire on zero rows.

- A new error code `0-3-1-17` reports a null in a component the DataStructure declares as not nullable. `0-3-1-3` stays what it says it is: the Identifier's own error.
- The pandas loader checks every non-nullable component, whatever its role, once empty values have been read as nulls.
- The DuckDb engine reads the component out of DuckDB's `table.column` constraint message and reports it by its role, instead of scanning the Identifiers only.
- The DataFrame path reports a missing non-nullable component as missing, the way the CSV path and the pandas loader already reported it, instead of filling it with NULL.

The new check also caught a defect of the same family in `nvl`, on the engine's own grammar fixture: `nvl` declared its result as not nullable whatever the replacement was, so a null replaced by another null came out as a null in a Measure that could not hold one. The rule already applied to a Scalar replacement now covers a Data Set and a Component operand as well: the result stays nullable where the replacement is itself nullable, and is not nullable where the replacement cannot be null. The grammar reference for `nvl_ds_ds` was updated to `nullable: true` for `Me_str`, which is both what the engine now computes and what that fixture's CSV has always held; it was the only reference with that inconsistency.

## Checklist

- [x] Tests pass (`pytest`) — 5668 passed on DuckDb, 5060 on pandas

## Impact / Risk

- **Behaviour change, and the point of the fix:** a null in a non-nullable component now raises `0-3-1-17` on the pandas engine instead of loading. A script reading such a file was working on data that contradicted its own structure, so this turns a silent violation into an error at load time. Declaring the component nullable is the fix for anyone relying on it.
- New error code `0-3-1-17`; `0-3-1-3` keeps its meaning and its message unchanged.
- A DataFrame missing a non-nullable component now raises `0-3-1-5` on the DuckDb engine, with or without rows, as the pandas engine already did.
- `nvl` results now report nullable where the replacement is nullable. This changes only the declared structure, never the data.
- Data/SDMX compatibility: none. A file whose values match its declared structure loads exactly as before.

## Notes

- The loader tests live in `tests/DataLoad/test_dataload.py` in both of that file's patterns — code fixtures for the CSV cases, a class parametrized over `use_duckdb` for the DataFrame path — and the `nvl` test in `tests/Bugs/test_bugs.py` follows that file's `BaseTest` pattern, where the reference structures pin the nullability. Reverting the source fails 7 of them on each engine.

Closes #1069


javihern98 (dismissed):
Looks good, thanks! 😊

mla2001 (dismissed):
Looks fine! 😊

javihern98 (approved):

Issue #653, opened 2026-04-09
UTF-8 BOM files produce misleading errors instead of being handled transparently


javihern98:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

When loading files encoded as UTF-8 with BOM (byte order mark), the engine produces misleading errors instead of handling the BOM transparently. The BOM (`\xef\xbb\xbf`) contaminates the first column name in CSV files, causes `json.JSONDecodeError` in JSON files, and triggers confusing ANTLR parse errors in VTL script files.

Error code `0-1-2-5` ("File {file} must be encoded in utf-8 (without BOM)") is defined in `messages.py` but never raised anywhere — there is no BOM detection logic in the codebase.

Affected file loading paths:
- CSV datapoints (`_validate_csv_path`, `_pandas_load_csv`, `_detect_delimiter`)
- JSON data structures (`_load_datastructure_single`)
- JSON value domains (`_load_single_value_domain`)
- JSON external routines (`_load_single_external_routine_from_file`)
- VTL scripts (`load_vtl`)

### Reproducible Example

```python
import tempfile
from pathlib import Path

import pandas as pd

from vtlengine import run

script = """
    DS_r <- DS_1;
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Number", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

# Write a CSV file with UTF-8 BOM
with tempfile.TemporaryDirectory() as tmp:
    csv_path = Path(tmp) / "DS_1.csv"
    csv_path.write_bytes(b"\xef\xbb\xbfId_1,Me_1\n1,10\n2,20\n3,30\n")

    datapoints = {"DS_1": csv_path}
    result = run(script=script, data_structures=data_structures, datapoints=datapoints)
    print(result)
```

### Expected vs Actual

Expected: The BOM is stripped transparently and the file loads correctly, producing the same result as a non-BOM UTF-8 file.

Actual: The BOM character (`\ufeff`) is prepended to the first column name, causing a misleading error:

| File type | Misleading error |
| --- | --- |
[...]

PR #1087, opened 2026-08-25
Fix #1070: read a CSV the same way on both engines, BOM and encoding alike

### LINKED ISSUE #1070: The DuckDb loader fails on UTF-8 BOM CSVs and lets a raw UnicodeDecodeError escape for non-UTF-8 files
### Initial Checks

### Summary

Two encoding problems on the DuckDb CSV path:

1. **UTF-8 BOM.** #653 made every user-facing read use `utf-8-sig`, but the DuckDb loader re-reads the header with plain `utf-8` (`duckdb_transpiler/io/_io.py:248`, and `_detect_csv_format`), so the first column name comes back as `﻿Id_1` and the load fails with `0-1-1-8` "The following Identifiers Id_1 were not found". The same file loads fine on the pandas engine — this is the #653 behaviour regressing for the second engine.
2. **Non-UTF-8 file.** A latin-1 encoded file raises a bare `UnicodeDecodeError` that escapes the `DataLoadError` mapping entirely, instead of a catalogued error (the pandas engine reads it with replacement characters).

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Me_1", "type": "String", "role": "Measure", "nullable": True},
            ],
        }
    ]
}


def show(label, payload):
        csv_path.write_bytes(payload)
            try:
                result = run(
                    script=script,
                    use_duckdb=use_duckdb,
                )
            except Exception as e:


show("UTF-8 BOM ", b"\xef\xbb\xbfId_1,Me_1\n1,abc\n")
show("latin-1   ", "Id_1,Me_1\n1,café\n".encode("latin-1"))
```

### Expected vs Actual

**Actual:**

```
UTF-8 BOM  | use_duckdb=False: loads [{'Id_1': 1, 'Me_1': 'abc'}]
UTF-8 BOM  | use_duckdb=True: InputValidationException: ('The following Identifiers Id_1 were not found, review file DS_1.csv.', '0-1-1-8')
latin-1    | use_duckdb=False: loads [{'Id_1': 1, 'Me_1': 'café'}]   # U+FFFD replacement char in place of é
latin-1    | use_duckdb=True: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 15: invalid continuation byte
```

**Expected:** the BOM is stripped transparently on the DuckDb path too (`utf-8-sig` for reading, as #653 established), and a non-UTF-8 file produces the same catalogued error on both engines — `0-1-2-5` exists for exactly this — instead of a raw `UnicodeDecodeError`.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

Found while auditing datapoint-loading parity between the two engines on `1.9.X`. Item 1 is a regression of #653 (fixed in 1.7.0 for the pandas paths, never applied to the DuckDb loader).

albertohernandez1995:
## Summary

Two encoding problems on the DuckDb CSV path:

1. **UTF-8 BOM.** #653 made every user-facing read use `utf-8-sig`, but the DuckDb loader re-read the header as plain UTF-8 in two places, so the mark stayed on the first column name and the file failed with `0-1-1-8` for an Identifier it does carry. DuckDB's own `read_csv` strips the BOM already — the break was only in those two Python-side reads, which now read the header the way every other read reads it.
2. **A file in another encoding.** A latin-1 file let a raw `UnicodeDecodeError` escape the error mapping. DuckDB refuses such a file and offers no lenient mode, so the only way both engines can agree is for both to refuse it: the pandas loader now reads strictly and reports the error instead of loading values with the characters it could not read replaced (`café` came in as `caf<?>`), and the DuckDb path reports the same, both from its own header read and from DuckDB's "invalid unicode" error.

> [!NOTE]
> The issue says `0-1-2-5` exists for this. It did — "File {file} must be encoded in utf-8 (without BOM)" — and #653 deleted it, because from then on a BOM was read transparently. It is restored here with wording that matches what the engines do now: a BOM is read transparently, any other encoding is refused.

## Checklist

- [x] Tests pass (`pytest`) — 5661 passed on DuckDb, 5053 on pandas

## Impact / Risk

- **Behaviour change on the pandas engine:** a file that is not valid UTF-8 now raises `0-1-2-5` instead of loading with every character it could not read replaced by U+FFFD. Those values were never the ones the file held, so this turns silent corruption into an error at load time. Re-encoding the file as UTF-8 is the fix.
- A UTF-8 BOM file now loads on the DuckDb engine, where it used to fail. Nothing that worked before stops working.
- Error code `0-1-2-5` is in use again, with a message of its own; no existing code changes meaning.
- Three fixtures in the repo are not valid UTF-8 (`GL_81-19-1.csv`, `GL_81-21-1.csv`, `testencode3.csv`) and no test references any of them, so nothing depended on the old reading.

## Notes

- `test_22` was a commented-out TODO for a utf-16 file, left over because the error was "line contains NUL". It is a real test again, on a fixture named the way the helpers expect.
- `TestBOMHandling::test_bom_csv` called `run()` without `use_duckdb`, so it only ever read one engine — which is how this regression got in. It is parametrized over both now, and it is one of the tests that fails when the source is reverted.

Closes #1070


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #1088, opened 2026-08-26
Fix #1071: a DuckDb load error names the column that failed and its type

### LINKED ISSUE #1071: DuckDb cast failures at load time name the wrong column, or leave the column and type blank
### Initial Checks

### Summary

`map_duckdb_error` (`duckdb_transpiler/io/_validation.py:138-160`) attributes a cast failure by scanning the generated SQL text for the first component name it can find. The result is a `0-3-1-6` that regularly names a column that had nothing to do with the failure, or no column at all:

- a Boolean value that cannot be cast in `Me_1` is reported as `not possible to cast column Id_1 to Integer`, and the raw SQL fragment leaks into the message;
- a Date failure in `Me_1` is reported as `not possible to cast column  to ` — both the column and the type are empty.

The pandas loader names the failing column and its target type.

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"


def structures(measure_type):
    return {
        "datasets": [
            {
                "name": "DS_1",
                "DataStructure": [
                    {"name": "Me_1", "type": measure_type, "role": "Measure", "nullable": True},
                ],
            }
        ]
    }


def show(label, measure_type, csv_text):
        csv_path.write_text(csv_text)
        try:
            run(
                script=script,
                use_duckdb=True,
            )
            print(f"{label}: loaded without error")
        except Exception as e:
            print(f"{label}: {type(e).__name__}: {e}")


show("Boolean garbage in Me_1", "Boolean", "Id_1,Me_1\n1,abc\n")
show("padded Date in Me_1   ", "Date", "Id_1,Me_1\n1, 2020-01-01\n")
```

### Expected vs Actual

**Actual:**

```
Boolean garbage in Me_1: DataLoadError: ('On Dataset DS_1 loading: not possible to cast column Id_1 to Integer. Error found: "Conversion Error: Could not convert string 'abc' to BOOL

LINE 12:                     END AS "Id_1", CAST(NULLIF(REPLACE("Me_1", '"', ''), '') AS BOOLEAN) AS...
                                            ^" Please check loaded file', '0-3-1-6')
padded Date in Me_1   : DataLoadError: ('On Dataset DS_1 loading: not possible to cast column  to . Error found: "Invalid Input Error: Date  2020-01-01 is not in the correct format. Use YYYY-MM-DD or YYYY-MM-DD HH:MM:SS." Please check loaded file', '0-3-1-6')
```

**Expected:** `not possible to cast column Me_1 to Boolean` and `not possible to cast column Me_1 to Date`, without the SQL fragment.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

Found while auditing datapoint-loading parity between the two engines on `1.9.X`. This misattribution shows up in the messages of several other load-time divergences, so fixing it makes those reports readable too. A more reliable approach than scanning the SQL text would be casting per column (or mapping the DuckDb error position back to the column that produced it).

albertohernandez1995:
## Summary

`map_duckdb_error` attributed a cast failure by searching the error text for the first component name it could find. DuckDB appends the statement it was running to its message, and that statement carries every column name, so the search regularly landed on a column that had nothing to do with the failure — a Boolean value that could not be read in `Me_1` was reported as `not possible to cast column Id_1 to Integer`, with the generated SQL trailing behind it. A Date failure had the opposite problem: the engine's own check named no column, so it fell through to the generic branch and reported `not possible to cast column  to `, both empty.

Three changes attribute the failure properly:

- The statement DuckDB reports is taken off before the message is reported or read for a name. It is SQL this engine generated; it says nothing to whoever wrote the file.
- The column is resolved from the `Column X:` marker the engine's own checks write, then from what DuckDB itself named, and, where neither names it, by reading each column back on its own through the very expression the load used until one fails — the approach the issue suggests. That last step only runs on a load that already failed, and `COUNT()` reads every value without holding the column in memory.
- Both Date checks now write that marker, the way the Integer check already did, and the type reported is the one the component is declared as.

The conversion branch and the generic fallback are one path now, so anything DuckDB reports about a value is attributed the same way.

## Checklist

- [x] Tests pass (`pytest`) — 5663 passed on DuckDb, 5050 on pandas

## Impact / Risk

- Message-only change: the same loads fail and succeed as before, with the same `0-3-1-6` code. What changes is which column and type the message names.
- A load that fails may now do a little extra work before reporting, when neither the marker nor DuckDB names the column: one read per column, over a file the load was going to reject anyway.
- DuckDB's own CSV diagnostics (its sniffer option dump) are left as they are — that is DuckDB reporting on the file, not the statement this engine generated.

## Notes

- `TestRegisterDataframesPartialTime` asserts a load error never says "unknown". It was passing on `cast column  to ` — the empty-column symptom this issue reports. The DataFrame path's Date check named no column either; it writes the marker now, so the test passes on a message that actually names `Me_1`.
- The new tests live in `tests/duckdb_transpiler/test_efficient_io.py`. Two of them pin cases the old search could never get right: the second of two Boolean columns, and an out-of-range day in the second of two Date columns, where the message used to name the first Date column whatever happened. All five fail when the source is reverted.

Closes #1071


javihern98 (approved):
Looks good, thanks! 😊

javihern98 (dismissed):

javihern98 (approved):

PR #1089, opened 2026-08-26
Fix #1072: read a datapoint dictionary value as what it is, and name the engine a parquet file needs

### LINKED ISSUE #1072: Datapoints container validation diverges: DuckDb accepts a dict mixing DataFrames and Paths, and a parquet file on pandas fails with a misleading 0-1-1-8
### Initial Checks

### Summary

Two API-level input-handling divergences:

1. **Mixed dict.** `API/_InternalApi.py` rejects a `datapoints` dict that mixes DataFrames and Paths ("All values in the dictionary must be Paths, or all values must be Pandas Dataframes"), but the DuckDb path (`extract_datapoint_paths`) accepts it and loads both. Either the restriction is real and both engines enforce it, or it is not and the pandas engine should drop it.
2. **Parquet on the pandas engine.** Parquet input is DuckDb-only by design, but the pandas engine reads the parquet bytes as CSV and reports `0-1-1-8` "The following Identifiers Id_1 were not found, review file DS_1.parquet", which sends the user looking for a structure problem instead of telling them parquet needs `use_duckdb=True`.

### Reproducible Example

```python
import tempfile
from pathlib import Path

import pandas as pd

from vtlengine import run


data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        },
        {
            "name": "DS_2",
            "DataStructure": [
            ],
        },
    ]
}

    csv_path = Path(tmp) / "DS_2.csv"
    csv_path.write_text("Id_1,Me_1\n1,5\n")
    parquet_path = Path(tmp) / "DS_1.parquet"
    pd.DataFrame({"Id_1": [1], "Me_1": [10.0]}).to_parquet(parquet_path)

    for label, script, datapoints in (
        (
            "dict mixing DataFrame + Path",
            "DS_r <- DS_1 + DS_2;",
            {"DS_1": pd.DataFrame({"Id_1": [1], "Me_1": [10.0]}), "DS_2": csv_path},
        ),
        ("parquet datapoints         ", "DS_r <- DS_1;", {"DS_1": parquet_path}),
    ):
            try:
                result = run(
                    script=script,
                    datapoints=datapoints,
                    use_duckdb=use_duckdb,
                )
            except Exception as e:
```

### Expected vs Actual

**Actual:**

```
dict mixing DataFrame + Path | use_duckdb=False: InputValidationException: Invalid datapoints. All values in the dictionary must be Paths, or all values must be Pandas Dataframes.
dict mixing DataFrame + Path | use_duckdb=True: loads [{'Id_1': 1, 'Me_1': 15.0}]
parquet datapoints           | use_duckdb=False: InputValidationException: ('The following Identifiers Id_1 were not found, review file DS_1.parquet.', '0-1-1-8')
parquet datapoints           | use_duckdb=True: loads [{'Id_1': 1, 'Me_1': 10.0}]
```

**Expected:** both engines agree on whether a mixed dict is valid, and a parquet datapoint file on the pandas engine produces an error that says parquet requires the DuckDb engine.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

albertohernandez1995:
## Summary

Two divergences in how the engines read the `datapoints` container:

1. **A dictionary holding both a DataFrame and a file.** The pandas engine refused it ("All values in the dictionary must be Paths, or all values must be Pandas Dataframes") while the DuckDb engine read both. A dictionary names one dataset per key and says where that dataset's Data Points are, so there is nothing to refuse: each value is now read as what it is, on both engines. The restriction is dropped rather than added to the DuckDb engine, which would have taken away behaviour that already worked.

   The old message describes a restriction neither engine has any more, so an invalid value now gets the wording the DuckDb engine already used: `Invalid datapoint for DS_1. Must be DataFrame, Path, or string.`

2. **A Parquet file on the pandas engine.** Parquet input is read by the DuckDb engine only, but the pandas engine read the parquet bytes as a CSV and reported the Identifiers it could not find in them, which sends whoever loaded it looking for a problem in the data structure. It now says what the file needs.

> [!NOTE]
> Saying it needed one more fix. `InputValidationException` takes `message` as a named parameter and then replaced it with the catalogued text for the code, so a message given beside a code never reached anyone — 12 of the 14 `0-1-1-2` call sites pass one, the S3 check among them. A message given beside a code is appended to the catalogued text now, which is what every one of those callers wrote it for.

## Checklist

- [x] Tests pass (`pytest`) — 5664 passed on DuckDb, 5056 on pandas

## Impact / Risk

- A dictionary mixing DataFrames and files now loads on the pandas engine, where it used to raise. Nothing that worked before stops working.
- A Parquet datapoint file on the pandas engine now raises `0-1-1-2` naming the engine, instead of `0-1-1-8` naming Identifiers. Both refuse the file; only the reason given changes.
- Messages passed beside an error code are now visible. No code changes meaning, and no catalogued text changes — the caller's own sentence is added after it, so anything matching on the catalogued text still matches.
- Data/SDMX compatibility: none.

Closes #1072


javihern98 (dismissed):
Looks good, thanks! 😊

src/vtlengine/API/_InternalApi.py
```diff
@@ -192,6 +192,21 @@ def _generate_single_path_dict(
...
 
 
+def _check_datapoint_format(datapoint: Path) -> None:
+    """A Parquet file is read by the DuckDb engine only.
+
+    The pandas engine read its bytes as a CSV and reported the Identifiers it could
+    not find in them, which sends whoever loaded it looking for a problem in the
+    data structure instead of naming the engine the file needs.
+    """
+    if datapoint.suffix.lower() == ".parquet":
```
mla2001, line 205: In some lines like 257 this would raise an error even if use_duckdb is set to True, add an additional check to avoid it

mla2001 (changes requested):

src/vtlengine/API/_InternalApi.py
albertohernandez1995: Added a check to prevent that case issue.

mla2001 (dismissed):
Looks fine! 😊

javihern98 (approved):

PR #1090, opened 2026-08-26
Fix #1073: a String value keeps the double quotes it holds, on all four load paths

### LINKED ISSUE #1073: String values keep their double quotes on the DuckDb DataFrame path but are stripped on every other path
### Initial Checks

### Summary

Loading a `String` value that contains a double quote gives three identical results and one different one. Both CSV paths and the pandas DataFrame path strip `"` from String values (`files/parser/__init__.py`, and the `REPLACE(..., '"', '')` in the DuckDb CSV SELECT), but the DuckDb DataFrame registration path (`duckdb_transpiler/io/_io.py`, the `register`/`CREATE TABLE ... SELECT` branch) has no such REPLACE, so the value keeps its quote.

The same data therefore compares equal or not depending on how it was handed to the engine.

### Reproducible Example

```python
import tempfile
from pathlib import Path

import pandas as pd

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Me_1", "type": "String", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

value = 'he"llo'

    for label, datapoints in (
        ("DataFrame", {"DS_1": pd.DataFrame({"Id_1": [1], "Me_1": [value]})}),
        ("CSV      ", {"DS_1": csv_path}),
    ):
            result = run(
                script=script,
                datapoints=datapoints,
                use_duckdb=use_duckdb,
            )
            print(f"{label} | use_duckdb={use_duckdb}: loads {result['DS_r'].data['Me_1'].tolist()}")
```

### Expected vs Actual

**Actual:**

```
DataFrame | use_duckdb=False: loads ['hello']
DataFrame | use_duckdb=True: loads ['he"llo']
CSV       | use_duckdb=False: loads ['hello']
CSV       | use_duckdb=True: loads ['hello']
```

**Expected:** the four loads agree.

> [!NOTE]
> The stripping itself is worth a second look: silently deleting `"` from a String value loses data. Whatever is decided, the four paths should implement the same rule — the CSV quote character is a dialect concern, not something to remove from the value of an in-memory DataFrame.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

albertohernandez1995:
## Summary

The same String value loaded four ways gave three results and one different one: both CSV paths and the pandas DataFrame path deleted every `"` from it, while the DuckDb DataFrame path kept it, so `he"llo` compared equal or not depending on how it was handed to the engine.

There were two ways to make the four agree, and the choice was measured rather than guessed. Removing the stripping from every path and running the suites failed **one** test on the DuckDb engine and **none** on pandas — and that test is about a Boolean written as `"""TRUE"""`, not about a String. Nothing depended on taking quotes out of a String value, so the four paths agree by keeping the value instead of by spreading the loss to the fourth:

- The pandas loader stores a String value as it read it.
- The DuckDb CSV String branch stores it the same way; the Boolean branch keeps its own quote handling, which is what the `"""TRUE"""` case needs.
- The DuckDb DataFrame path is untouched — it was the one already storing the value whole.

A quote that delimits a value in a CSV is the file's dialect and both readers already take it off: `"a,b"` still loads as `a,b`.

> [!IMPORTANT]
> This also fixes a loss nobody had reported. A quote written twice inside a quoted value is one quote in that value, so a CSV holding `"say ""hi"""` means `say "hi"` — and it was loading as `say hi` on **both** engines. That case is one of the tests, and it fails on both when the source is reverted.

## Checklist

- [x] Tests pass (`pytest`) — 5666 passed on DuckDb, 5058 on pandas

## Impact / Risk

- **Behaviour change, and the point of the fix:** a String value holding a double quote now keeps it on the three paths that were deleting it. Anyone whose data contains quotes was silently losing them; results and CSV output change for exactly those values, and change to what the file said.
- Nothing is refused that was accepted before, and no error codes change.
- Boolean values written as `"""TRUE"""` load exactly as before.
- Data/SDMX compatibility: a value with no quote in it loads byte for byte as before.

Closes #1073

### DIFF
--- a/src/vtlengine/duckdb_transpiler/io/_validation.py
+++ b/src/vtlengine/duckdb_transpiler/io/_validation.py
@@ -478,8 +478,7 @@ def build_select_columns(
                     stripped = f"NULLIF({stripped}, '')"
                 select_cols.append(f'CAST({stripped} AS BOOLEAN) AS "{comp_name}"')
             elif csv_type == "VARCHAR" and comp.data_type == String:
-                # Strip double quotes from String values (match pandas loader behavior)
-                expr = f"""REPLACE("{comp_name}", '"', '')"""
+                expr = f'"{comp_name}"'
                 if comp.nullable:
                     expr = f"NULLIF({expr}, '')"
                 select_cols.append(f'{expr} AS "{comp_name}"')
--- a/src/vtlengine/files/parser/__init__.py
+++ b/src/vtlengine/files/parser/__init__.py
@@ -273,9 +273,7 @@ def _validate_pandas(
                     except ValueError:
                         raise ValueError(f"Duration values are not correct in column {comp_name}")
             else:
-                data[comp_name] = data[comp_name].map(
-                    lambda x: str(x).replace('"', ""), na_action="ignore"
-                )
+                data[comp_name] = data[comp_name].map(str, na_action="ignore")
             data[comp_name] = data[comp_name].astype(comp.data_type.dtype())  # type: ignore[call-overload]
 
     except (ValueError, InputValidationException) as e:
--- a/tests/DataLoad/test_dataload.py
+++ b/tests/DataLoad/test_dataload.py
@@ -1146,3 +1146,55 @@ class TestBOMHandling:
         ds = result["DS_r"]
         assert "Id_1" in ds.data.columns
         assert "\ufeffId_1" not in ds.data.columns
+
+
+_STRING_STRUCTURE = {
+    "datasets": [
+        {
+            "name": "DS_1",
+            "DataStructure": [
+                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
+                {"name": "Me_1", "type": "String", "role": "Measure", "nullable": True},
+            ],
+        }
+    ]
+}
+
+
+@pytest.mark.parametrize("use_duckdb", [False, True], ids=["pandas", "duckdb"])
+class TestStringKeepsItsQuotes:
+    """A double quote a String value holds is part of it, whichever of the four ways
+    the value reaches the engine (issue #1073)."""
+
+    def _load(self, datapoints, use_duckdb):
+        result = run(
+            script="DS_r <- DS_1;",
+            data_structures=_STRING_STRUCTURE,
+            datapoints=datapoints,
+            use_duckdb=use_duckdb,
+        )
+        return result["DS_r"].data["Me_1"].tolist()
+
+    def test_dataframe(self, use_duckdb):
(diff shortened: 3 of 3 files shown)


javihern98 (commented):
Need to give evidences here whether this may raise problems downstream, like when using a string operator and if it also works for single quotes, given they must be prohibited as they must be used only with reserved words

albertohernandez1995:
Evidence below, on both engines.

**Single quotes are not touched by this change, before or after.** The stripping that this PR removes was `"` only — `str(x).replace('"', "")` on the pandas loader and `REPLACE(col, '"', '')` in the DuckDb SELECT. Running the same values against `origin/1.9.X` and against this branch:

| input value | before, pandas / duckdb | after, pandas / duckdb |
| --- | --- | --- |
| `it's` | `it's` / `it's` | `it's` / `it's` |
| `he"llo` | **`hello`** / `he"llo` | `he"llo` / `he"llo` |
| `a'b"c` | **`a'bc`** / `a'b"c` | `a'b"c` / `a'b"c` |

The single quote in `a'b"c` came through untouched before this PR and comes through untouched after it. Only the double quote changes hands.

**String operators.** Every one of these gives the same result on both engines, for values holding a single quote, a double quote, or both (`he"llo`, `it's`, `"q"`, `a'b"c`):

`length`, `upper`, `substr`, `||`, `replace`, `instr`, `trim`, `filter`, `in` against a set literal, `match_characters` with a regex, and `if Me_1 = Me_2` between two components.

**Output round-trip**, which is the part that could not work before: the value is written and read back byte-identically, on both engines.

| value | file written | read back |
| --- | --- | --- |
| `he"llo` | `1,"he""llo"` | `he"llo` |
| `say "hi"` | `1,"say ""hi"""` | `say "hi"` |
| `a,b` | `1,"a,b"` | `a,b` |
| `it's` | `1,it's` | `it's` |

Both engines write the same bytes and both read them back. Before this change the read step deleted the quote, so the value could not survive a save and reload of its own output.

**Quotes in data never reach the parser or the SQL text.** A value comes from a file or a DataFrame, never from the script, and on the DuckDb path it is read by `read_csv` or from a registered frame rather than being written into the statement. Values built to break that come through unharmed and identical on both engines: `'; DROP TABLE "DS_1"; --`, `a" OR "1"="1` and `O'Brien` all load, pass through `upper()` and `||`, and give the same result on either engine.

javihern98 (approved):
Looks good now after clarifying comments, thanks! 😊

PR #1091, opened 2026-08-26
Fix #1075: an empty datapoint file is reported as empty, on both engines

### LINKED ISSUE #1075: A zero-byte CSV raises 0-1-1-6 "Duplicated records" on pandas and 0-1-1-8 on DuckDb
### Initial Checks

### Summary

An empty (zero-byte) datapoint file produces a different — and in the pandas case actively wrong — error on each engine:

- pandas raises `0-1-1-6`, whose catalog text is "Duplicated records. Combination of Identifiers are repeated." There are no records at all, let alone duplicated ones;
- DuckDb raises `0-1-1-8` "The following Identifiers Id_1 were not found", which points at the structure rather than at the empty file.

Neither says the file is empty.

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

    csv_path.write_bytes(b"")  # zero-byte file
        try:
            result = run(
                script=script,
                use_duckdb=use_duckdb,
            )
        except Exception as e:
```

### Expected vs Actual

**Actual:**

```
use_duckdb=False: InputValidationException: ('Duplicated records. Combination of Identifiers are repeated.', '0-1-1-6')
use_duckdb=True: InputValidationException: ('The following Identifiers Id_1 were not found, review file DS_1.csv.', '0-1-1-8')
```

**Expected:** the same error on both engines, saying the file is empty or has no header — a header-only file already loads as an empty dataset on both, so the empty file is the only unhandled shape.

### vtlengine version

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

albertohernandez1995:
## Summary

A zero-byte datapoint file was reported as `0-1-1-6` "Duplicated records. Combination of Identifiers are repeated." on the pandas engine — there are no records in it at all — and as `0-1-1-8` "The following Identifiers Id_1 were not found" on the DuckDb engine, which points at the data structure rather than at the file.

The wrong text turned out to be a mislabel rather than a wrong code: `0-1-1-6` has a single caller in the whole codebase, the empty-file check, and duplicated records already have their own code (`0-3-1-7`). So no new code was needed:

- `0-1-1-6` now says what its only caller reports, and names the file. The caller was already passing the file in, into a message with nowhere to put it.
- The DuckDb loader raises the same code when the file names no columns, before the check that was reporting an Identifier as missing.
- Both engines name the file the same way `0-1-1-8` does, so the message is identical on either engine.

A file holding only the line that names its columns is a Data Set with no Data Points and loads on both engines, as it did before: the empty file is the only shape refused.

## Checklist

- [x] Tests pass (`pytest`) — 5660 passed on DuckDb, 5052 on pandas

## Impact / Risk

- Both engines refuse an empty file, as they did before; only the code and the message change. The DuckDb engine reports `0-1-1-6` where it used to report `0-1-1-8`.
- The text of `0-1-1-6` changes for the one case that raises it. Nothing else in the codebase uses that code, so no other message changes.
- Data/SDMX compatibility: none. No file that loaded before stops loading.

## Notes

- `test_GH_676_1` was skipped on the DuckDb engine, with the reason "DuckDB backend handles empty CSVs differently and does not surface 0-1-1-6". That skip is why the divergence went unseen; the test runs on both engines now, and it is the one that fails when the source is reverted.
- `test_GH_1075_1` pins the shape next to it: a header-only file still loads as an empty Data Set on both engines.

Closes #1075


javihern98 (changes requested):
This solution is wrong, we must not change the duplicated records message, but I am open to create other message with other code for this error. Ensure the duplicated records check is enforced as well on duckdb

albertohernandez1995:
Requested changes were applied.

javihern98:
@albertohernandez1995 Before approving, could you give as well evidences that the duplicate records check is working as intended?

albertohernandez1995:
Sure, these are the evidences:

**A repeated Identifier is refused:**

| input | pandas | duckdb |
| --- | --- | --- |
| CSV, same Id twice with different Measures | `0-3-1-7` | `0-3-1-7` |
| CSV, two fully identical rows | `0-3-1-7` | `0-3-1-7` |
| CSV, the repeat not adjacent (`1`, `2`, `1`) | `0-3-1-7` | `0-3-1-7` |
| CSV, two Identifiers, both repeating | `0-3-1-7` | `0-3-1-7` |
| DataFrame, same Id twice | `0-3-1-7` | `0-3-1-7` |
| Parquet, same Id twice | — | `0-3-1-7` |
| the same period written `2020M1` and `2020-M01` | `0-3-1-7` | `0-3-1-7` |
| a Data Set with no Identifiers holding 2 Data Points | `0-3-1-4` | `0-3-1-4` |

The period row is the interesting one: it is caught because the DuckDb check runs after the period values are normalised, so two spellings of one period count as the same Data Point rather than slipping through as two.

**What is not a duplicate still loads**, on both engines:

| input | result |
| --- | --- |
| CSV, distinct Ids | loads 2 rows |
| two Identifiers, only one of them repeating (`1,a` and `1,b`) | loads 2 rows |
| distinct Ids carrying the same Measure value | loads 2 rows |

**The messages:**

pandas: On Dataset DS_1 loading: Duplicated Identifiers are not allowed, found on row 1
duckdb: On Dataset DS_1 loading: Duplicated Identifiers are not allowed, found on row (duplicate keys detected)

javihern98 (approved):
Looks good now, thanks! 😊

Issue #1074, opened 2026-08-24
A CSV Date column is rendered with a time on every row on DuckDb when any single row carries one


albertohernandez1995:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

The DuckDb loader always creates a CSV `Date` column as `TIMESTAMP` (`duckdb_transpiler/io/_io.py`, `csv_date_overrides`), and the output representation decides date-vs-datetime **per column** by checking whether every value is midnight (`duckdb_transpiler/io/_execution.py`). So one row with a time makes every other row in that column render with `T00:00:00`, which is not what the input said. The pandas engine keeps each value in the form it was given.

### Reproducible Example

```python
import tempfile
from pathlib import Path

from vtlengine import run


script = "DS_r <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Date", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

with tempfile.TemporaryDirectory() as tmp:
    csv_path = Path(tmp) / "DS_1.csv"
    csv_path.write_text("Id_1,Me_1\n1,2020-01-01\n2,2020-06-15T23:59:59\n")
    for use_duckdb in (False, True):
        result = run(
            script=script,
            data_structures=data_structures,
            datapoints={"DS_1": csv_path},
            use_duckdb=use_duckdb,
        )
        print(f"use_duckdb={use_duckdb}: loads {result['DS_r'].data['Me_1'].tolist()}")
```

### Expected vs Actual

**Actual:**

```
use_duckdb=False: loads ['2020-01-01', '2020-06-15T23:59:59']
use_duckdb=True: loads ['2020-01-01T00:00:00', '2020-06-15T23:59:59']
```

**Expected:** `['2020-01-01', '2020-06-15T23:59:59']` on both engines — a value that had no time component does not gain one.

### vtlengine version

1.9.2 (branch `1.9.X`, commit `877bb24a`)

### Python version

3.12.3

### OS

Ubuntu 24.04

### Additional Context

[...]

Issue #895, opened 2026-07-10
Parquet loader has no Date→TIMESTAMP detection: datetime values in Date columns are silently truncated to dates


javihern98:
### Summary

`_load_parquet` creates the target table directly from the component definitions (`build_create_table_sql(dataset_name, components)`) with **no Date→TIMESTAMP override detection** — unlike the CSV path and the DataFrame path (`_detect_date_type_overrides`). A `Date` component is therefore always created as `DATE`, and any datetime value in a parquet file is silently truncated to a bare date by the `CAST`, after passing the strict `VALID_DATE_REGEX` validation that declares it acceptable input.

The identical value loaded from CSV keeps the full datetime, so the parquet and CSV input paths disagree.

### Reproducible Example

```python
import tempfile
from pathlib import Path

import pandas as pd

from vtlengine import run

script = "DS_A <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_1", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Date", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

value = "2020-01-15T10:30:00"
tmp = Path(tempfile.mkdtemp())

# Parquet path (string-typed column in the file)
parquet = tmp / "DS_1.parquet"
pd.DataFrame({"Id_1": [1], "Me_1": [value]}).to_parquet(parquet)
res = run(script=script, data_structures=data_structures, datapoints={"DS_1": str(parquet)})
print("Parquet ->", res["DS_A"].data["Me_1"].tolist())

# Same value via CSV path
csv = tmp / "DS_1.csv"
csv.write_text(f"Id_1,Me_1\n1,{value}\n")
res = run(script=script, data_structures=data_structures, datapoints={"DS_1": str(csv)})
print("CSV     ->", res["DS_A"].data["Me_1"].tolist())
```

### Expected vs Actual

**Actual:**

```
Parquet -> ['2020-01-15']
CSV     -> ['2020-01-15T10:30:00']
```

**Expected:** the parquet load preserves the time component (`2020-01-15T10:30:00`), matching the CSV path.

> [!NOTE]
> A fix can reuse the zero-row schema scan the loader already performs (`_read_parquet_columns` returns the parquet column types) or probe the values, and pass `type_overrides` to `build_create_table_sql` exactly as the DataFrame path does.

### vtlengine version

[...]

PR #1094, opened 2026-08-28
Fix #894: a Date value keeps its time when a DataFrame carries it short or already parsed

### LINKED ISSUE #894: DataFrame Date columns silently lose their time component when the TIMESTAMP override detection misses them (1-digit-month strings, typed datetime64)
### Summary

On the DataFrame load path, `_detect_date_type_overrides` decides whether a `Date` component is stored as `DATE` or `TIMESTAMP` by inspecting string values with a fixed positional check (`len(val) > 10 and val[10] in ("T", " ")`). Two kinds of values escape it:

1. **Datetime strings with 1-digit month/day** (`2020-1-1T12:30:45`): the separator is not at position 10, so no override is detected. The value passes the strict Date validation gate (`VALID_DATE_REGEX` deliberately allows `\d{1,2}` month/day), the column is created as `DATE`, and the `CAST` silently drops the time — the exact silent truncation the validation gate exists to prevent.
2. **Typed `datetime64` columns** carrying times: the detector only looks at `isinstance(val, str)` values, so a pre-parsed `pd.to_datetime` column never triggers the override and is truncated to a bare date.

In both cases the identical value loaded through the CSV path keeps the full datetime, so the two input paths disagree.

### Reproducible Example

```python
import tempfile
from pathlib import Path

import pandas as pd

from vtlengine import run

script = "DS_A <- DS_1;"

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
            ],
        }
    ]
}

value = "2020-1-1T12:30:45"  # 1-digit month/day, full time

# DataFrame path
res = run(
    script=script,
    datapoints={"DS_1": pd.DataFrame({"Id_1": [1], "Me_1": [value]})},
)
print("DataFrame (1-digit-month str) ->", res["DS_A"].data["Me_1"].tolist())

# Same value via CSV path
csv = Path(tempfile.mkdtemp()) / "DS_1.csv"
csv.write_text(f"Id_1,Me_1\n1,{value}\n")
res = run(script=script, data_structures=data_structures, datapoints={"DS_1": str(csv)})
print("CSV       (same value)        ->", res["DS_A"].data["Me_1"].tolist())

# Typed datetime64 column carrying a time
res = run(
    script=script,
    datapoints={
        "DS_1": pd.DataFrame({"Id_1": [1], "Me_1": pd.to_datetime(["2020-01-01 12:30:45"])})
    },
)
print("DataFrame (datetime64 w/time) ->", res["DS_A"].data["Me_1"].tolist())
```

### Expected vs Actual

**Actual:**

```
DataFrame (1-digit-month str) -> ['2020-01-01']
CSV       (same value)        -> ['2020-01-01T12:30:45']
DataFrame (datetime64 w/time) -> ['2020-01-01']
```

**Expected:** all three loads preserve the time component (`2020-01-01T12:30:45`), matching the CSV path.

> [!NOTE]
> A fix should make the override detection agree with `VALID_DATE_REGEX` (accept 1-2 digit month/day before the separator) and treat `datetime64`-typed columns with any non-midnight time as `TIMESTAMP`, instead of relying on a fixed separator position over string values only.

### vtlengine version

main (2.0.0rc3, commit db24535c)

### Python version

3.12.3

### OS

Ubuntu 24.04

albertohernandez1995:
## Summary

A Date value handed over in a DataFrame lost the time it carried, in two ways the CSV path handles correctly, and on `1.9.X` one of them does not truncate but crashes:

- **A month or a day written short.** `2020-1-1T12:30:45` moved the separator off position 10, which is where the override detection looked for it, so the column was created as `DATE` and the cast dropped the time. The same value in a CSV keeps it.
- **A column pandas had already parsed.** The detection read `isinstance(val, str)` only, so a `datetime64` column never triggered the override and came out as a bare date on the DuckDb engine. On the pandas engine it did not truncate: `check_date` assumed text and let a raw `AttributeError` escape — `'Timestamp' object has no attribute 'strip'` — for a column carrying a time **and** for one carrying only dates.

Both engines read such a value now:

- `check_date` takes what pandas hands it. A datetime becomes its ISO text; one sitting exactly at midnight becomes the bare date it stands for, so a parsed column of dates reads as dates rather than gaining a `T00:00:00`. A plain `date` object is read as well.
- `_detect_date_type_overrides` asks whether each value writes a time, rather than looking for a separator at a fixed place in text: a regex that allows a month or a day written short, as `VALID_DATE_REGEX` allows them, and a non-midnight check for a value already parsed.

## Checklist

- [x] Tests pass (`pytest`) — 5718 passed on DuckDb, 5110 on pandas

## Impact / Risk

- A DataFrame column that lost its time now keeps it, on both engines. Results and CSV output change for exactly those values, and change to what the input said.
- A `datetime64` column no longer raises `AttributeError` on the pandas engine; it loads.
- Nothing that loaded correctly before changes: a column of bare dates, of date strings, or of strings carrying a time reads as it did.
- Data/SDMX compatibility: none.

## Notes

- Measured across eight input shapes, all agreeing on both engines: a month or day written short with a time, a parsed datetime with a time, with microseconds and date-only, Python `date` objects, plain date strings, strings carrying a time, and the same value through a CSV.
- The tests are new `dataload_params` entries in `tests/DateTime/test_datetime.py`, which is that file's DataFrame-load idiom and runs on whichever engine the suite is under. Reverting the source fails three of them on each engine, and not the same three: the DuckDb engine fails the short month and the parsed datetimes, while the pandas engine fails the three parsed cases, the date-only one among them, since that is the crash.

> [!NOTE]
> A Date column mixing values that carry a time with values that do not still renders every value with `T00:00:00` on the DuckDb engine. That is #1074, and it is not fixable where this fix lives: a TIMESTAMP sitting at midnight cannot say whether the input wrote a time, so the column-level decision has to guess. Closing it, and #895 with it, means having the DuckDb engine keep the form the value arrived in. `dataload_params[mixed_date_and_datetime]` still carries a per-engine expectation for that reason.

Closes #894

### DIFF
--- a/src/vtlengine/DataTypes/_time_checking.py
+++ b/src/vtlengine/DataTypes/_time_checking.py
@@ -2,6 +2,7 @@ import calendar
 import re
 from datetime import date, datetime
 from functools import lru_cache
+from typing import Any
 
 from vtlengine.DataTypes.TimeHandling import TimePeriodHandler
 from vtlengine.Exceptions import InputValidationException
@@ -12,6 +13,21 @@ _DATE_PART_RE = re.compile(r"^(?P<year>\d{4})-(?P<month>\d{2})-(?P<day>\d{2})")
 def _has_time_component(value: str) -> bool:
     """Check if a date string includes a time component (T or space separator at position 10)."""
     return len(value) > 10 and value[10] in ("T", " ")
+
+
+def _from_date_value(value: Any) -> str:
+    """The text of a value pandas has already parsed into a date or a datetime.
+
+    A datetime at midnight is written as the bare date it stands for, so a column of
+    dates reads as dates whether pandas parsed it or left it as text.
+    """
+    if isinstance(value, datetime):
+        if (value.hour, value.minute, value.second, value.microsecond) == (0, 0, 0, 0):
+            return value.strftime("%Y-%m-%d")
+        return value.isoformat(sep=" ")
+    if isinstance(value, date):
+        return value.isoformat()
+    return str(value)
 
 
 def _truncate_nanoseconds(value: str) -> str:
@@ -105,14 +121,19 @@ def _pad_date_parts(value: str) -> str:
     return f"{year}-{int(month):02d}-{int(day):02d}{rest}"
 
 
-def check_date(value: str) -> str:
+def check_date(value: Any) -> str:
     """Check a date is in the correct format.
 
     Accepts ``YYYY-MM-DD`` and ``YYYY-MM-DD[T| ]HH:MM:SS[.ffffff][+HH:MM|Z]`` (ISO 8601).
     A time component, when present, must be a COMPLETE ``HH:MM:SS``; partial times such
     as ``HH`` or ``HH:MM`` are rejected. Datetime output uses a single space separator;
     nanosecond input is truncated to microsecond precision.
+
+    A DataFrame may carry a column pandas has already parsed, so a date or a datetime
+    arrives as itself rather than as text. It used to reach here and raise a raw
+    AttributeError for the strip a string would have.
     """
+    value = _from_date_value(value)
     value = _pad_date_parts(value.strip())
     has_time = _has_time_component(value)
     try:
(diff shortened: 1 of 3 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #1095, opened 2026-08-28
Fix #895: a Date value in a parquet file keeps the time it carries

### LINKED ISSUE #895: Parquet loader has no Date→TIMESTAMP detection: datetime values in Date columns are silently truncated to dates

albertohernandez1995:
Closes #895 

## Summary

`_load_parquet` created its table straight from the component definitions, with none of the Date to TIMESTAMP detection the CSV and DataFrame paths do, so every Date column was created as `DATE` and a value carrying a time was cast down to the bare date — after passing the validation that declares such a value acceptable. The same value read from a CSV keeps the time, so the two input paths disagreed.

The loader now reads the file's own column types from the zero-row scan it already performs, and decides per Date column:

- a column the file types as a timestamp is stored as one;
- a column of text is read the way the CSV path reads it, by asking whether any value writes a time after the date, with a month or a day allowed to be written short as `VALID_DATE_REGEX` allows them;
- anything else stays a `DATE`.

Those overrides go to `build_create_table_sql` and to `_build_dataframe_select_columns`, exactly as the DataFrame path passes them.

> [!NOTE]
> The issue reports a string column losing its time. A column the parquet file already types as a **timestamp** lost it too, for the same reason, and is covered here.

## Checklist

- [x] Tests pass (`pytest`) — 5750 passed on DuckDb, 5142 on pandas

## Impact / Risk

- A parquet Date column carrying a time now keeps it. Results and output change for exactly those values, and change to what the file said.
- A column of bare dates loads as before, whether the file types it as a date or writes it as text.
- The table is created after the file has been read rather than before, so the file's types can be taken into account. A file DuckDB cannot read is still reported as `0-3-1-16`, since the create now sits inside the same error mapping.
- One extra query per parquet load, and only when a Date column is stored as text.

## Notes

- The tests live in `tests/Parquet/test_parquet_io.py` and follow that file's pattern. They cover a time written as text, a column the file types as a timestamp, microseconds, a month and day written short, bare dates staying dates, and the value read both ways to check the two paths agree. Five of the six fail when the source is reverted.
- **Expect one conflict with branch `cr-1069` (issue #1069)**, in `_io.py`: that branch adds a `dataset_name` parameter to `_build_dataframe_select_columns` while this one passes `type_overrides` to the same call. The resolution is the union of both arguments:
  ```python
  select_exprs = _build_dataframe_select_columns(
      components, dataset_name, df_columns=parquet_cols, type_overrides=date_overrides
  )


javihern98 (dismissed):
Looks good, thanks! 😊

javihern98 (approved):

PR #1025, opened 2026-08-11
Fix #1024: a dataset-level if corrupts its condition dataset

javihern98:
## Summary

A dataset-level `if` mutated its condition dataset in place: `If.dataset_level_evaluation` popped the boolean measure column out of `condition.data`, which is the very DataFrame the stored condition variable holds (`visit_VarID` hands out shallow copies). After the `if` executed, the variable's data no longer had its measure column, so any later statement reading it — the `case` in the issue, but a second `if` fails identically — raised `KeyError` on the measure name.

The condition working frame is now built from copies (`condition.data[ids].copy()` plus the measure column), leaving the operand untouched. `Case.dataset_level_evaluation` already worked this way; `if` was the only mutator.

Fixes #1024

## Checklist

- [x] Code quality checks pass (`ruff format`, `ruff check`, `mypy`)
- [x] Tests pass (`pytest`)
- [ ] Documentation updated (if applicable)

## Impact / Risk

- No breaking changes: the `if` result itself was always correct; the fix only stops the side effect on the condition operand.
- Pandas engine only — the DuckDB transpiler never had the defect (the new test runs on both backends).
- Release notes: bug fix, `if` no longer corrupts a reused condition dataset.

## Notes

- New file-based test `GH_1024` in `tests/Bugs` mirrors the issue script (`if` + `case` over the same renamed condition dataset) and pins all five intermediate/final datasets, so it also guards the condition variable itself against future in-place mutation.
- Full pandas suite: 4954 passed. DuckDB backend on `tests/Bugs`, `tests/IfThenElse`, `tests/ThreeValueLogic`: 327 passed.

PR #1100, opened 2026-08-31
Port main-only changes to 1.9.X (excluding the pandas backend removal)

### LINKED ISSUE #288: Implement nullability over scalars

### LINKED ISSUE #804: Support running vtlengine under Pyodide (WebAssembly)
### Initial Checks

## Problem Description

vtlengine cannot currently be installed or run under [Pyodide](https://pyodide.org) (CPython compiled to WebAssembly). This blocks browser-based use cases: in-browser validation/execution, JupyterLite, PyScript, and interactive documentation playgrounds.

Two things prevent it:

1. The compiled `vtl_cpp_parser` extension (ANTLR4 C++ runtime + pybind11) has no Emscripten/`wasm32` wheel — only `manylinux`/`macOS`/`Windows` wheels are published.
2. Two dependencies don't fit the Pyodide environment:
   - `psutil` — cannot be installed under Pyodide (no OS process APIs).
   - the `duckdb` pin (`>=1.4,<1.5`) excludes the version Pyodide ships (1.5.1).

## Proposed Solution

Make vtlengine buildable and runnable under Pyodide:

- **Build** — cross-compile `vtl_cpp_parser` for Emscripten. The blocker is a `wasm-ld` *"function signature mismatch"* on `Recognizer::getRuleIndexMap`/`getTokenTypeMap`, caused by mixing pybind11's LTO (enabled in Release) with the non-LTO ANTLR4 static runtime — `wasm-ld` matches symbols by name **and** type signature, unlike native linkers. Disabling LTO under Emscripten (`NO_EXTRAS`, guarded by `if(EMSCRIPTEN)`) resolves it; native builds keep LTO.
- **psutil** — remove the dependency. It was only used to size DuckDB's `memory_limit` for percentage values. DuckDB's own default is already 80% of RAM, so percentage limits defer to DuckDB and absolute limits pass straight through.
- **duckdb** — widen the range to `>=1.4,<2.0` (keep 1.4 as the LTS / locked default) so Pyodide's 1.5.x is allowed.

> [!NOTE]
> Removing psutil changes one behaviour: a *custom* percentage limit (e.g. `VTL_MEMORY_LIMIT="50%"`) is no longer computed — it now defers to DuckDB's built-in default of 80% of RAM. The default (`"80%"`) and all absolute values are unaffected.

**Status — prototyped and validated:**

- `vtl_cpp_parser` cross-compiles to a valid Pyodide wheel and parses VTL (valid + invalid) in a Pyodide/Node runtime.
- Full test suite passes under duckdb 1.5.3 (4862 passed, 0 failures).
- psutil removed with the memory-limit change; ruff + mypy clean.

## Alternatives Considered

- *Keep psutil but guard the import with a fallback* — still requires `psutil` as a dependency (which Pyodide can't install) and is more code than letting DuckDB apply its own default.
- *Pass percentage strings straight to DuckDB* — rejected: DuckDB does not parse `%` syntax (`Unknown unit for memory: '%'`).

## Additional Context

Remaining after this issue:

- Target the cp314 / 2026 Pyodide line (where duckdb 1.5.1 and pyarrow 22 live) rather than the 0.27 line used for the prototype.
- End-to-end `run()` validation in Pyodide (only the parser was exercised in isolation so far).
- Decide how to distribute the wasm wheel (publish it, or add a recipe to `pyodide-recipes`).

### LINKED ISSUE #890: unpivot into a component name colliding with a viral attribute raises a bare ValueError instead of a SemanticError

### LINKED ISSUE #891: Date values with 1-2 fractional-second digits fail on Python 3.9/3.10 (validator regex accepts, fromisoformat cannot parse)

### LINKED ISSUE #901: create_ast is not thread-safe: C++ parser keeps the last parse tree in process-global state

javihern98:
## Summary

Back-ports every main-only change to `1.9.X`, excluding the removal of the pandas execution backend. main's last content sync into the 1.9 line was 1.9.1 (#886, the reverse direction); since then main accumulated fixes and features 1.9.X never received. Scope, commit by commit against `origin/1.9.X..origin/main`:

| main commit | PR | Status here |
| --- | --- | --- |
| 4b8e4bed | #762 — nullability over scalars (#288) | **Ported**, adapted to both engines; un-skips the tests parked as "deactivated until scalars nullability implementation" |
| 56c17ed6 | #863 — updated DataStructure JsonSchema validation | **Ported**; the test helper's pandas loader now validates with the same schema |
| a3452759 | #871 — numpy>=2.4.4 marker split for cp314 wheels | **Ported** (lock regenerated with Poetry) |
| de6f5160 | #805 — Pyodide/WebAssembly support (#804) | **Ported in full**: CMake `NO_EXTRAS` under Emscripten, psutil removed, `duckdb<2.0`, JupyterLite demo + docs workflow job |
| b32238a2 | #873 — pin GitHub Actions to commit SHAs | **Ported, then refreshed** (`c1a8f8f0`): every action is pinned to the commit of its latest published release as of 2026-09-02 — checkout v7.0.1, setup-python v7.0.0, deploy-pages v5.0.1, gh-action-pypi-publish v1.14.2, cache v6.1.0, github-script v9.0.0, configure-pages v6.0.0, upload-pages-artifact v5.0.0 — and `release.yml` moves from cibuildwheel 3.4.1 / artifact actions v4.x to cibuildwheel 4.2.0 / upload-artifact v7.0.1 / download-artifact v8.0.1 (main is on 4.1.0 / v7.0.1 / v8.0.1) |
| 8254248d | #899 — unpivot name collision with a viral attribute (#890) | **Ported** (new SemanticError 1-1-6-14) |
| 7b624d20 | #896 — DAG kept ViralPropagationDef out of single-expression scripts | **Ported**; its tests renumbered 13-3/13-4 → 14-1/14-2 (codes already taken on 1.9.X) |
| b485bfcb | #898 — datapoint ruleset whose signature variable is a viral attribute | **Ported**, adapted to also keep the signature column in the pandas data; test renumbered 13-2 → 14-3 |
| ec0e64fd | #900 — normalize fractional seconds to microseconds (#891) | **Ported** (pads short fractions; 1.9.X only truncated long ones), keeping 1.9.X's `_from_date_value` from #1094 |
| 586b124c | #902 — serialize C++ parser access (#901) | **Ported**, combined with 1.9.X's `recursion_headroom` (#939); fixes a real intermittent failure seen locally under xdist on this branch's baseline |
| db24535c | #886 — merge 1.9.1 into main | Its **main-only adaptation** (Date string guard only on string-like source columns) is ported: a tz-aware `datetime64` DataFrame loaded on pandas but failed on DuckDb |
| 8a42d244, 492b0664, f52caff2, 18301c7f, 20ce4d5a, 5b1d10a2, 8e4012fe, 94d4914c | #882, #791, #794, #797, #799, #788, #781, #775 | Already on 1.9.X (peer PRs #883, #795, or verified byte-equivalent code) — nothing to do, except #791's `test_create_ast_with_comments_syntax_error_raises`, which 1.9.X had lost when #815 made its script valid; restored (`6d0ed8cf`) |
| fe146641 | #897 — window check_datapoint viral aggregate | **Not ported**: superseded by 1.9.X's combination-point model (#907), under which check_datapoint is row-preserving and copies viral attributes |
| a6316898 | #783 — viral propagation during DuckDB execution | Not ported: superseded by #878/#907 on 1.9.X |
| 7abdec50, 1050194b + cr-682 series | #768, #718 | **Excluded**: pandas backend removal and its dead-code cleanup — except two engine-neutral fixes buried in the series, ported by the completeness audit below: `a86916af` (UDO name resolution in the SQL transpiler, `-x` cherry-pick `868baec0`) and the DataFrame column-name normalization from `88723f67` (`fb6be370`) |
| 7c8d1b62, c5b81eba, e5dc21e5 | #872, #869, #865 | Superseded: 1.9.X's lock (#1098) already carries pysdmx 1.19.0 and pytest 9.1.1 |
| 5cabbc7a, 59b3bd5f, e38ecd3f, 57861fb6 | #862, version bumps, 1.9.0 merge | N/A (main-side test alignment / version housekeeping) |

Fixes #288, fixes #804, fixes #890, fixes #891, fixes #901. References #863, #873, #896, #897, #898, #886.

### Adaptations worth reviewing

- **Scalar nullability (#762) on 1.9.X's evolved operators**: `Nvl.validate` keeps 1.9.X's promoted result type (#1016) and derives nullability from the existing `keeps_nulls` (nullable only when the replacement can be null), instead of main's unconditional `nullable=False`. `Date_Add.validate` keeps 1.9.X's `Date` result type. `Cast` threads `nullable` through 1.9.X's `scalar_evaluation`.
- **BIRD semantic references**: the `INPUT_LAYER_TO_ENRICHED_INPUT_LAYER` semantic test had been skipped since 1.9.0 awaiting this feature; 15 components across 3 reference structures updated to current semantics (a calc'd non-null constant is not nullable; a pass-through Measure keeps its declared nullability). main still passes with the old values only because later statements there mutate shared structures — 1.9.X fixed those mutations (#982, #1025, #1058).
- **JsonSchema (#863) fallout**: `tests.Helper.LoadDataset` (pandas branch) now validates input JSON with the public schema so both engines report the same structural errors, and the GL_473/474/475 UDO inputs' literal `__index_id__` column (schema-invalid, incidental to those tests) was renamed `INDEX_ID`.
- **#898 on the pandas engine**: the interpreter keeps the signature-named viral attribute in the ruleset dataset's *data* too, not only its components.

## Checklist

- [x] Tests pass (`pytest`) — full suite green on both engines (duckdb: 6074 passed / 19 skipped; pandas: 5135 passed / 630 skipped). The pandas job runs fewer tests than before the completeness audit because the 167 `test_reference_duckdb` cases no longer run through DuckDB there (see the `_use_duckdb_backend()` fix below); it gained random cases 184/185 and 15 previously always-skipped tests
- [x] Documentation updated (`docs/environment_variables.rst`, `docs/jupyterlite/README.md`, walkthrough link — all from #805)

## Impact / Risk

- **Behaviour changes** (all matching main):
  - `semantic_analysis`/`run` scalar results now carry and propagate `nullable`; scalar structure output includes a `"nullable"` field.
  - Input `data_structures` are validated against a stricter JSON schema up front: role/type enums, component-name pattern (no leading underscore), `oneOf` DataStructure/structure — previously-silent malformed inputs now raise `0-2-1-1`.
  - A custom percentage `VTL_MEMORY_LIMIT` (e.g. `"50%"`) defers to DuckDB's built-in 80%-of-RAM default; absolute values unchanged (psutil dropped).
  - `duckdb` dependency range widened to `>=1.4,<2.0` (lock keeps 1.4.5); pip installs may resolve newer duckdb — same range main ships.
- **Release/publish path**: `release.yml` now runs cibuildwheel 4.2.0 and the Node 24 artifact actions (see table). cibuildwheel 4.x adds `delvewheel` as the default Windows repair step and `abi3audit` after repair; the extension links the ANTLR runtime statically and the wheels are per-version, so both are expected to be no-ops. None of these pins has run a release in this repository yet (every release so far came from 1.9.X with the old pins; main has not released since #873), so the first rc after this lands is the one to watch — together with the psutil-less wheel and the widened duckdb range in the fresh-resolution smoke test.
- No SDMX compatibility concerns.

### Post-review fixes on this branch

A multi-angle code review of the port surfaced two defects in my adaptations, both fixed and pinned by tests (`8b72cf95`):

- **#898 port crashed the pandas engine in the default `invalid` output mode** (`KeyError: ['VAt_1'] not in index`): the kept signature viral column was re-attached a second time by `Validation.evaluate`. Only missing virals are re-attached now; new case `14-4` covers the default mode on both engines.
- **`nvl` scalar nullability was value-dependent** (`keeps_nulls` reads `right.value`, populated only at run time), so `semantic_analysis()` and `run()` disagreed on the same script's structure. It now reports `nullable=False`, exactly as main does and as main's ported test pins.

An earlier round (`c36a883c`) already moved `create_dag` outside `parser_lock` (main's structure), deduplicated the parquet schema scan, and compiled the JSON-schema validators once (~53s → ~34s full-suite, since the test helper now validates every structure).

### Completeness audit (2026-09-02)

A hunk-level audit of all 41 main-only commits against this branch found three engine-neutral fixes hidden inside the excluded pandas-removal series or its merge adaptations, now ported:

- **UDO name resolution in the SQL transpiler** (`a86916af`, cherry-picked as `868baec0`): a user-defined operator that references a component parameter in a clause, aggregation or analytic failed on DuckDB with `Binder Error: Referenced column "<param>" not found` — 21 of the reference-manual UDO variants. On 1.9.X the substitution also had to move inside `_resolve_partition_cols`, otherwise `partition except <param>` silently partitions by every identifier (main carries that latent gap; not filed). The UDO reference variants now run on DuckDB too (`test_reference_defined_operators_duckdb`, `1726bcb9`) and UDO test `1-1-1-13` pins the partition case.
- **DataFrame column names on DuckDB** (from `88723f67`, `fb6be370`): the pandas loader strips a UTF-8 BOM and coerces labels to str (#653); DuckDB rejected `\ufeffId_1` as undefined and crashed with a `TypeError` on a non-string label. Both engines now apply the same normalization, on a shallow copy of the DataFrame; the tests run on both engines.
- **`test_create_ast_with_comments_syntax_error_raises`** (#791) restored (`6d0ed8cf`).

Also fixed while there (`e0be6a94`): seven test skip conditions in four modules used the `_use_duckdb_backend` function object as a bool, which is always truthy. The pandas CI job was running the DuckDB reference-manual tests through DuckDB and dropping random cases 184/185, and the whole VirtualCounter module plus DataLoad `test_12`/`test_15` and Semantic `test_46` were skipped on both engines. They now evaluate the switch; the un-skipped tests pass on pandas.

Parity fix found while testing the above (`384d4b52`): an empty string in a DataFrame column loaded as null on the pandas engine for every type but String, while the DuckDB DataFrame path failed the cast (`0-3-1-6`, or a bare `ValueError` at output time for `Time_Period`) or kept it as a value (`Time`). String-like source columns are now read through `NULLIF(..., '')` for the non-String types, so a DataFrame loads like a CSV with an empty field on both engines; `TestEmptyStringFromDataFrame` covers every type, the String exception and the non-nullable Measure case (`0-3-1-17`) on both engines.


albertohernandez1995 (approved):
Nice job!

PR #1104, opened 2026-09-02
Fix #1102: resolve UDO component parameters before applying partition except

### LINKED ISSUE #1102: `partition except` with a component parameter inside a user-defined operator partitions by every identifier
### Initial Checks

### Summary

Inside a user-defined operator, an analytic `partition except <component parameter>` ignores the exclusion and partitions by every identifier of the operand. What gets removed from the identifier list is the parameter name, not the identifier bound to it, so nothing is removed. `partition by <component parameter>` and `partition except all` are not affected, and the same expression written inline (outside the UDO) is correct.

### Reproducible Example

```python
import pandas as pd

from vtlengine import run


script = """
    define operator sum_except (x dataset, c component, e component, m component)
        returns dataset is
        x[calc Me_2 := sum(m over (partition except c order by e))]
    end operator;

    DS_A <- sum_except(DS_1, Id_1, Id_2, Me_1);
    DS_B <- DS_1[calc Me_2 := sum(Me_1 over (partition except Id_1 order by Id_2))];
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Id_2", "type": "Integer", "role": "Identifier", "nullable": False},
                {"name": "Me_1", "type": "Integer", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

data_df = pd.DataFrame(
    {"Id_1": ["A", "A", "B", "B"], "Id_2": [1, 2, 1, 2], "Me_1": [10, 20, 30, 40]}
)

print("DS_A (through the user-defined operator):")
print(run_result["DS_A"].data)
print("DS_B (same expression written inline):")
print(run_result["DS_B"].data)
```

### Expected vs Actual

Expected: `DS_A` (through the UDO) equals `DS_B` (the same expression inline). `Me_2` is the sum of `Me_1` over the partition formed by all identifiers except `Id_1`, i.e. by `Id_2`. This is what `DS_B` returns:

| Id_1 | Id_2 | Me_1 | Me_2 |
| --- | --- | --- | --- |
| A | 1 | 10 | 40.0 |
| B | 1 | 30 | 40.0 |
| A | 2 | 20 | 60.0 |
| B | 2 | 40 | 60.0 |

Actual: `DS_A` partitions by `Id_1` and `Id_2`, so every partition holds a single row and `Me_2` is just `Me_1`:

| Id_1 | Id_2 | Me_1 | Me_2 |
| --- | --- | --- | --- |
| A | 1 | 10 | 10.0 |
| B | 2 | 40 | 40.0 |
| B | 1 | 30 | 30.0 |
| A | 2 | 20 | 20.0 |

### vtlengine version

2.0.0rc3 (`main` at 7b624d20)

### Python version

3.12.3

### OS

Ubuntu 24.04.4 LTS

### Additional Context

`SQLTranspiler._resolve_partition_cols` builds `excluded = set(listed)` from the raw `node.partition_by` names, i.e. the parameter name `c`, and removes those from the operand's identifiers. The UDO substitution (`_resolve_udo_name`) is applied by the callers (`_build_over_clause` and the viral over-clause) only to the list the helper returns, which is why `partition by c` works and `partition except c` does not.

> [!NOTE]
> On `1.9.X`, PR #1100 already moves the substitution inside `_resolve_partition_cols` (commit 868baec0, the `-x` cherry-pick of a86916af with that adaptation) and UDO test `1-1-1-13` covers this case. Fixing it here keeps both branches aligned.

javihern98:
## Summary

Inside a user-defined operator, `partition except <component parameter>` partitioned by every identifier of the operand: `_resolve_partition_cols` removed the parameter name from the identifier list instead of the identifier bound to it, while the UDO substitution was applied by the callers only to the list the helper returned. The substitution now happens inside the helper, and `_build_over_clause` and the viral over-clause receive already-resolved names. This is the same adaptation the 1.9.X port carries in PR #1100 (commit 868baec0), so the two branches stay aligned.

UDO test `1-1-1-13` (component parameters in `partition except ... order by ...`, same data files as on 1.9.X) pins the behaviour; it failed on `main` before the change.

Fixes #1102

## Checklist

- [x] Code quality checks pass (`ruff format`, `ruff check`, `mypy`) — mypy reports three `unused-ignore` errors in `src/vtlengine/Model/__init__.py`, a file this PR does not touch; they appear on the untouched `main` checkout in the same environment, which runs mypy 2.3.1 (the 1.9.X lock) instead of the 1.20.2 that `main`'s lock pins
- [x] Tests pass (`pytest`) — full suite: 5054 passed, 11 skipped

## Impact / Risk

- Behaviour change only for user-defined operators that use `partition except` with a component parameter: they returned wrong values silently and now partition as written. `partition by <parameter>`, `partition except all` and inline analytics are unchanged.
- No API or SDMX compatibility concerns.

## Notes

- The 1.9.X counterpart is already in PR #1100; the test files are identical on both branches.


javihern98:
Closing without merging: this fix already lives on 1.9.X (PR #1100, commit 868baec0 with UDO test 1-1-1-13) and 1.9.X will be merged into main in about a month, which brings it here. Opening the same change twice only creates merge noise.

PR #1109, opened 2026-09-03
Fix #1107: Support pandas 3.0, move the JupyterLite demo to Pyodide 314 and publish a wasm wheel to PyPI

### LINKED ISSUE #1107: Support pandas 3.0, move the JupyterLite demo to Pyodide 314 and publish a wasm wheel to PyPI
### Summary

PR #1100 ports the JupyterLite demo and the Pyodide/Emscripten build support to `1.9.X` (Emscripten CMake fix, psutil removal, `duckdb>=1.4,<2.0`). The demo still targets the **outgoing Pyodide 0.29.3 runtime** (Python 3.13), and it is pinned there for one reason: vtlengine caps `pandas>=2.1.4,<3.0`, while the current Pyodide line (314.x = Python 3.14.2) ships **pandas 3.0.2** in its distribution and no pandas 2.x wasm wheel exists anywhere. The pandas cap is the single vtlengine dependency that cannot resolve on Pyodide 314.

Three changes, in order:

1. **Support pandas 3.0.** Widen the constraint as a single unmarked change: `pandas>=2.1.4,<3.0` → `pandas>=2.1.4,<4.0`. No Python-version markers: pandas 3.0 declares `requires-python >= 3.11`, so resolvers keep selecting pandas 2.x on Python 3.9/3.10 by themselves, and the floor stays at 2.1.4 because the Ubuntu 24.04 workflow tests the APT pandas 2.1.4 on Python 3.12. The engine therefore has to support 2.1.4 through 3.x simultaneously, on both execution backends — headline pandas 3 changes to absorb: copy-on-write always on, the dedicated `str` dtype as default (PDEP-14) instead of `object`-dtype strings, and removed deprecated APIs.
2. **Move the JupyterLite demo to the latest Pyodide 314.x** (314.0.4 at time of writing, Python 3.14.2), dropping the 0.29-era workarounds that become unnecessary.
3. **Publish the wasm wheel to PyPI from `release.yml`.** PyPI accepts PEP 783 `pyemscripten_*_wasm32` platform tags since April 2026, and cibuildwheel v4.2.0 — the version `release.yml` already pins — builds Pyodide 314 wheels by default via its pyodide platform. One extra build job produces `vtlengine-<version>-cp314-cp314-pyemscripten_2026_0_wasm32.whl` into the same `wheelhouse/`, and the existing trusted-publishing step uploads it unchanged. With pandas 3 allowed, every dependency then resolves on Pyodide 314 (duckdb 1.5.1, pyarrow 22.0.0, numpy 2.4.3 and msgspec 0.20.0 in the distribution; sqlglot/networkx as pure-Python wheels from PyPI), so `micropip.install("vtlengine")` works end-to-end in the browser.

> [!IMPORTANT]
> The pandas work gates the rest: the Pyodide 314 distribution only carries pandas 3.0.2, so neither the demo move nor dependency resolution for the published wheel works until vtlengine accepts pandas 3. (The wheel itself could be uploaded earlier, but installing it would require the `micropip.install(..., deps=False)` + preload workaround.)

### Scope

**pandas 3.0:**

- `pyproject.toml` + `poetry.lock`: widen the single constraint to `pandas>=2.1.4,<4.0` — no version markers; the lock must resolve pandas 3.0.x for Python >= 3.11 environments while 3.10 stays on 2.x
- `src/`: compatibility with both pandas 2.1.4 and pandas 3.x, on both the pandas and DuckDB backends (copy-on-write, `str` dtype, removed APIs)
- `.github/workflows/ubuntu_test_24_04.yml`: **unchanged** — pandas 2.1.4 remains in-range on Python 3.12, and that job doubles as the pandas 2.x regression leg

**JupyterLite demo (`docs/jupyterlite/` + the `build-jupyterlite` job in `.github/workflows/docs.yml`):**

- `requirements.txt`: `jupyterlite-core` / `jupyterlite-pyodide-kernel` `~=0.7.0` → `~=0.8.0` (0.8.x is the first line shipping the Pyodide 314 kernel); drop the `wheel>=0.45` retag helper
- `build-wheel.sh`: host Python 3.13 → 3.14, `pyodide xbuildenv install 0.29.3` → 314.0.x, emsdk 4.0.9 → the version the chosen xbuildenv pins (5.0.3 for 314.0.x); **delete the retag step** — pyodide-build's native `pyemscripten_2026_0_wasm32` tag is exactly what Pyodide 314's micropip (0.11.1) expects
- `build.sh` + `patch_lock.py`: drop the xlwings `duckdb-1.5.0-cp313` URL wheel (duckdb 1.5.1 is in the 314 distribution and satisfies `>=1.4,<2.0`); drop the `pysdmx==1.15.1` pin (the distribution's msgspec 0.20.0 supports current pysdmx); pandas now comes from the distribution — update the injected-wheel list and the `depends` entries in the patched lockfile accordingly
- `docs.yml`: `python-version: '3.13'` → `'3.14'`; emsdk, xbuildenv and Pyodide-dist cache keys

**wasm wheel in `release.yml`:**

- `pyproject.toml`: new `[tool.cibuildwheel.pyodide]` section — `before-build = "bash {project}/scripts/setup_antlr4_runtime.sh"` must be repeated there (the `[tool.cibuildwheel.linux]` section does not apply to the pyodide platform), plus a pyodide-specific `test-command`: `tests/API` will not pass wholesale under wasm (no threads/multiprocessing, filesystem/S3 differences), so start with a smoke test (`import vtlengine` + a minimal `run()`) executed in cibuildwheel's `pyodide venv` (requires Node on the runner), or disable tests for this job initially
- `.github/workflows/release.yml`: one additional ubuntu build job (or matrix entry) running cibuildwheel with `CIBW_PLATFORM: pyodide`; the build identifier `cp314-pyodide_wasm32` is already matched by the existing `build = "… cp314-*"` filter, and the pinned cibuildwheel v4.2.0 targets Pyodide 314.0.4 by default, so no version bump is needed; the wheel lands in `wheelhouse/` and flows through the existing artifact download + PyPI trusted-publishing steps untouched

**Out of scope:**

- `main` — it receives this via the periodic 1.9.X merge

### Testing

- Full test suite on both engines under pandas 3.0.x (Python >= 3.11) **and** under pandas 2.1.4 (the Ubuntu 24.04 workflow / older interpreters), proving the widened single range
- Demo: build the wasm wheel and the site locally; verify `import vtlengine`, `run()`, `semantic_analysis()` and the `run_sdmx` notebook in the Pyodide 314 kernel (Node `loadPackagesFromImports` check, as done for PR #805)
- `build-jupyterlite` job green; deployed demo loads at `/jupyterlite/lab/`
- Release job: run `release.yml` (workflow_dispatch, or the first rc release after merge) and confirm the `pyemscripten_2026_0_wasm32` wheel is built and accepted by PyPI; then `micropip.install("vtlengine")` end-to-end in a Pyodide 314 runtime (Node) against the published version

> [!WARNING]
> Even with a single unmarked constraint in `pyproject.toml`, the lock ends up with per-Python pandas entries (pandas 3 only covers Python >= 3.11 while the Poetry environment floor is 3.10), and Dependabot lock regenerations can collapse such split entries (dependabot/dependabot-core#5862 — already bites the numpy split). Review any Dependabot lock diff touching pandas, and regenerate the lock with Poetry instead.

### Additional Context

- Base: `1.9.X`, after PR #1100 lands
- pandas 3.0 release notes: https://pandas.pydata.org/docs/whatsnew/v3.0.0.html (pandas 3.0.5 declares `requires-python >= 3.11` — this is what lets a single unmarked range serve every supported interpreter)
- Pyodide 314.0 = Python 3.14.2 + Emscripten 5.0.3, wheel tag `pyemscripten_2026_0_wasm32` per PEP 783 (accepted 2026-04-06; PyPI accepts the tag since ~2026-04-21)
- Pyodide 314.0.2 distribution versions (checked 2026-09-02): pandas 3.0.2, duckdb 1.5.1, pyarrow 22.0.0, numpy 2.4.3, msgspec 0.20.0, jsonschema 4.26.0, micropip 0.11.1
- jupyterlite-pyodide-kernel 0.8.0 (2026-06-24) is the first release on Pyodide 314; latest is 0.8.5
- cibuildwheel's pyodide platform: Linux/macOS host, build frontend `pyodide-build`, `pyodide-version` option, `pyodide-prerelease`/`pyodide-eol` enable flags (v4.0.0 added the PEP 783 tags, v4.1.0 made Pyodide 314 the default, v4.2.0 — the pinned version — bundles 314.0.4)

javihern98:
## Summary

Fixes #1107

Three changes, in the order the issue lists them:

1. **pandas 3.0 support.** `pandas>=2.1.4,<3.0` becomes `pandas>=2.1.4,<4.0`, a single unmarked range (pandas 3 declares `requires-python >= 3.11`, so the lock resolves 2.3.3 on Python 3.10 and 3.0.5 on >= 3.11; `pandas-stubs` widened the same way). The pandas engine is adapted to what pandas 3 enforces, on both execution engines:
   - `drop(columns=..., axis=1)` now raises (drop / keep clauses — the bulk of the 190 failures);
   - Copy-on-Write: an in-place `where` on a column selection no longer writes through (hierarchy rule priority);
   - writing `None` into a bool column, or Arrow / string values into a float Series, raises: the DPR `when` clause widens the column first, and the Case component-level evaluation accumulates on `object` and casts the declared type at the end;
   - `Series.map` on a nullable Integer column returns float64 instead of object, so whole results read back as `"1.0"` for the Integer cast: the Numeric scalar operations iterate the values instead;
   - a multi-column row assignment indexes the row positionally (`fill_time_series`);
   - an outer merge on several keys fails with `ArrowInvalid: cannot construct ChunkedArray from empty vector and omitted type` when both operands are empty and an Arrow key column has no chunk (what `astype` leaves on an empty Arrow column): every merge of the pandas engine goes through the new `Utils/_dataframe.merge_frames`, which rebuilds those columns;
   - `concat(copy=)` and `assert_frame_equal(check_datetimelike_compat=)` are deprecated and dropped.
2. **JupyterLite demo on Pyodide 314.0.6** (Python 3.14, Emscripten 5.0.3, PEP 783 `pyemscripten_2026_0_wasm32` wheels): jupyterlite 0.8.x, no retag step, no duckdb wheel download (1.5.1 ships in the distribution), no `pysdmx==1.15.1` pin, pure-Python deps at the versions `poetry.lock` pins, `ssl` dropped from the patched lockfile (part of Pyodide 314's core). `docs.yml` follows (host Python 3.14, cache keys; the wheel cache key now also hashes `build-wheel.sh`). The served Pyodide distribution is then pruned to what the demo can reach (`prune_dist.py`, run by `build.sh` after `patch_lock.py`): the dependency closure of `vtlengine` and of the kernel, 51 of the 362 packages the tarball ships, 59 MB instead of 377 MB, taking the demo site from 444 MB to 126 MB (the last Pages artifact on `main` was 446 MB, 402 MB of it demo; GitHub caps a published site at 1 GB). Visitors download the same files either way, since Pyodide only fetches what a notebook imports; the cost is that `%pip install` of a *compiled* package outside that closure (scipy, polars...) no longer works in the demo, pure-Python packages still resolve from PyPI. The three notebooks state that both execution engines are available and that `use_duckdb=True` selects the DuckDB engine on `run` and `run_sdmx`; vtl-demo and the walkthrough repeat their first run on the DuckDB engine (both engines exercised in Node against the served Pyodide, `run` and `run_sdmx` each).
3. **wasm wheel from `release.yml`**: a `build-wasm-wheel` job runs the pinned cibuildwheel v4.2.0 with `only: cp314-pyodide_wasm32`; `[tool.cibuildwheel.pyodide]` repeats the ANTLR before-build step and runs the same `tests/API` command as the native wheels against the installed wheel, in cibuildwheel's `pyodide venv` (`pytest` plus `pyodide-unix-timezones` as test requirements). The wheel flows through the existing artifact download + trusted publishing unchanged.

> [!IMPORTANT]
> One dependency change beyond the pandas range: the numpy floor for Python >= 3.11 goes from `2.4.4` to `2.4.3`, the version the Pyodide 314.0.0–314.0.4 distributions ship (cibuildwheel v4.2.0 builds and tests against 314.0.4). With 2.4.4 the wasm test step cannot resolve numpy (pip in the pyodide venv is `only-binary`) and `micropip` cannot either. The lock stays on 2.4.6; the Dependabot split is kept.

## Checklist

- [x] Code quality checks pass (`ruff format`, `ruff check`, `mypy`) — mypy passes with pandas-stubs 3.0.5 (Python >= 3.11 lock) and 2.3.3 (Python 3.10 lock)
- [x] Tests pass (`pytest`) — full suite on both engines under pandas 3.0.5 (Python 3.12) and under pandas 2.1.4 + numpy 1.26 (the Ubuntu 24.04 combination); `ubuntu_test_24_04.yml` is unchanged and remains the pandas 2.x leg
- [x] Documentation updated (if applicable) — `docs/jupyterlite/README.md`

## Impact / Risk

- Breaking changes? No API change. Behaviour differences are limited to the pandas 3 code paths listed above; the pandas 2.1.4 suite is unchanged (5135 + 6074 tests).
- Data/SDMX compatibility concerns? None in the engine. See the note on `pysdmx` below for the wasm install story.
- Notes for release/changelog: vtlengine now supports pandas 3; the first release after this ships a `cp314-cp314-pyemscripten_2026_0_wasm32` wheel on PyPI.

## Notes

Local verification done here:
- cibuildwheel v4.2.0 `--only cp314-pyodide_wasm32` on this tree: builds the wheel (Pyodide 314.0.4 xbuildenv, pyodide-build 0.39.0) and `tests/API` passes inside the `pyodide venv` (Node 24, pandas 3.0.2, duckdb 1.5.1, pyarrow 22.0.0): 285 passed, 1 skipped in 31 s, the same result as the native wheels. `pyodide-unix-timezones` had to be a test requirement: pip there follows wheel metadata only, while the wasm pyarrow imports `unix_timezones`, a dependency only the Pyodide lockfile carries.
- `build.sh` against Pyodide 314.0.6 with that wheel, then the site in Node with `loadPackagesFromImports("import vtlengine")` on the patched lockfile: the vtl-demo and walkthrough cells (`run()` on both engines, scalars, `semantic_analysis()`, `run_sdmx` with the shipped SDMX-ML files, a `TransformationScheme`) all run. The served site was also opened in headless Chrome and the vtl-demo and walkthrough notebooks executed cell by cell through the real pyodide kernel, with no console errors. That browser run caught a `build.sh` bug: rebuilding with a stale `.jupyterlite.doit.db` skipped the tasks that patch `jupyter-lite.json`, so the kernel silently fell back to its CDN Pyodide and `import vtlengine` failed; `build.sh` now clears that state and fails if the built config does not point the kernel at `static/pyodide`. The remote `read_sdmx(url)` cell of the sdmx-objects notebook is not verifiable from Node (XMLHttpRequest), as before.
- The prune follows both graphs the demo loads packages through: the lockfile `depends` (what `import x` auto-loads) and the wheels' METADATA requirements with PEP 508 markers evaluated for Pyodide, extras included (what micropip resolves when the kernel installs itself at boot: `ipython` pulls `jedi` and `parso` that way, not through `depends`). Checked against a headless-Chrome run of the three notebooks on the unpruned site: the kept set covers all 54 files the browser fetched from `static/pyodide`, plus `idna` and `tzdata` (METADATA-only requirements). The runtime files (`pyodide.asm.wasm`, `python_stdlib.zip`...) are never touched.
- `micropip.install("vtlengine")` still cannot resolve end to end on stock Pyodide 314. Reproduced in Node against the 314.0.6 distribution plus PyPI: `Can't find a pure Python 3 wheel for: 'lxml>=6.1.0; extra == "xml"'`. This is outside vtlengine's control: `pysdmx >= 1.17` requires `lxml>=6.1.0` (a security floor, bis-med-it/pysdmx#618: lxml 6.1.0 fixes CVE-2026-41066 and bundles patched libxml2/libxslt), while the 314 distributions ship lxml 6.0.2 built against libxml2 2.9.10 / libxslt 1.1.33 (the Pyodide recipe is still 6.0.2 today), and micropip does not backtrack (pip in the pyodide venv does, to pysdmx 1.16.0). A one-call `micropip.install(["pysdmx[xml]==1.16.0", "vtlengine"])` does **not** work either (micropip resolves vtlengine's own pysdmx requirement to 1.19.0 in parallel); installing `pysdmx[xml]==1.16.0` first and `vtlengine` in a second call does, on that same lxml 6.0.2. The floor is being addressed on the Pyodide side: pyodide/pyodide-recipes#656 (on top of pyodide/pyodide-recipes#650) moves the recipes to lxml 6.1.3, libxslt 1.1.45 and libxml2 2.15.3; once a Pyodide release ships it, the plain `micropip.install("vtlengine")` resolves and the demo's lockfile no longer runs `pysdmx` below its lxml floor.The demo's patched lockfile bypasses the version check the same way, so the demo runs pysdmx 1.19.0 on the distribution's lxml 6.0.2. The real fix is Pyodide updating lxml, requested in pyodide/pyodide-recipes#655 (on top of pyodide/pyodide-recipes#650, which bumps libxml2).
- The `build-jupyterlite` and `release.yml` jobs themselves have not run yet: the former runs on merge (issue #1107 carries the `documentation` label), the latter at the next release.
- `main` is out of scope; it receives this through the periodic 1.9.X merge.


albertohernandez1995 (dismissed):
Good effort!!

albertohernandez1995 (approved):

Issue #1026, opened 2026-08-11
ratio_to_report differences


Pablolgn:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/vtlengine/issues?q=is%3Aissue) for duplicates
- [x] I have read the [documentation](https://docs.vtlengine.meaningfuldata.eu)

### Summary

**These are the errors/differences between both engines**

| # | VTL code | What fails | pandas | duckdb |
|---|---|---|---|---|
| 1 | `DS_bug_rtr_1_anidado := DS_1[drop Va_str][calc m := ratio_to_report(`<br>`    ratio_to_report(Me_num over (partition by Id_2))`<br>`    over (partition by Id_3))];` | The SQL generator puts both window calls in the same `SELECT`, and SQL does not allow nesting window functions: the inner one needs a subquery or a CTE. Not specific to `ratio_to_report` — same failure with `ratio(sum over)`, `sum(ratio over)` and `first_value(lag over)` | `1.0` ×10 ✅ | 💥 `BinderException: Binder Error: window function calls cannot be nested` |
| 2 | `DS_bug_rtr_2_cero_entre_cero := DS_1[drop Va_str][calc m := ratio_to_report(`<br>`    Me_num - Me_num over (partition by Id_2))];` | Every value in the group is 0, so the sum is 0 and the quotient is `0/0`. pandas only detects infinity (`x/0` with x≠0); `0/0` yields `NaN` and is passed through as null. DuckDB aborts on any division by zero. One of the two behaviours has to go, and the silent one is the dangerous one | `None` ×10, **silently** ❌ | 💥 `RunTimeError 2-1-3-1: At op division: Division by zero produced infinite values` |
| 3 | `DS_bug_rtr_3_part_medida := DS_1[drop Va_str][calc m := ratio_to_report(Me_num`<br>`    over (partition by Me_bool))];` | Rejecting it is correct (only Identifiers can be partitioned by), but `Me_bool` **does** exist in `DS_1` and the message sends you looking for a component that isn't there. The right message is the one `rank` gives through the dataset branch: `Only Identifiers are allowed for partitioning, found Me_bool - Role.MEASURE`. Also affects `first_value`, `lag` and analytic `sum` in the `[calc ...]` form | `SemanticError: Component Me_bool not found in Dataset __VDS_1__` | Same — fails in `semantic_analysis`, common to both engines |
[...]

PR #1110, opened 2026-09-03
Fix #1106: bind time_agg operands and the calc-identifier guard once

### LINKED ISSUE #1106: DuckDB engine: `time_agg` over `cast(<date>, time_period)` takes ~8 s to plan per statement, whatever the row count
### Summary

On the DuckDB engine, a `calc` with `time_agg("M", "D", cast(<Date component>, time_period))` spends about 8 seconds per statement inside DuckDB's binder and optimizer, independent of the number of rows (2 rows or 4070 rows cost the same). Execution itself takes 0.00 s, and the pandas engine runs the same script in 0.03 s. The results are identical on both engines, so this is a cost problem, not a correctness one.

The cause is how the transpiler nests the time-period SQL macros. DuckDB expands a scalar macro by substituting the argument expression at every place the parameter is used, so `vtl_time_agg_tp(vtl_period_parse(vtl_date_to_period(col)), 'M')` multiplies copies of the inner expressions at each level: 54 copies of the innermost `STRFTIME` survive into the final plan, and the binder processes far more before common-subexpression elimination. The calc-identifier null guard then wraps the whole expression as `CASE WHEN (expr) IS NULL THEN error(...) ELSE (expr) END`, so all of it is bound twice.

What to change, most general first:

1. Bind nested time-period operands once: have the transpiler materialize the operand as an aliased column of a subquery and apply the outer macro to the alias (measured at 0.03 s for this statement). This fixes every nested macro chain, not only this one.
2. Evaluate the guarded calc-identifier expression once, with the null guard over an alias. This halves what remains.
3. Optionally special-case a Date operand: `time_agg` over `cast(<date>, time_period)` equals `vtl_time_agg_date(col, target)` (0.00 s), but that only covers this shape.

### Scope

- `duckdb_transpiler/Transpiler/__init__.py`: `visit_TimeAggregation` (emits `vtl_time_agg_tp(vtl_period_parse({operand_sql}), '{target}')`), the `cast(<date>, time_period)` emission (`vtl_date_to_period(...)`), `_apply_time_agg_conf`, and the calc-identifier null guard (the `2-1-1-16` `CASE`, which repeats `expr_sql`).
- Possibly `duckdb_transpiler/sql/init.sql` and `time_operators.sql` if the macros themselves are reshaped (`vtl_time_agg_tp` references `p` about a dozen times, `vtl_tp_end_date` another dozen, `vtl_period_parse` uses `input` five times, `vtl_period_normalize` about thirty).
- Out of scope: the pandas engine (unaffected), the semantics of any time operator, and result values, which must not change.

### Testing

- `tests/Aggregate` `test_GL_466_1` goes from 8 s alone (15 s under `-n auto`, the slowest test of the suite) to well under a second; `GL_563_1` in TimePeriod and Cast (`time_agg` plus `cast`) from 2.3 s / 1.6 s likewise.
- The TimePeriod, Time, DateTime, Cast and Aggregate suites plus the reference-manual tests keep passing on both engines; the full suite stays green on both.
- Planning cost is measured with `PREPARE s AS <sql>` versus `EXECUTE s`, or DuckDB's detailed profile (`planner_binding`, `all_optimizers`). cProfile shows nothing because the time is inside DuckDB's `execute` call.

### Additional Context

Reproducible example (run on 1.9.3 code, Python 3.12.3, Ubuntu 24.04.4, DuckDB 1.4.5):

```python
import time

import pandas as pd

from vtlengine import run


script = """
    DS_r <- DS_1[calc identifier Id_3 := time_agg("M", "D", cast(Id_2, time_period))];
"""

data_structures = {
    "datasets": [
        {
            "name": "DS_1",
            "DataStructure": [
                {"name": "Me_1", "type": "Integer", "role": "Measure", "nullable": True},
            ],
        }
    ]
}

    data_df = pd.DataFrame(
        {"Id_1": ["A", "A"], "Id_2": ["2024-02-29", "2024-03-01"], "Me_1": [1, 2]}
    )
    start = time.perf_counter()
    run_result = run(
        script=script,
        datapoints={"DS_1": data_df},
        use_duckdb=use_duckdb,
    )
    print(f"use_duckdb={use_duckdb}: {time.perf_counter() - start:.2f}s")
```

Actual output:

```
use_duckdb=False: 0.03s
use_duckdb=True: 8.40s
```

Both engines return the same result:

| Id_1 | Id_2 | Me_1 | Id_3 |
| --- | --- | --- | --- |
| A | 2024-02-29 | 1 | 2024M2 |
| A | 2024-03-01 | 2 | 2024M3 |

Measurements on the exact SQL the transpiler emits for `GL_466_1` (407 rows):

| Step | Time |
| --- | --- |
| `vtl_period_parse(vtl_date_to_period(col))` | 0.03 s |
| `vtl_tp_end_date(...)` on that nested expression | 0.55 s |
| `vtl_time_agg_tp(..., 'M')` on that nested expression | 3.7 s |
| The same inside the calc-identifier null guard (expression repeated) | 8.1 s |
| `vtl_time_agg_tp(p, 'M')` with `p` materialized as a column | 0.01 s |
| Nested expression on 1 row / on 4070 rows | 4.1 s / 4.0 s |
| `PREPARE` alone / `EXECUTE` of the prepared statement | 3.7 s / 0.00 s |

DuckDB's detailed profile for one evaluation: `planner_binding` 1.92 s, `all_optimizers` 1.78 s (`optimizer_common_subexpressions` 0.97 s, `optimizer_expression_rewriter` 0.66 s). Projection execution: 0.00 s.

Impact: the cost is per statement, not per row, so a script with several such `calc`s pays seconds for each one regardless of data size.

> [!NOTE]
> Not introduced by PR #1100: the SQL macros and these transpiler lines are unchanged from `1.9.X`, and `main` emits the same nested expression. The pandas engine is unaffected.

albertohernandez1995:
## Summary

On DuckDB, `time_agg("M", "D", cast(Id_2, time_period))` spent about 6.6 s per statement
inside the binder and optimizer, independent of the row count. DuckDB expands a scalar macro
by substituting the argument at every place the parameter is used, so
`vtl_time_agg_tp(vtl_period_parse(vtl_date_to_period(col)), 'M')` multiplied copies of the
inner expressions at each level, and the calc-identifier null guard then repeated the whole
expression, so all of it was bound twice.

Both are addressed. A sub-expression that is read repeatedly is now computed once as a
column of a wrapping subquery, following the pattern the nested-analytics materialization
(#1026) already uses, and the outer expression reads that column:

```sql
SELECT t.* EXCLUDE ("__vtl_hoisted_0__", "__vtl_hoisted_1__"),
       CASE WHEN "__vtl_hoisted_1__" IS NULL
            THEN error('VTL 2-1-1-16: null value on a calculated Identifier')
            ELSE "__vtl_hoisted_1__" END AS "Id_3"
FROM (SELECT t.*, vtl_time_agg_tp("__vtl_hoisted_0__", 'M') AS "__vtl_hoisted_1__"
      FROM (SELECT t.*, vtl_period_parse(vtl_date_to_period("Id_2")) AS "__vtl_hoisted_0__"
            FROM (SELECT * FROM "DS_1") AS t) AS t) AS t
```

- `time_agg` hoists its parsed operand at component, clause, aggregation, dataset and
  scalar level.
- The `2-1-1-16` guard tests and returns a column instead of naming the expression twice,
  for every calc identifier rather than only the time ones.
- In a calc chain the hoisted columns are computed after the nested-analytic wrappers, take
  part in the level-conflict detection so dependencies still split into levels correctly,
  and are excluded from the passthrough so they never reach the result.

Where no enclosing statement can hold the extra column, the expression is emitted unchanged.

Measured on this branch:

| | before | after |
| --- | --- | --- |
| The issue's script (2 rows), DuckDB | 6.60 s | 0.13 s |
| The issue's script, pandas | 0.06 s | 0.06 s |
| The same shape, 500 000 rows | 8.79 s | 1.96 s |
| `tests/Aggregate` `test_GL_466_1` | 7.96 s | 0.10 s |
| `Cast` / `TimePeriod` `GL_563_1` | 0.98 s / 0.95 s | 0.15 s / 0.13 s |
| Aggregate + TimePeriod + DateTime + Cast suites | 16.68 s | 6.32 s |

## Checklist

- [x] Tests pass (`pytest`) — 6082 passed on DuckDB, 5135 passed on the pandas backend

## Impact / Risk

- No behaviour change: the same values come back on every `time_agg` shape and every calc
  identifier, error paths included. Only the shape of the generated SQL changes.
- The pandas engine is untouched.
- No API or SDMX compatibility concerns.
- Release note: `time_agg` and calc-identifier statements no longer cost seconds of planning
  each on the DuckDB engine.

## Notes

- Binding the argument inside the SQL macros instead (`list_transform([p], q -> ...)[1]`)
  removes the same planning cost but makes per-row evaluation 5.4× slower (0.115 s → 0.622 s
  on 2 M rows), so it was discarded in favour of the subquery column.
- `mypy` still reports 3 errors in `src/vtlengine/Model/__init__.py`; they are pre-existing
  on `1.9.X` and unrelated to this change.
- Reverting the transpiler change fails all 7 new tests in
  `tests/duckdb_transpiler/test_time_transpiler.py`.

Closes #1106

### DIFF
--- a/src/vtlengine/duckdb_transpiler/Transpiler/__init__.py
+++ b/src/vtlengine/duckdb_transpiler/Transpiler/__init__.py
@@ -290,6 +290,14 @@ class SQLTranspiler(StructureVisitor, ASTTemplate):
     _pending_nested_analytics: List[Tuple[str, str]] = field(default_factory=list, init=False)
     _materialize_nested_analytics: bool = field(default=False, init=False)
 
+    # Hoisted sub-expressions (#1106): DuckDB expands a scalar macro by substituting
+    # the argument at every place the parameter is used, so a nested chain of macros
+    # multiplies copies of the inner expressions and the binder pays for each one. An
+    # operand collected here is computed once as a column of a wrapping subquery and
+    # the outer expression reads that column instead
+    _hoisted: List[List[Tuple[str, str]]] = field(default_factory=list, init=False)
+    _hoist_counter: int = field(default=0, init=False)
+
     def __post_init__(self) -> None:
         """Initialize available tables."""
         self.datasets = {**self.input_datasets, **self.output_datasets}
@@ -315,6 +323,41 @@ class SQLTranspiler(StructureVisitor, ASTTemplate):
             self._in_clause = old_in_clause
             self._current_dataset = old_current_ds
             self._column_prefix = old_prefix
+
+    @contextmanager
+    def _hoist_scope(self) -> Generator[List[Tuple[str, str]], None, None]:
+        """Collect the sub-expressions hoisted while building one statement."""
+        self._hoisted.append([])
+        try:
+            yield self._hoisted[-1]
+        finally:
+            self._hoisted.pop()
+
+    def _hoist(self, expr_sql: str) -> str:
+        """Compute *expr_sql* once in a wrapping subquery and return its column name.
+
+        Returns the expression unchanged when no enclosing statement can hold the
+        extra column (issue #1106).
+        """
+        if not self._hoisted:
+            return expr_sql
+        alias = f"__vtl_hoisted_{self._hoist_counter}__"
+        self._hoist_counter += 1
+        self._hoisted[-1].append((alias, expr_sql))
+        return quote_name(alias)
+
+    def _hoisted_source(self, src: str, hoisted: List[Tuple[str, str]]) -> str:
+        """Wrap *src* in the subquery that computes the hoisted columns."""
+        if not hoisted:
+            return src
+        for alias, expr_sql in hoisted:
+            src = (
+                SQLBuilder()
+                .select("t.*", f"{expr_sql} AS {quote_name(alias)}")
+                .from_table(self._as_subquery(src), "t")
+                .build()
+            )
+        return self._as_subquery(src)
 
     @contextmanager
     def _stash_assignment(self) -> Generator[None, None, None]:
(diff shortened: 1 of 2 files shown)


javihern98 (approved):
Looks good, thanks! 😊

PR #1120, opened 2026-09-17
Fix #1119: Reject aggregations nested inside calc and filter expressions

### LINKED ISSUE #1119: Unhandled error in filter
### Initial Checks

### Summary

An aggregate with no grouping context is correctly rejected inside calc — DS1[calc x := sum(OBS_VALUE)] raises SemanticError: "At op calc: Cannot perform an aggregation inside this clause" — but the same construct inside filter never reaches that guard: DS1[filter OBS_VALUE = sum(OBS_VALUE)] fails with an uncaught AttributeError: 'NoneType' object has no attribute 'data'. The rejection itself is right.

### Reproducible Example

```python
a := DS1[filter OBS_VALUE = sum(OBS_VALUE)];
```

### Expected vs Actual

Expected: Brings a standard error message
Current: Give unhandler error. 

### vtlengine version

1.9.3

### Python version

_No response_

### OS

_No response_

### Additional Context

_No response_

javihern98:
## Summary

An aggregation inside a `filter` clause was only rejected with `SemanticError 1-2-14` when it was the whole condition (`DS_1[filter sum(Me_1) > 1]`). As soon as it was combined with a Component, as in `DS_1[filter Me_1 = sum(Me_1)]`, the clause child was no longer a Data Set: `Binary.validate` has no branch for a Component against a Data Set and returned `None`, which then crashed further down with `AttributeError: 'NoneType' object has no attribute 'data'`. The same gap affected `calc` (`Me_1 + sum(Me_1)` aborted in `Assignment.validate`), `not (...)`, `if ... then ... else` conditions and filter clauses inside join bodies, each with a different `AttributeError`.

The interpreter now records the clause op alongside `regular_aggregation_dataset` and raises `1-2-14` in `_resolve_aggregation_operand`, the point where an aggregation is resolved against the clause's Data Set, whenever that clause is `calc` or `filter`. The existing top-level guard in `visit_RegularAggregation` stays in place. Analytic invocations (`sum(Me_1 over ())`) and the `aggr` clause are unaffected.

Both engines take this path because the DuckDB backend runs the semantic pass first; the reproduction cases were probed with `use_duckdb=False` and `use_duckdb=True` and give the same `1-2-14`.

Fixes #1119

## Checklist

- [x] Tests pass (`pytest`)
- [ ] Documentation updated (if applicable) — not needed, the `1-2-14` description already covers calc and filter

## Impact / Risk

- Breaking changes? No. Scripts that previously crashed with an `AttributeError` now get `SemanticError 1-2-14`. No valid script changes behaviour.
- Notes for release/changelog? Bug fix: aggregations nested inside `calc`/`filter` expressions are rejected with `1-2-14` instead of an unhandled `AttributeError`.

## Notes

- New tests `GH_1119_1` to `GH_1119_4` in `tests/Bugs` (`ClauseBugs`) cover filter, calc, filter inside a join body and an `if` condition inside filter.
- `Binary.validate` returns `None` for an unsupported operand pair (Component against Data Set) by design; callers keep such pairs unreachable, which is what this change does for clauses.


mla2001 (approved):
Looks fine! 😊

